Issue #1, opened 2017-08-15
fix Vagrantfile


dogi:

paulbert:
Just a note that this is fixed as part of #5

paulbert:
Closing this since we got it working last week

PR #5, opened 2017-08-21
Merge reboot projects

paulbert:
This PR merges this project back with the https://github.com/ole-vi/BeLL-angular-reboot

In addition:
* The login page has been modified to use CSS rather than a static background image
* The Vagrantfile has been modified to work with the Angular CLI, fixing #1


paulbert:
Had a few more files to clean up and there was an issue that Vagrant was fixing a port collision on its own between my host computer & virtual machine's CouchDB services.  I just adjusted the host port on those to make sure the app connects to the virtual machine's CouchDB instance.

While I was working on that I found a few other things to fix, too.

Issue #19, opened 2017-08-30
Login redirect not working


paulbert:
I had implemented the AuthGuard based on [this tutorial](http://jasonwatmore.com/post/2016/12/08/angular-2-redirect-to-previous-url-after-login-with-auth-guard) to include functionality to save the route the user requests in the query string and then redirect to that page after login.

Currently that's not working as intended, so needs fixing.

nirojdyola:
1 .Login Page (stateURL = meetups)
![loginurl](https://user-images.githubusercontent.com/7495421/29909504-e767e158-8e45-11e7-8cb5-86582cf1eed6.png)

2. Meetups after login page
![meetups](https://user-images.githubusercontent.com/7495421/29909506-eae5c5d4-8e45-11e7-9681-844265684753.png)

Here, 1. checking meetups before login as shown . and 2. redirect to meetups page after login as it needs to maintain the state.

PR #21, opened 2017-08-31
[#19] Login redirect state url fixed

nirojdyola:


src/app/login/login.component.ts
```diff
@@ -32,30 +32,36 @@ require('./login.scss');
...
+export class LoginComponent {
     constructor(
         private couchService: CouchService,
-        private router: Router
+        private router: Router,
+        private route: ActivatedRoute
     ) {}
-    
+
+    ngOnInit() {
```
paulbert: I don't think you actually should use ngOnInit here, just assign returnUrl to the route.

ngOnInit is used mainly for accessing the services to initialize data.  `returnUrl` can be set with just simple assignment, so it doesn't need to utilize that lifecycle hook.

nirojdyola: Used returnUrl in constructor

paulbert:
You don't need to add it into the Constructor, either.  Just add it as a normal property of the class.

paulbert:
@nirojdyola See my latest commit.  `returnUrl` can be set immediately.  You don't need to set it in a method.

Issue #16, opened 2017-08-29
Add resource component


paulbert:
The resource component will need to be implemented in a few steps.  Let's start with the form so we can populate a resources database.  It should:

* Add a document to the database with the form fields found in the current BeLL-Apps on a form submit click with the exception of the attachment
* Not add a document if a resource with the same title currently exists
* Be linked to from the `LIBRARY` nav link

Notes:
* Adding attachments to CouchDB adds a little more complexity to this than needed.  Let's get the form working, then at a later time we can implement the attachments.
* Here's an image of the current form for reference:
![Current BeLL resource form](http://open-learning-exchange.github.io/pages/uploads/images/htmlresources.png)

nirojdyola:
1. Attachment (tested in mozilla and chrome)
   Tested files : image, pdf, docs, excel,audio

![resources](https://user-images.githubusercontent.com/7495421/30018770-e36d8874-917d-11e7-9873-729a27273114.png)

Note : Other fields need to add.
@razu9861 over to you .

PR #22, opened 2017-09-04
[#16] added attachment

nirojdyola:


src/app/resources/resources.component.html
```diff
@@ -0,0 +1,9 @@
+<form (ngSubmit)="onSubmit(resourcesForm.value)" #resourcesForm="ngForm">
+  File Upload : <input  (change)="onChange($event)" type="file" /> 
+</form>
+<div>
+  {{message}}
+  <ol>
+    <li *ngFor="let attachment of attachments">{{_.keys(attachment.doc._attachments)[0]}}</li>
```
paulbert: This looks like the only place you are using Underscore.  The resource list won't be listing out attachments, so I don't think we need to add an entire library for this one function that we will be removing in the future.

.angular-cli.json
```diff
@@ -24,7 +24,8 @@
       ],
       "scripts": [
         "../node_modules/jquery/dist/jquery.min.js",
-        "../node_modules/bootstrap/dist/js/bootstrap.min.js"
+        "../node_modules/bootstrap/dist/js/bootstrap.min.js",
+        "../node_modules/underscore/underscore-min.js"
```
paulbert: See the other comment.  I don't think we need to add Underscore at this time.

src/app/shared/couchdb.service.ts
```diff
@@ -61,7 +61,19 @@ export class CouchService {
             .then(this.handleRes)
             .catch(this.handleError);
     }
-    
+
+    //For Attachment files
+    saveAttachment(db:string,data:any,content_type:any): Promise<any> {
```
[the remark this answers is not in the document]
paulbert: Ah, I see the slight difference with the `content_type` parameter.  Rather than write a completely new function, the opts parameter for the `put` method can be used to overwrite the header.

paulbert:
To make sure this keeps moving in the right direction, I went ahead and did some work on this today.  

- I added a resource view component which uses HTML5 tags for rendering audio, video, and PDFs.  This way we can see how these browser players work with Angular.
- Adding a resource is now done in one step, a PUT request with the file attachment included as a Base64 string within the JSON.
- I added a basic file type dropdown to the resource add page which is used to tell the resource view component which player to use.
- This means we no longer need the saveAttachments method.  I'll reiterate that the methods on the CouchService all allow for changing the headers of the HTTP request.  If you need to modify the headers you do not need to add a new function.
- Also fully removed Underscore (it was still in the cli.json file and the package.json)
- Removed the resources.scss.  If we don't have custom css, no need for the file.  We can always add it later.

I think that covers it.

nirojdyola:
@paulbert  I think we need to define the type of audio and video . Since , html5 doesn't support all type of audio and video.

paulbert:
Merging this so we can continue working on resources, but I'll create an issue so we can continue the discussion about video & audio support for HTML5

PR #24, opened 2017-09-04
Testing .travis.yml

daisywatson:


.travis.yml
```diff
@@ -0,0 +1,89 @@
...
+language: node_js
+node_js:
+  - "8"
+  
+addons:
+apt:
+  sources:
+    - google-chrome
+  packages:
+    - google-chrome-stable
```
mappuji: dont use chrome, use phantom js, if we just confirm the functionalities.

http://karma-runner.github.io/0.8/plus/Travis-CI.html

.travis.yml
```diff
@@ -0,0 +1,89 @@
...
+  - TRAVIS_NODE_VERSION="7.9.0"
+cache:
+   directories:
+     - $(npm root -g)
+     - node_modules
+
+notifications:
+  webhooks:
+    urls:
+      - https://webhooks.gitter.im/e/063276d3941fa8443947
```
mappuji: I think we can use private environment variable and hiding our notifications link. We don't want some people spamming our channel right? https://docs.travis-ci.com/user/environment-variables/

dogi: @mappuji thanks I think we should open an extra/new issue for all our `.travis.yml`es ;)

/me thinks we have at least 4 ...

.travis.yml
paulbert: As the tests are set up, they are using Chrome in headless mode so these are needed to run the tests.

Issue #23, opened 2017-09-04
Loading | Spinner


nirojdyola:
I think we need some loading gif or spinner when the page is loading.

paulbert:
Adding to the project page.  Please make sure you add any new issues to the project page, thanks!

PR #29, opened 2017-09-06
[#23] Showing loader/spinner in login page and meetup page

Rupesh87:


paulbert:
Quick note: you should use `requestAnimationFrame` rather than `setTimeout` for animations.

Documentation: https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame

PR #37, opened 2017-09-10
[#8] Unit test for meetup component

lucy811:
I used *ng test* to check the failure reasons.

1) MeetupsaddComponent: 
Failed: Template parse errors: There is no directive with "exportAs" set to "ngForm" ("..
I added import { FormsModule } from '@angular/forms'; and imports: [ FormsModule] in meetupsadd.component.spec.ts

2)Error: No provider for CouchService!
I added @Component({ providers:[CouchService] }) in meetups.component.ts and meetupsadd.component.ts

3)Error: No provider for Http 
Expected undefined to be truth
I added import { HttpModule } from '@angular/http'; and imports: [  HttpModule ], in meetups.component.spec.ts and meetupsadd.component.spec.ts

unittest1

unittest2

### DIFF
--- a/src/app/meetups/meetups.component.ts
+++ b/src/app/meetups/meetups.component.ts
@@ -3,6 +3,7 @@ import { CouchService } from '../shared/couchdb.service';
 
 @Component({
   templateUrl: './meetups.component.html',
+  providers:[CouchService]
 })
 export class MeetupsComponent implements OnInit {
   message = "";
--- a/src/app/meetups/meetups.component.ts
+++ b/src/app/meetups/meetups.component.ts
@@ -3,7 +3,6 @@ import { CouchService } from '../shared/couchdb.service';
 
 @Component({
   templateUrl: './meetups.component.html',
-  providers:[CouchService]
 })
 export class MeetupsComponent implements OnInit {
   message = "";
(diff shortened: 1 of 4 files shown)


src/app/meetups/meetups.component.ts
```diff
@@ -3,6 +3,7 @@ import { CouchService } from '../shared/couchdb.service';
 
 @Component({
   templateUrl: './meetups.component.html',
+  providers:[CouchService]
```
paulbert: This `providers` line should go into the `configureTestingModule` in the `meetups.component.spec.ts` file.

I had to read up to double check, but this actually instantiates a new `CouchService` class.  While at this point it wouldn't break the app (the class isn't holding a shared state across components), for now we should stick with having a singleton service to connect to the DB.

src/app/meetups/meetupsadd.component.ts
```diff
@@ -16,7 +16,8 @@ import { CouchService } from '../shared/couchdb.service';
         </div>
         <button name="singlebutton" class="btn btn-primary" type="submit">Save</button>
     </form>
-    `
+    `,
+    providers:[CouchService]
```
paulbert: Same comment as above, please move this into the `meetupsadd.component.spec.ts` file

lucy811:
@paulbert I have moved providers:[CouchService] to .spec.ts files.

Issue #38, opened 2017-09-11
Translate (Multilanguage)


nirojdyola:
We need translation as we have done in Bell App. We need to integrate translation in angular project too.

PR #39, opened 2017-09-11
[#38] - MultiLanguage Integrated

nirojdyola:
MulitLanguage is used in login and navigation page. 
ng 2 translate modeule is used
Please check it once.
language1

### DIFF
--- a/.angular-cli.json
+++ b/.angular-cli.json
@@ -8,7 +8,8 @@
       "root": "src",
       "outDir": "dist",
       "assets": [
-        "assets"
+        "assets",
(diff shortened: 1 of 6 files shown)


paulbert:
@nirojdyola I'm going to have to read up a bit on this ng2-translate module.

Did you look at the official Angular internationalization implementation (link below)? From what I can tell the coding is much simpler, but the build process may be a little involved.

https://angular.io/guide/i18n

paulbert:
Closing this since we're moving on with Angular built in internationalization

PR #46, opened 2017-09-20
used i18n for translation

nirojdyola:
ng serve --aot --i18n-file=src/i18n/messages.es.xlf --locale=es --i18n-format=xlf

sketch

@Paul But I want to switch the language dynamically without using the above command for other other language.
I have followed the doc from https://medium.com/@feloy/deploying-an-i18n-angular-app-with-angular-cli-fc788f17e358 . I think it needs to deploy in production . I"ll also try it.


src/app/home/home.scss
```diff
@@ -1,4 +1,4 @@
-@import '../variables';
+@import '../variables.scss';
```
paulbert: Is this needed for the translation?  I heard there was difficulty compiling the sass, so could you post the error you're seeing here & we can decide if that requires a separate issue.

dogi:
@dogi created https://crowdin.com/project/ole-planet/ - please join

src/app/home/home.scss
nirojdyola: I got an error 
$ ./node_modules/.bin/ng-xi18n
Error: Compilation failed. Resource file not found: D:/planet/src/app/variables

lmmrssa:
`ng xi18n --output-path src/i18n`
This will update message.xlf file. Then it will be synced to crowdin and changes done there will create pull request to the branch. Once that pull request is merged, all other language files will be updated inside src/i18n folder

Issue #53, opened 2017-09-26
Create admin user first


paulbert:
Since we will creating a CouchDB admin to secure the database, we should create an admin as the first user created in the app so our development can follow with the DB out of the "Admin Party" mode where any client can make any change to the database.

See the below for information on the HTTP request to create an admin user:
http://docs.couchdb.org/en/2.1.0/intro/security.html

PR #56, opened 2017-09-29
update login.compononent.ts to create admin on first time login (#53)

ajGingrich:
As discussed in the Angular hangout yesterday, I am creating a working PR so the process will be more transparent. **Note**: Still a work in process!

Before Submitting, `checkNumberUsers` looks for existing users and if there aren't any, then creates an Admin user. This is fine for now but for production will need to be modified so that there is a *superUser* or *root* and the other *admins* only have privileges so nobody can go rogue. 

As described in CouchDB docs
> In our previous discussions, we dropped some keywords about how things without the Admin Party 
> work. First, there’s admin itself, which implies some sort of super user. Then there are privileges. Let’s 
> explore these terms a little more. 

To do:
- [ ] fix createAdmin `put` because does not actually creating an admin

Please provide feedback if you think this isn't heading the right direction, so I don't stray any further!

### DIFF
--- a/src/app/login/login.component.ts
+++ b/src/app/login/login.component.ts
@@ -43,12 +43,16 @@ export class LoginComponent {
 
     createMode:boolean = this.router.url.split('?')[0] === '/login/newuser';
     returnUrl = this.route.snapshot.queryParams['returnUrl'] || '/';
-    model = { name:'', password:'', repeatPassword:'' }
+    model = { name:'', password:'', repeatPassword:'' };
     message = '';
 
     onSubmit() {
         if(this.createMode) {
-            this.createUser(this.model);
+            if (this.checkAdminExistence()) {
+                this.createAdmin(this.model);
+            } else {
+                this.createUser(this.model);
+            }
         } else {
             this.login(this.model);
         }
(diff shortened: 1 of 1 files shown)


ajGingrich:
- [x] I am able to successfully add an admin user if there currently isn't one. However, we then obviously run into problems with authentication for future actions such as creating a regular user. 

I'm not sure how we want to handle this. Would it best to create a default admin user in the vagrant file and then hardcode in the authentication?

src/app/login/login.component.ts
```diff
@@ -43,12 +43,16 @@ export class LoginComponent {
...
+    model = { name:'', password:'', repeatPassword:'' };
     message = '';
 
     onSubmit() {
         if(this.createMode) {
-            this.createUser(this.model);
+            if (this.checkAdminExistence()) {
+                this.createAdmin(this.model);
+            } else {
+                this.createUser(this.model);
```
paulbert: As written, the checkAdminExistence method actually doesn't return anything, so this will always be false.  The functions within a `.then()` chain don't return their value to a parent function, they return it to the next `.then()`.

As a note when you send off an async request with a promise like the `couchService.get()` that anything that needs to happen after needs to be within one of the `.then()` functions attached to the promise.

src/app/login/login.component.ts
```diff
@@ -64,12 +68,36 @@ export class LoginComponent {
...
+                    console.log('Need admin privilage to create a user');
+            });
+        } else {
+            this.message = 'Passwords do not match';
+        }
+    }
+
+    createAdmin({name,password,repeatPassword}:{name:string,password:string,repeatPassword:string}) {
+        if(password === repeatPassword) {
+            this.couchService.put('_node/nonode@nohost/_config/admins/' + name + ' -d ' + "'" + '"' + password + '"' + "'", {})
```
paulbert: The PUT request we send from the browser will differ from the `curl` requests in the CouchDB docs.  The first parameter should be the URL, which in this case is `'_node/nonode@nohost/_config/admins/' + name`, and the second parameter is the data sent, which in this case should just be the password.

src/app/login/login.component.ts
```diff
@@ -64,12 +64,34 @@ export class LoginComponent {
...
+            this.couchService.put('_node/nonode@nohost/_config/admins/' + name, password)
+                .then((data) => {
+                    this.reRoute();
                 }, (error) => this.message = '');
         } else {
             this.message = 'Passwords do not match';
         }
     }
 
+    checkAdminExistence() {
```
lmmrssa: isn't there other way to check admin exists not checking document.

ajGingrich:
I am closing and opening a new branch because the commit history and merge conflicts is giving me some problems

Issue #59, opened 2017-10-02
Set up linting before git push


paulbert:
I'm working on setting up automatic linting before any push to the repository.  First thought is to have a pre-push hook, but there is a slight downside that git hooks aren't automatically stored in the repository so everyone would need to run a command to set up the hooks.

That's certainly the simplest solution, though, so I'll be setting this up for the meeting on Wednesday for everyone to take a look at.

PR #63, opened 2017-10-03
Pre push linting

paulbert:
Fix for #59

Started this branch off the e2e one by accident, but I think after merging the master back in this should be OK.

Adds three npm commands: 
* `lint` and `lint-fix` which run linting from the host machine
* `install-hooks` which copies all hooks in the `git-hooks` folder to the `.git/hooks` folder.  The `.git` folder isn't included in the repository, so after any changes to the hooks you'll need to run this (maybe I should just add this to the Vagrantfile?)

I'm calling it a night, but we can go over this tomorrow at the meeting!


lmmrssa:
@paulbert It looks good but it already did the linting as you pushed this looks huge to have a glance at :) . Actual change seems to be only the package included 👍

paulbert:
@lmmrssa Yes I went through and made the app comply with all of the linting rules in addition to adding the pre push hook.  I had to do that anyway to test the hook!

package.json
```diff
@@ -7,8 +7,10 @@
     "start": "ng serve",
     "build": "ng build",
     "test": "ng test",
-    "lint": "ng lint",
-    "e2e": "ng e2e"
+    "lint": "vagrant ssh -c 'cd /vagrant;ng lint'",
```
dogi: can we have a normal lint and a vagrant lint?

PR #64, opened 2017-10-04
Add Courses component

LKhadka:
To Do:
- [x] Separate validation logic from the components
- [x] Add more validations for different fields
- [x] Display proper validation errors
- [x] Navigate to another page when form is submitted
- [ ] ~~Add design json files to Docker~~ Does not work in Docker


src/app/form-error-messages/form-error-messages.component.ts
```diff
@@ -0,0 +1,54 @@
...
+      list-style-type: none;
+      font-size: .8rem;
+      padding-left: .2em;
+    }`
+  ]
+})
+export class FormErrorMessagesComponent {
+  private static readonly errorMessages = {
+    // even though they have the same return type, different names are used since it's more clear & easier to understand/debug
+    required: () => 'This field is required',
```
lmmrssa: is this component going to be used for other forms as well?
If so can we make some of these like required as well to be params as it seems like errors will show up accumulated in one place.

PR #57, opened 2017-10-02
Comment out one test in login component and add unit tests in meetup component 

lucy811:
1. Because" 'Hi, ' + name + '!' "has been deleted in the login.component.ts, so I comment out 'Should greet users' unit test in login.component.spec.ts.

2. I add 'should make a get request to couchService', 'should getmeetups' and 'should show There was a problem getting meetups' these three unit tests to test  getMeetups() in meetups.component.ts.

3. I add 'should make a delete request to couchService', 'should delete a specific meetup' and 'should There was a problem deleting this meetup' these three unit tests to test deleteMeetup() in
meetups.component.ts.

4. I add  'should make a post request to couchService', 'should show meetup created: title correctly', 'should message Please complete the form' and 'should message There was a problem creating the meetup' these four unit tests to test onSubmit() in meetupsadd.compoent.ts.

unittestresult

PR #70, opened 2017-10-09
Update unit test in meetup and fix unit test failures in resources and users

lucy811:
Follow up of pull request [#57](https://github.com/ole-vi/planet/pull/57)

Based on the work last week (commented out 'Should greet users' in login, and added unit tests in meetup and meetupadd)

I update some unit tests this week, and they are shown below:

1) I revise 'should getmeetups' and 'should delete a specific meetup' two unit tests in meetup component. I change from ' expect(component.getMeetups()).toBeUndefined()' and  'expect(component.deleteMeetup(testModel.id, testModel.rev)).toBeUndefined();' to expect the content in object.

2) I fix the failure 'should be  created' in ResourcesComponent. 

3) I fix the failure 'Should make two GET requests to CouchDB for admin' in Users(Init).

results
test2


src/app/meetups/meetups.component.spec.ts
```diff
@@ -4,27 +4,99 @@ import { MeetupsComponent } from './meetups.component';
...
+        expect(spy).toHaveBeenCalledWith('meetups/' + meetupdata1.id + '?rev=' + meetupdata1.rev);
+    });
+  });
+
+  it('should delete a specific meetup', async() =>{
+    spy = spyOn(couchService, 'delete').and.returnValue(Promise.resolve(meetuparray));
+      component.deleteMeetup('1', 'qwrjksf');
+      fixture.whenStable().then(() => {
+          fixture.detectChanges();
+          expect(meetuparray.calls.allArgs()).toEqual({id:'2', rev:'ghjjdrt',title:'angularhangout', description:'twice a week'});
```
lmmrssa: do we need these rev id in data? It might come up with some random id by couchdb itself.

paulbert:
Hi Lucy,
I updated the meetups test to give an example of what I meant in the meeting last week.

The test now grabs the HTML elements that Angular creates after `getMeetups()` runs and checks if the title and description in the HTML matches the title and description from the mock data.

While updating that I realized the `should delete a specific meetup` test isn't necessary, because after a meetup is deleted the app calls `getMeetups()` again.

The other key difference that you should look at is the mock data format.  Data from an `all_docs` request to CouchDB comes back in this format:

```
{
  rows: [
    { doc: { <data for first item> } },
    { doc: { <data for second item> } },
    ...
  ]
}
```

There were a few other small changes I made:
* I added some classes to the HTML with a `km-` prefix.  I think I'll add this to our onboarding, but it'll probably be a little less error prone if we add classes to HTML elements we want to grab in tests.
* I removed the async function.  As I was looking through the Angular testing documentation I noticed that it said async isn't necessary when using Webpack (which the CLI uses under the hood).  It's not readily apparent that we are using Webpack, so it's fine that you used it before.

I actually think that's it for this PR, but I'd like to check the meetupadds tests a little more closely.  Let me know if you have any questions.

lucy811:
@paulbert Thank you. I have learned these commits.

lucy811:
Try to resolve conflict and to update some name convention in meetupadd.component.spec.ts.  But after I resolve conflict, it appears new unit test errors in other components.  I back up my previous code about unit tests in meetup component. Close this pull request. Plan to create a new branch to fix it.

PR #71, opened 2017-10-09
[WIP] Auto push docker imge

mappuji:
Here it comes, docker image automagically build for you. It is only the x86 version right now, will add ARM in couple of commits ;)


mappuji:
Close for move to upstream branch

Issue #76, opened 2017-10-10
Can we have our backend connect to DB instead of our client via browser?


mappuji:
https://github.com/ole-vi/planet/blob/d80af984ba5df1a8e388cda8ccc6eb241ac7597c/src/app/shared/couchdb.service.ts#L12

1. Hardcoding stuff is not good, we need to create a config file
2. Can we abstracting away that the things that connect to DB is the backend not the client via their browser, so that we can have multiple domain?

paulbert:
I would say not at this stage of development since to create a backend would be duplicating work already done for us with the `ng serve` command and the CouchDB API. @dogi if you disagree we can get working on a solution that would work for this.

1. We do have the Angular environment variables (as used for the port) available which change based on the build.  You're right that this shouldn't be hardcoded, but I also think rather than setting this in a configuration we could alternatively use the domain name the client is at as the default.  Unless you expect the domain/ip address of the app docker to be different from the CouchDB docker.

2. Having a backend "middleman" to handle which domain a request goes to isn't necessary on its own.  What are the scenarios you are thinking we need this for?  @dogi if you have any more detail please let me know.

mappuji:
Sorry for not giving the clear explanation, but I can demonstrate it if you are in talk.

Here is the use cases:
* I setup nation on a server which has IP Address and/or domain.
* I access from outside so I use the domain, tada 🎉 I can't access since db is pointing to `127xxx`
* Our problem is the one who makes the `API Call` to CouchDB is our client in their browser not our app with internal link.
* Actually our Angular is a webs server, we should able to make internal link from our angular app to CouchDB instead of our client's browser do the call which is less secure and bound to the host declared in config or in the config file. If we can connect with internal link, it will be easier if we have multiple hostname.

mappuji:
Solved by #81

PR #81, opened 2017-10-13
Environment var for db

paulbert:
Related to #76, but doesn't fully add the different backend.  Uses environment variable for CouchDB/backend URL for all environments.

### DIFF
--- a/src/environments/environment.test.ts
+++ b/src/environments/environment.test.ts
@@ -1,4 +1,4 @@
 export const environment = {
   production: false,
-  couchPort: '5984'
+  couchAddress: 'http://127.0.0.1:5984/'
 };
--- a/src/environments/environment.ts
+++ b/src/environments/environment.ts
@@ -5,5 +5,5 @@
 
 export const environment = {
   production: false,
-  couchPort: '2200'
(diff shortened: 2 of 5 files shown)


mappuj1:
This is sound a good approach 👌

src/environments/environment.ts
```diff
@@ -5,5 +5,5 @@
 
 export const environment = {
   production: false,
```
mappuj1: how about if I want to build for prod with `ng build --prod`?

src/environments/environment.test.ts
```diff
@@ -1,4 +1,4 @@
 export const environment = {
   production: false,
-  couchPort: '5984'
+  couchAddress: 'http://127.0.0.1:5984/'
```
dogi: @paulbert so for test we access couchdb on 5984 in all others 2200 ... right?

mappuj1: @paulbert by the way how to switch between environment with `ng`?

mappuj1: @dogi since test passed seems the config used is the `environment.ts` instead of `environment.test.ts` what do you think @paulbert ?

PR #85, opened 2017-10-13
[WIP] Add separate script to setup couchdb

paulbert:
In one of the post lint PRs I ended up losing the curl command for the _resources_ db in the Vagrantfile and thought it'd be nice to have a script to setup the databases without running `vagrant destroy` and `vagrant up`.

The script will add databases (throws an error if the database exists which is OK) and replace the courses validator design doc.  We can probably improve this by writing a function for the design doc "upsert", but for now this works.

Need to update the README for the script flags

`-p`: Port number
`-u`: Username (if there is an admin on your DB)
`-w`: password


paulbert:
Is this something that can work with your Docker setup, @mappuji?  I think it's nice to have a single file for DB setup, but if not this can still be a temporary solution for those of us working in Vagrant.

I can add a flag for domain name as well so it's not always `127.0.0.1`.

PR #86, opened 2017-10-14
Production docker

mappuji:
Requires testing in CI and actual device.

Known issue:
* Docker in our Rpi server and travis is out of date and not supporting our way of optimizing docker with multistage build.

---
**Update**: it is verified to work in our rpi-server called `kraken` but in travis we need to update docker, I think do it in separate PR is better.

result here
* https://hub.docker.com/r/treehouses/planet/tags/ --> production
* https://hub.docker.com/r/treehouses/planet-dev/tags/ --> development

**Sorry** for pushing many commits only for docker and deployment stuffs


docker/planet-dev.yml
```diff
@@ -7,11 +7,11 @@ services:
       - "2200:5984"
       - "2201:5986"
   db-init:
-    image: treehouses/planet-dev:db-init-latest
+    image: treehouses/planet:db-init-latest
     depends_on:
       - couchdb
   planet:
-    image: treehouses/planet-dev:latest
+    image: treehouse/planet-dev:latest
```
lmmrssa: it probably should be treehouses.

mappuji: thank you for catching this

docker/rpi-planet.yml
```diff
@@ -0,0 +1,21 @@
...
+      - "2201:5986"
+  db-init:
+    image: treehouses/planet:rpi-db-init-latest
+    depends_on:
+      - couchdb
+  planet:
+    image: treehouses/planet:rpi-latest
+    ports:
+      - "3000:3000"
+    volumes:
```
mappuji: you don't need volume for production, but the couchdb would

docker/planet.yml
```diff
@@ -0,0 +1,25 @@
...
+      - couchdb
+  planet:
+    image: treehouses/planet:latest
+    ports:
+      - "80:80"
+    environment:
+      - HOST_PROTOCOL=http
+      - DB_HOST=127.0.0.1
+      - DB_PORT=2200
+    volumes:
```
mappuji: you also dont need this for production

PR #87, opened 2017-10-14
add list/table scss (Fixes #15)

### LINKED ISSUE #15: Add list/table scss
Add reusable table scss (either in style.scss or another file) to help implement mockup:

Planet mockup

LKhadka:
- ~~I thought the issue was about making a reusable table so I moved it to~~ Created a new shared module because different modules cannot import the same component. I also moved the form error messages component up to the shared module.
- I did not rely on bootstrap's classes cause it would be easier to move away from bootstrap if we were needed to do so in the future and I felt Bootstrap's grids were quite limiting. I could use Bootstrap's default grids if we need to.
- Responsive design is an issue


paulbert:
@LKhadka Hmm...I was thinking the scope for this would be more focused to just CSS that would be reusable across all of our tables.  I'll have to take a look at this module approach a little more, but the CSS part is looking good so far.

LKhadka:
I deleted the list item component and moved the html code to the resources file with the mock data

ajGingrich:
It looks like the sass would great for for html resources but less helpful for tables such as meetups or courses. For example, this is the table from Bell Apps which has a lot less information such as stars and is primarily actions. 
screenshot-2017-10-19 bell myapp

LKhadka:
@paulbert are we using the old design or is it supposed to look similar to the new resources table?

paulbert:
I merged this up to master to get this moving.  The one thing that I changed is removing the `sharedModule` for the `planetFormsModule`.  I like the idea of having all of the required modules for forms in the `planetFormsModule`, and we could switch to that as a next step or in another PR.

I'm ok with merging this now and creating an issue to switch over to real resources rather than the mock data, but will defer to @lmmrssa & @dogi if they want to wait.

PR #106, opened 2017-10-20
Courses date validation fix (Fixes #102)

### LINKED ISSUE #102: Courses not adding
In the current courses component isn't adding courses, the PUT request is returning with a 403 error.  Could be an issue with having an admin on the CouchDB.

paulbert:
Thanks to @ajGingrich for pointing me in the right direction, while looking into what was off with the validation I saw it was a pretty quick fix.

Also added error catching for the `addCourse` function.


src/app/courses/courses.component.ts
```diff
@@ -111,8 +111,12 @@ export class CoursesComponent {
 
   async addCourse(courseInfo) {
     // ...is the rest syntax for object destructuring
-    await this.couchService.post(this.dbName, { ...courseInfo });
-    this.router.navigate(['/']);
+    try {
+      await this.couchService.post(this.dbName, { ...courseInfo });
+      this.router.navigate(['/']);
+    } catch (err) {
+      console.log(err);
```
LKhadka: Rather than logging the error to the console maybe it would be better to so the error directly to the user?

paulbert: I think we'll need to add an error display popup/feature in the future, so I could add a comment there that'll make it easier to pick out as a todo when implementing that.

In the meantime for development getting the full error output from CouchDB will be helpful for us whenever we access the database.

ajGingrich:
Thanks for the fix and glad I was able to to point to the correct whereabouts. The changes look good to me but should just probably be tested before merging.

PR #107, opened 2017-10-20
add courses list view (Fixes #83)

### LINKED ISSUE #83: Add courses list view
Similar to #82, the courses link should land on a list of available courses with buttons to manage the course.  I'll try to get a screenshot of the design for you to work off of, but in the meantime something similar to the current BeLL-apps will be OK.

For the scope of this issue, please just add HTML & TypeScript code for this.  I'd like broader CSS for tables to be used, and have included that as a separate issue #17.  That being said, coordinating development with the work on the CSS to incorporate that into this PR would be great!

ajGingrich:
This is a working pull request for #83. This looks like a lot of changes and new files, but the majority is moving the add course component to a sub-directory. 

I'm not sure how we want to handle the directory structure with subcomponents. I suggest this that each subcomponent should have a directory. `courses - add - coursesadd.component.ts, coursesadd.component.html, spec, scss` etc With a separate directory for each `add, request, open, resign, manage...etc`.

In thinking about this, I added a `courses-module` and `courses-router-module`


LKhadka:
The [angular style guide](https://angular.io/guide/styleguide#separate-file-names-with-dots-and-dashes) recommends using dashes to separate words in a file. For eg, you should rename `coursesadd.component.html` to `courses-add.component.html`. You should also rename folders to be more descriptive like you could change `add` to `add-courses`

src/app/courses/request/coursesrequest.component.ts
```diff
@@ -0,0 +1,8 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'planet-courses',
+  templateUrl: 'coursesrequest.component.html',
+  styleUrls: ['coursesrequest.component.scss']
```
LKhadka: Since you are using the same styles as the add component, you can import the add component stylesheet instead of recreating them same file.

ajGingrich: Yes I agree, a dash should be used for the courses-add files. For now I was just following the existing format used in the `meetupsadd` component. As for the courses-request styles, it is just a placeholder at the moment. There is nothing in the html file either.

LKhadka: I see. With so many components for Courses, should it be made into a feature module, @paulbert?

paulbert: Yes, I agree both to the dash notation and converting this to a feature module.

As for folder naming, I think we can go with a `<descriptor>-<feature>` format rather than the opposite which the Angular style guide seems to suggest.  That way we can have folder names which will auto complete with TAB when navigating in the shell.

LKhadka: Yep you are right @paulbert. So the folder names should be opposite to what I said like the `add` folder should be renamed to `courses-add`

ajGingrich:
@LKhadka We all discussed this at the Angular Meeting last night and decided to used `feature-descriptor` for the folders

LKhadka:
OH I see @ajGingrich

src/app/courses/courses.component.html
```diff
@@ -1,119 +1,30 @@
...
-
-    </div>
-  </div>
-  <button type="submit" class="btn btn-primary">Submit</button>
-  <button type="button" class="btn btn-danger" (click)="cancel()">Cancel</button>
-</form>
-
-<p>Form value: {{ courseForm.value | json }}</p>
-<p>{{ courseForm.valid }}</p>
+<div class="container">
```
LKhadka: We are using the `.container` class in a parent component. Som you don't need to use it here.

src/app/courses/add-courses/constants.ts
```diff
@@ -0,0 +1,42 @@
+export const gradeLevels = [
```
lmmrssa: this seems to have same constants as in course-request/constants

Issue #96, opened 2017-10-17
Nation : List all communities of nations


Pr0chin:

paulbert:
@Rupesh87 Make sure you remove the `.scss` file if you do not add styles for this component.

Also, @Pr0chin, could you elaborate on what the goal of this issue is?  Is this list linked to from the nations list for an administrator?

Pr0chin:
Yes @paulbert . 

* It a nation list that has been fetched from nation database. Its a similar concept that has been implemented on *BeLL-App*.
* Its an issue where we add Nation, showing list of the nations that had already registered.
* And the list of communities and the accepting / pending following communities

nirojdyola:
Snapshot : 
1. "Community" database is used.
2. Listed communities
3. Filter can be done by typing nation name and selecting status(Pending/Accepted/Rejected) from the select box.
4. can delete the community.

![Community.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/58aeb9eb257d931a000df4b2/125516c66e82c728ace21e0d46cb9b972e978da09bf0fd0fe64de29b234324676b5ee17e2981b72dbc142d00590947ae40585f46aaaddf7ea2b73862964e49bc8f3900299f7744eeaa880dbe06b0218e4f8dbebf8c4cc4e13a24e299c2.png)

PR #100, opened 2017-10-18
[#94] addnation (connects #94)

Pr0chin:

PR #99, opened 2017-10-17
Nation: list nation

razu9861:
screen shot 2017-10-17 at 4 33 31 pm

PR #111, opened 2017-10-23
96 list all communities of nations (connects #96)

Rupesh87:


paulbert:
I think this should go into a center/nation/community management module rather than its own component as part of #100 and #99.  When you combine these PRs please consider this approach.

src/app/community/community.component.ts
```diff
@@ -0,0 +1,50 @@
...
+  community = [];
+  filter = '';
+
+  constructor(
+    private couchService: CouchService
+    ) { }
+
+  getcommunitylist() {
+      this.couchService.post('community/_find',{
+                    "selector": {
```
paulbert: Please make sure indentation is consistently 2 spaces

PR #124, opened 2017-10-25
Style guide for contributing

paulbert:
Here's a first draft.  Want to keep it short and relatively simple, and we can add to this as more things come up that we'd want a rule for.

### DIFF
--- a/Style-Guide.md
+++ b/Style-Guide.md
@@ -1,6 +1,8 @@
 # Planet Style Guide
 ## Editor Config
 If you have an IDE that supports a `.editorconfig` file, please make sure you take the necessary steps to install those files.  There's moer info on specific IDE installations [here](http://www.editorconfig.org).
+
+This will help with a few things: maintaining 2 space indentations, ensuring there is a newline at the end of each file, and removing unnecessary whitespace at the end of a line.
 ## Angular & TypeScript
 ### Template & Style
 If the HTML & CSS are fewer than 12 lines total, please keep the `template` and `style` in the `.component.ts` file.  Otherwise, you should move one or both to separate `.component.html` or `.component.scss` file(s).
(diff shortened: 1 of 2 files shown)


paulbert:
Quick link to the guide: https://github.com/ole-vi/planet/blob/style_guide/Style-Guide.md

Style-Guide.md
```diff
@@ -0,0 +1,53 @@
...
+Below is a brief summary of code standards we'd like to keep with this project.  For the most part this follows the [Angular Style Guide](https://angular.io/guide/styleguide), but in a more concise way so you have a quick to read reference guide.
+## Editor Config
+If you have an IDE that supports a `.editorconfig` file, please make sure you take the necessary steps to install those files.  There's more info on specific IDE installations [here](http://www.editorconfig.org).
+
+This will help with a few things: maintaining 2 space indentations, ensuring there is a newline at the end of each file, and removing unnecessary whitespace at the end of a line.
+## Angular & TypeScript
+### Template & Style
+If the HTML & CSS are fewer than 12 lines total, please keep the `template` and `style` in the `.component.ts` file.  Otherwise, you should move one or both to separate `.component.html` or `.component.scss` file(s).
+
+If there are more than 12 lines, but the component has little to no TypeScript functionality, then an exception can be made if code reviewers agree to it.
```
ajGingrich: Should we just remove this line about leaving the html within the file if it is longer than 12 lines? That way we will have a set rule concerning the 12 lines and we don't have to discuss this any further. I think it will be more efficient in the long run.

Style-Guide.md
```diff
@@ -0,0 +1,53 @@
...
+Class name: CoursesComponent
+```
+#### Resources Review Service
+```
+File: resources-review.service.ts
+Class name: ResourcesReviewService
+```
+We suggest using max one sub-feature word to keep file and class names from getting too long.
+### App Directory Structure
+Within the `src/app` directory, each feature should have its own directory.  Within that directory if, as we add sub-features, there are more than 9 files we should create a sub-directory with the same naming convention as files (i.e. Resources Review would be in a `resources-review` directory).
```
ajGingrich: there should be an `if` after the second comma in the second sentence. `if there`

PR #130, opened 2017-10-30
[#129] Community mock data (connects #129)

Pr0chin:


couchdb-setup.sh
```diff
@@ -43,6 +43,8 @@ curl -X PUT $COUCHURL/_global_changes
 curl -X PUT $COUCHURL/meetups
 curl -X PUT $COUCHURL/resources
 curl -X PUT $COUCHURL/courses
-
+curl -X PUT $COUCHURL/communityregistrationrequests
 # Add or update design docs
 upsert_design courses course-validators ./design/courses/course-validators.json
+# Adding communityregistrationrequests dummy data/objects
```
dogi: wonder if there is a better place to add this ...

Issue #134, opened 2017-10-31
Nation: convert add nation form to modal 


razu9861:

razu9861:
make type filed hidden

Pr0chin:
- [ ] Made Modal for Add Nation.
- [ ] View and layout changed.
- [ ] Added nation data render at first and removed navigation function.

PR #141, opened 2017-11-01
[#134] nationmodal (connects #134)

Pr0chin:


src/app/nation/nation.component.html
```diff
@@ -1,61 +1,66 @@
...
-  </div>
-  <div class="row">
-    <div class="form-group col-4">
-      <label i18n>Nationurl</label>
-      <input type="text" class="form-control" formControlName="nationUrl">
-      <planet-form-error-messages [control]="nationForm.controls.nationUrl"></planet-form-error-messages>
+<div class="container">
+  <h1 i18n>Nation List</h1>
+  <!-- Trigger the modal with a button -->
+  <button type="button" class="btn btn-info btn-lg" (click)="refresh()" data-toggle="modal" data-target="#myModal">Add</button>
```
lmmrssa: try to make modal name more appropriate

src/app/nation/nation.component.ts
```diff
@@ -36,49 +37,45 @@ export class NationComponent implements OnInit {
...
   onSubmit(nation) {
-    this.i = 0;
-    if (nation.nation_name !== '' && nation.nationurl !== '' && nation.type !== '') {
-      this.couchService.post('nations', {
-          'admin_name': nation.adminName,
-          'nation_name': nation.name,
-          'nationurl': nation.nationUrl,
-          'type': nation.type
-        })
+    if (nation.adminName !== '' && nation.name !== '' && nation.nationUrl !== "") {
```
lmmrssa: if validation is working this should not be required

src/app/nation/nation.component.ts
```diff
@@ -36,49 +37,45 @@ export class NationComponent implements OnInit {
...
+      this.couchService.post('nations', formdata)
         .then((data) => {
-        alert('Nation has been sucessfully created');
-        this.router.navigate(['nation']);
-        location.reload();
-      }, (error) => this.message = 'Error');
+          formdata['_id'] = data.id;
+          formdata['_rev'] = data.rev;
+          this.nation.push({doc:formdata});
+          jQuery('#myModal').modal("hide");
```
lmmrssa: try to use angular events

src/app/nation/nation.component.ts
```diff
@@ -1,7 +1,7 @@
 import { Component, OnInit } from '@angular/core';
 import { Location } from '@angular/common';
 import { Router } from '@angular/router';
-
+declare var jQuery:any;
```
lmmrssa: this might not be required if used angular

src/app/nation/nation.component.html
```diff
@@ -1,61 +1,66 @@
...
-  <button type="submit" class="btn btn-primary" i18n>Submit</button>
-  <button type="button" class="btn btn-danger" (click)="cancel()" i18n>Cancel</button>
-</form>
+</div>
 
-<p>Form value: {{ nationForm.value | json }}</p>
-<p>{{ nationForm.valid }}</p>
-</ng-template>
+<div>
+  <p>{{message}}</p>
```
lmmrssa: fix indentation below

src/app/nation/nation.component.html
```diff
@@ -1,61 +1,66 @@
...
+                    <input type="text" class="form-control" formControlName="nationUrl">
+                    <planet-form-error-messages [control]="nationForm.controls.nationUrl"></planet-form-error-messages>
+                </div>
+              </div>
+              <div>
+                <p>Form value: {{ nationForm.value | json }}</p>
+                <p>{{ nationForm.valid }}</p>
+              </div>
+              <div class="modal-footer">
+                <button type="submit" class="btn btn-primary" i18n>Submit</button>
```
lmmrssa: add ng-events

paulbert:
I added some commits to help move this along.  On the use of jQuery in this component for the modal:

Bootstrap includes simple methods on jQuery objects for adding and removing the modal in addition to attributes that can be placed in the HTML to control the popup appearing.  To avoid this we could:

- Write a function in native JS - more custom code for us to write that is already written for us in `bootstrap.js`.
- Use [ngBootstrap](https://github.com/ng-bootstrap/ng-bootstrap) which was in the package.json but not implemented.  The downside with this is an increase to the app size and the ng-bootstrap module itself would be a sizeable portion of our app.  The production build remains <1MB, but the size increases by ~60% just by adding the library.

I think we should stick to the jQuery solution for the time being.  We might find we can get a lot of benefit from using ng-bootstrap later, but right now it would just switch one line of code.

src/app/nation/nation.component.html
```diff
@@ -1,61 +1,67 @@
...
-      <planet-form-error-messages [control]="nationForm.controls.type"></planet-form-error-messages>
+</div>
+
+<!-- Modal for adding nations -->
+<div class="modal fade" id="nationAdd" role="dialog" data-backdrop="static">
+  <div class="modal-dialog">
+    <!-- Modal content -->
+    <div class="modal-content">
+      <div class="modal-body">
+        <h3>Add Nation</h3>
```
lmmrssa: have i18n over here

PR #154, opened 2017-11-03
Nation mock data property match (Fixes #149)

### LINKED ISSUE #149: Mock data & form do not match for nation name
Mock data has the nation name as the `name` property, the nation form for adding & displaying on the `/nation` table has it as `nation_name`.

Let's switch the mock data so it matches the form.  Also we should check to make sure the `type` property names match in both.

paulbert:
Changes the `name` property in the nation mock data to `nation_name` to match the `nation.component.ts`.

Note this branch won't pass linting since there were some errors that snuck into the master branch.  Those are fixed in #153.


design/nations/nations-mockup.json
```diff
@@ -5,35 +5,35 @@
 		   	"type": "nation",
 		   	"region": "uganda",
 		   	"admin_name": "ugandabell",
-		   	"name": "UgandaBeLL"
+		   	"nation_name": "UgandaBeLL"
```
lmmrssa: this looks nice for the format but old db seems to use "name" right now. @dogi  do you want to keep it old way or change to new structure?

Issue #162, opened 2017-11-07
Nation : View Button


nirojdyola:
need view button to show the relative nation details.

PR #164, opened 2017-11-07
[#162] Added View Button (connects #162)

nirojdyola:
Added View Button to display nation details.
modal


mappuji:

src/app/nation/nation.component.ts
```diff
@@ -94,4 +95,15 @@ export class NationComponent implements OnInit {
     this.createForm();
   }
 
+  view(id) {
+    if(id){
+      this.couchService.get('nations/' + id)
```
lmmrssa: this data will be taken from another couchdb. which will be another url. so try to fix the link of getting nation detail. Planet will not have nation detail.

src/app/nation/nation.component.ts
```diff
@@ -94,4 +96,15 @@ export class NationComponent implements OnInit {
     this.createForm();
   }
 
+  view(url) {
+    if(url){
+      this.jsonp.request('http://' + url + '/configurations/_all_docs?include_docs=true&callback=JSONP_CALLBACK')  
+      .subscribe(res => {
+        this.view_data = res.json().rows.length > 0 ? res.json().rows[0].doc : [];
+      },(error) => console.log(error));
```
nirojdyola: it needs to hide modal when there is no data which is already done in other PR created by Paul

src/app/nation/nation.component.html
```diff
@@ -65,3 +67,64 @@ <h3 i18n>Add Nation</h3>
     </div>
   </div>
 </div>
+<!-- Modal for view nations -->
+<div class="modal fade" id="view" role="dialog">
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h3 i18n>View Details</h3>
```
lmmrssa: it should be labelled "Nation Detail"

src/app/nation/nation.component.html
```diff
@@ -65,3 +67,64 @@ <h3 i18n>Add Nation</h3>
     </div>
   </div>
 </div>
+<!-- Modal for view nations -->
+<div class="modal fade" id="view" role="dialog">
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h3 i18n>Nation Details</h3>
+        <button type="button" class="close" data-dismiss="modal">&times;</button>
```
lmmrssa: we can put title to button with just image. So hovering over could give the idea of what button does

lmmrssa:
I think we can move these couchdb queries to something shared folder and call it from there to remove duplication.

PR #166, opened 2017-11-07
[#156] Nationcommunitylist and removed repeated function (#connects #156 )

Pr0chin:


mappuji:

src/app/community/community.component.ts
```diff
@@ -51,7 +55,10 @@ export class CommunityComponent implements OnInit {
   }
 
   ngOnInit() {
+    var urlFragment = this.router.url.split('/')[2];
```
LKhadka: I think it's better to use `let` instead of `var`

src/app/community/community.component.ts
```diff
@@ -51,7 +55,10 @@ export class CommunityComponent implements OnInit {
...
+    if(urlFragment != "" && urlFragment != undefined){
```
LKhadka: You could shorten it to `if (!!urlFragment) {` since it's a truthy value. More info: https://basarat.gitbooks.io/typescript/docs/tips/truthy.html

LKhadka: You could shorten it to if (!!urlFragment) { since it's a truthy value. More info: https://basarat.gitbooks.io/typescript/docs/tips/truthy.html

src/app/community/community.component.ts
paulbert: There is also a lint rule for this (see other comment)

src/app/community/community.component.ts
paulbert: @LKhadka That's really useful!
@Pr0chin You should also use `!==` rather than `!=`.  There is a lint rule for this, so please make sure to run the `ng lint` either with the pre push hook or from your terminal and fix the errors it finds.

Pr0chin: ok thanks for your feedback @LKhadka @paulbert

src/app/community/community.component.ts
```diff
@@ -11,8 +11,12 @@ export class CommunityComponent implements OnInit {
   communities = [];
   filter = '';
   selectedValue = '';
+  nationName = '';
+
   constructor(
-    private couchService: CouchService
+    private couchService: CouchService,
+    private url:LocationStrategy,
```
razu9861: this code is giving problem 
screen shot 2017-11-14 at 8 56 00 am

Pr0chin: thanks @razu9861 :+1:

src/app/community/community.component.ts
Pr0chin: @LKhadka  if(!! variable) is not a feasible condition we can imply if(! variable) for truth values.
screeen

src/app/nation/nation.component.html
```diff
@@ -1,7 +1,7 @@
 <div>
   <h1 i18n>Nation List</h1>
   <!-- Trigger the modal with a button -->
-  <button type="button" class="btn btn-info btn-lg" (click)="openNationAddForm()" data-toggle="modal" data-target="#nationAdd">Add</button>
+  <button type="button" class="btn btn-info btn-lg" (click)="createForm()" data-toggle="modal" data-target="#nationAdd">Add</button>
```
Rupesh87: can we just have a i18n translation

paulbert:
Closing this because it duplicates #177

PR #167, opened 2017-11-07
Add delete button again after merge conflict mistake (connects #133)

paulbert:
I added a header for the delete column as well


paulbert:
I'm going to close this since it's included in #170

PR #170, opened 2017-11-08
Alert message component for delete document (connects #135)

paulbert:
A reusable component for a delete popup using Bootstrap modal.

Added a Bootstrap alert to display if there is an error to the user.  I'm using Angular animations here to show how it can work, but I think for the time being for other implementations we should just use ngIf to show/hide Bootstrap alerts.


mappuji:

LKhadka:
I don't think we should use jQuery since direct DOM Manipulation is not really the Angular way to do things.

paulbert:
@LKhadka I don't think using jQuery is ideal, either.  I didn't realize that [ngBootstrap](https://ng-bootstrap.github.io/#/home) allows you to remove the Bootstrap js & jQuery scripts, so I think it might be a net neutral change in terms of app size now.

The downside is that we lose the pre-built animations in Bootstrap because they haven't all been implemented in ngBootstrap (see https://github.com/ng-bootstrap/ng-bootstrap/issues/295), and I don't think we have the bandwidth to implement those ourselves for a little while.

src/app/community/community.component.ts
```diff
@@ -40,14 +43,19 @@ export class CommunityComponent implements OnInit {
...
-  deleteCommunity(communityId, communityRev) {
-    const val = confirm('Are you sure you want to delete it?');
-    if (val) {
-      this.couchService.delete('communityregistrationrequests/' + communityId + '?rev=' + communityRev)
-        .then((data) => {
-          this.getcommunitylist();
-        }, (error) => this.message = 'There was a problem deleting this community');
-    }
+  deleteClick(community, index) {
+    this.deleteItem = { ...community, index: index };
```
lmmrssa: what is this code doing "...community"

src/app/community/community.component.ts
```diff
@@ -40,14 +43,21 @@ export class CommunityComponent implements OnInit {
...
-    if (val) {
-      this.couchService.delete('communityregistrationrequests/' + communityId + '?rev=' + communityRev)
-        .then((data) => {
-          this.getcommunitylist();
-        }, (error) => this.message = 'There was a problem deleting this community');
-    }
+  deleteClick(community, index) {
+    // The ... is the spread operator. The below sets deleteItem a copy of the community
+    // object with an additional index property that is the index within the communites array
+    this.deleteItem = { ...community, index: index };
```
LKhadka: With [the new shorthand property names](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer) in ES6 you could shorten it to `this.deleteItem = { ...community, index};`.

src/app/meetups/meetups.component.spec.ts
```diff
@@ -29,8 +31,8 @@ describe('MeetupsComponent', () => {
     component = fixture.componentInstance;
     couchService = fixture.debugElement.injector.get(CouchService);
     de = fixture.debugElement;
-    meetupdata1 = { id: '1', rev: 'qwrjksf', title: 'happyhangout', description: 'once a week' };
-    meetupdata2 = { id: '2', rev: 'ghjjdrt', title: 'angularhangout', description: 'twice a week' };
+    meetupdata1 = { _id: '1', _rev: 'qwrjksf', title: 'happyhangout', description: 'once a week' };
```
Pr0chin: @paulbert  why this metadata json data's _id and _rev  static ???

src/app/meetups/meetups.component.ts
```diff
@@ -18,10 +21,19 @@ export class MeetupsComponent implements OnInit {
       }, (error) => this.message = 'There was a problem getting meetups');
   }
 
-  deleteMeetup(meetupId, meetupRev) {
+  deleteClick(meetup, index) {
+    // The ... is the spread operator. The below sets deleteItem a copy of the meetup.doc
+    // object with an additional index property that is the index within the meetups array
+    this.deleteItem = { ...meetup.doc, index: index };
```
Pr0chin: How does this this.deleteItem object work ??

PR #171, opened 2017-11-09
Unit Test Library Component

lucy811:
First, I fixed some existing unit test errors below:
1.unit test error in community.component.spec.ts
2.unit test error in nation.component.spec.ts
3.unit test error in nationvalidatorservice.spec.ts
4.unit test error “should make two GET requests to CouchDB for admin”(Init) in users.component.spec.ts
5.import { FormErrorMessagesComponent } from '../shared/form-error-messages.component’; in courses.component.spec.ts
6.unit test error in  courses-add.component.spec.ts

Second, I wrote unit tests in resources.component.spec.ts and resources-view.component.spec.ts
unittest_1

Actually, I have a question about test submitResources() in resources.component.ts, submitResources() contains reader.readAsDataURL(this.file). I wonder when I write unit test for submitResources(), how can I mock (this.file) parameter in reader.readAsDataURL() ?

### DIFF
--- a/src/app/resources/resources-view.component.spec.ts
+++ b/src/app/resources/resources-view.component.spec.ts
@@ -1,0 +1,74 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { RouterTestingModule } from '@angular/router/testing';
+import { CouchService } from '../shared/couchdb.service';
+import { HttpModule } from '@angular/http';
+import { ResourcesViewComponent } from './resources-view.component';
+import { FormsModule }   from '@angular/forms'
+
+describe('ResourcesViewComponent', () => {
+    let component: ResourcesViewComponent;
+    let fixture: ComponentFixture<ResourcesViewComponent>;
+    let postSpy: any;
+    let getSpy: any;
+    let id: string;
+    let couchService;
+    let statusElement;
+    let testdata;
+    let testimage;
+    let testresource1;
+    //let resourceSrc;
+    let de;
+
+    beforeEach(async(() => {
+        TestBed.configureTestingModule({
+          imports: [ FormsModule, RouterTestingModule, HttpModule ],
+          declarations: [ ResourcesViewComponent ],
+          providers: [ CouchService ]
+        })
+        .compileComponents();
+    }));
+
+    beforeEach(() => {
+        fixture = TestBed.createComponent(ResourcesViewComponent);
+        component = fixture.componentInstance;
+        de = fixture.debugElement;
+        couchService = fixture.debugElement.injector.get(CouchService);
+        testdata = { digest: "md5", length: 456321, revpos:1, stub:true };
+        testimage = { filename:"scenery.png", id: "scenery.png", mediaType: 'img', attachments:{ content_type: "application/image", testdata }};
+        testresource1= Object.assign({ 'filename': testimage.filename, '_id': testimage.id, '_attachments':testimage.attachments, mediaType: testimage.mediaType });
+      });
+    
+      it('should be created', () => {
+        expect(component).toBeTruthy();
+      });
+
+      //test getResource()
+      it('should make a get request to couchService', () =>{
+        getSpy = spyOn(couchService, 'get').and.returnValue(Promise.resolve(testresource1._id));
+        component.getResource(testresource1._id);
+        fixture.whenStable().then(() =>{
+          fixture.detectChanges();
+          expect(getSpy).toHaveBeenCalledWith('resources/' + testresource1._id);
+        });
+      });
+
+      it('should getResource', ()=>{
+        statusElement = de.nativeElement.querySelector('img');
+        getSpy = spyOn(couchService, 'get').and.returnValue(Promise.resolve(component.resourceSrc));
+        component.getResource(testresource1._id);
+        fixture.whenStable().then(() =>{
+          fixture.detectChanges();
+          expect(statusElement.textContext).toBe(component.resourceSrc);
+        });
+      });
+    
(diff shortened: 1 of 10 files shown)


mappuji:

paulbert:
To create a mock file in TS you can try this:
```
let blob = new Blob([""], { type: 'text/html' });
blob["lastModifiedDate"] = "";
blob["name"] = "filename";

let fakeF = <File>blob;
```

(from https://stackoverflow.com/questions/24488985/how-to-mock-file-in-javascript)

src/app/resources/resources.component.spec.ts
```diff
@@ -1,30 +1,124 @@
...
-  beforeEach(async(() => {
+  beforeEach((() => {
     TestBed.configureTestingModule({
-      imports: [ FormsModule, RouterModule, HttpModule ],
+     
+      imports: [ FormsModule, RouterTestingModule, HttpModule ],
       declarations: [ ResourcesComponent ],
       providers: [ CouchService ]
     })
     .compileComponents();
```
paulbert: We actually don't need to use `.compileComponents` (I think it's created with the `ng generate` command) since we use Webpack.  That's why we don't need `async()`, either, which I see you've removed.

src/app/resources/resources.component.spec.ts
```diff
@@ -1,30 +1,124 @@
...
+    getSpy = spyOn(couchService, 'get').and.returnValue(Promise.resolve());
+    component.getResources();
+    fixture.whenStable().then(() =>{
+      fixture.detectChanges();
+      expect(getSpy).toHaveBeenCalledWith('resources/_all_docs?include_docs=true');
+    });
+  });
+
+  it('should getResources', ()=>{
+    const resourceRows = de.nativeElement.querySelector('a');
```
paulbert: You should add a class to the element with a `km-` prefix instead of querying by the element type.  If we add another `a` element to this page that will break this test, but with a class it will still be OK.

src/app/resources/resources-view.component.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+        .compileComponents();
+    }));
+
+    beforeEach(() => {
+        fixture = TestBed.createComponent(ResourcesViewComponent);
+        component = fixture.componentInstance;
+        de = fixture.debugElement;
+        couchService = fixture.debugElement.injector.get(CouchService);
+        testdata = { digest: "md5", length: 456321, revpos:1, stub:true };
+        testimage = { filename:"scenery.png", id: "scenery.png", mediaType: 'img', attachments:{ content_type: "application/image", testdata }};
```
paulbert: In your mock data attachments should be an object with one property the same name as the id and containing another object with content_type: 
`attachments: { 'scenery.png': { content_type: "application/image" } }`

src/app/resources/resources-view.component.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+        })
+        .compileComponents();
+    }));
+
+    beforeEach(() => {
+        fixture = TestBed.createComponent(ResourcesViewComponent);
+        component = fixture.componentInstance;
+        de = fixture.debugElement;
+        couchService = fixture.debugElement.injector.get(CouchService);
+        testdata = { digest: "md5", length: 456321, revpos:1, stub:true };
```
paulbert: This isn't needed, so you can remove it.

src/app/resources/resources-view.component.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+        testimage = { filename:"scenery.png", id: "scenery.png", mediaType: 'img', attachments:{ content_type: "application/image", testdata }};
+        testresource1= Object.assign({ 'filename': testimage.filename, '_id': testimage.id, '_attachments':testimage.attachments, mediaType: testimage.mediaType });
```
paulbert: `testimage` & `testresource1` can be combined into one variable.  No need for `Object.assign`

src/app/resources/resources-view.component.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+        component.getResource(testresource1._id);
+        fixture.whenStable().then(() =>{
+          fixture.detectChanges();
+          expect(getSpy).toHaveBeenCalledWith('resources/' + testresource1._id);
+        });
+      });
+
+      it('should getResource', ()=>{
+        statusElement = de.nativeElement.querySelector('img');
+        getSpy = spyOn(couchService, 'get').and.returnValue(Promise.resolve(component.resourceSrc));
```
paulbert: `component.resourceSrc` should be `testimage`

src/app/resources/resources-view.component.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+        getSpy = spyOn(couchService, 'get').and.returnValue(Promise.resolve(testresource1._id));
+        component.getResource(testresource1._id);
+        fixture.whenStable().then(() =>{
+          fixture.detectChanges();
+          expect(getSpy).toHaveBeenCalledWith('resources/' + testresource1._id);
+        });
+      });
+
+      it('should getResource', ()=>{
+        statusElement = de.nativeElement.querySelector('img');
```
paulbert: Add a class with `km-` prefix to the parent `div` tag for this and select:
`querySelector('.km-resource-view img')`

lucy811:
I have revised my code according to suggestions on last Thursday. After I commit my code, It appears a confilicting file. I resolve this confilct in my local machine.  But after I resolve this confilict, it appears many new unit test errors in other components. I need more time to fix these unit test errors this week.

PR #174, opened 2017-11-10
Upgrade to Angular 5 (Fixes #160)

### LINKED ISSUE #160: migrating from angular 4 to 5
here we mostly want find out (on a branch) how good 5 is working

paulbert:
3 main changes:
* Update the `package.json` to install Angular 5 versions of modules (note Angular CLI latest is 1.5.0).  If you are testing this please make sure to remove `node_modules` and reinstall (`rm -rf node_modules` then `npm install`)
* Changed to the new `HttpClient` module.
* Changed the way we import rxjs observables and operators to what is recommended [here](https://github.com/ReactiveX/rxjs/blob/master/doc/lettable-operators.md)


LKhadka:
I think we should also use the `preserveWhitespaces` option in `tsconfig.json` (mentioned [here](https://alligator.io/angular/angular-5/)). It seems to decrease [the bundle size](https://www.reddit.com/r/Angular2/comments/73sxjs/angular_performance_tip_preserve_whitespace/).

PR #175, opened 2017-11-13
Create LICENSE

mappuji:
It is a proposal. Can we have license added to repository to protect our rights? :trollface:


LKhadka:
I agree that there should have a license but should we really be using GNU? I think something more lenient like MIT would be better but I am not sure about it

dogi:
my friend google has a problem with our license https://opensource.google.com/docs/using/agpl-policy/

mappuji:
Yes, we love making problem :trollface:

Issue #155, opened 2017-11-06
Dashboard : Nation and Community view All buttons


Pr0chin:
- [x] show limited data of community registration request and add view all button
- [x] show limited data of nations list and add view all button

PR #176, opened 2017-11-13
[#155] Manager View (connects #155)

Rupesh87:


paulbert:
This module should still live within the `HomeModule`.  There's some repetition of code with the second navigation component (you could instead control which navigation elements are shown through the `navigation.component.ts`).  In addition, we want the `/manager` route inaccessible when the user isn't logged in, so putting it into the `HomeModule` will achieve that.

I also think we need to discuss a bit more how managers will use the app.  This approach pulls them out into what is basically a second "app" (even though the same app is running) with different navigation and home page.  Do we want to work towards something more seamless where the manager can navigate between user and manager views more easily?

src/app/dashboard/dashboard.component.ts
```diff
@@ -6,6 +6,7 @@ import { UserService } from '../shared/user.service';
 @Component({
   template: `
     <div id="greeting">Hi, {{name}}</div>
+    <div style="float:right"><a href="manager" i18n class="btn btn-primary" >Manager View</a></div>
```
paulbert: Let's wait until we have better direction from the designers before styling.

Also please don't use inline styles unless it's absolutely necessary.

src/app/manager-dashboard/manager-dashboard.component.ts
```diff
@@ -0,0 +1,20 @@
+import { Component, OnInit } from '@angular/core';
+import { UserService } from '../shared/user.service';
+
+@Component({
+  selector: 'app-manager-dashboard',
+   template: `
+    <div style="float:right"><a routerLink=".." i18n class="btn btn-primary" >Normal View</a></div>
```
paulbert: Same as above about the `float`

src/app/manager/manager-navigation.component.ts
```diff
@@ -0,0 +1,36 @@
...
+  `,
+  styleUrls: [ './manager-navigation.scss' ]
+})
+export class ManagerNavigationComponent {
+  constructor(
+    private couchService: CouchService,
+    private router: Router
+  ) {}
+
+  components = [
```
paulbert: This is a copy of the `navigation.component.ts` except for this variable.  Rather than copy the component, you can modify the `navigation.component.ts` to switch between the two navigation sets when the buttons are clicked on the dashboard (using some [custom data binding](https://angular.io/guide/component-interaction)) or by removing the button and bringing that into the navigation bar.

src/app/manager/manager-navigation.scss
```diff
@@ -0,0 +1,38 @@
+@import '../variables.scss';
```
paulbert: There are other changes to make with this component to reduce repetition that will fix this.

In the future if you are building a new component that needs the same css rather than coping that file you should reference that file in the component `styleUrls` property.

src/app/manager/manager.component.scss
```diff
@@ -0,0 +1,6 @@
+@import '../variables.scss';
```
paulbert: Same as the other `.scss` file, don't copy just reference the file that you need the styles from.

nirojdyola:
I think we need some settings dropdown where we can change the profile/password and also to check manger settings or manager view.

lmmrssa:
new PR created https://github.com/ole-vi/planet/pull/178

PR #177, opened 2017-11-14
[#156] Nation communitylist (connects #156)

Pr0chin:


Pr0chin:
ang1
ang2

lmmrssa:
what is different between this and PR https://github.com/ole-vi/planet/pull/166/files
Both seems to have same code. Fix issues mentioned in that PR

paulbert:
I went ahead and made a few changes to keep this moving.  You add a `nation` parameter to the `community` route, but don't use that in the component.  I made a change to give an example of how to use that here ([found in theAngular docs](https://angular.io/guide/router#snapshot-the-no-observable-alternative)).

Also the select in the Community component was based on admin_name, so the links weren't showing the correct filter (now fixed)

src/app/nation/nation.component.html
```diff
@@ -11,6 +11,7 @@ <h1 i18n>Nation List</h1>
           <th i18n>Nation Name</th>
           <th i18n>Admin Name</th>
           <th i18n>Nation Url</th>
+          <th i18n>Community List</th>
```
lmmrssa: This one can come along with Action. No need to separate `<td>` just for a button

src/app/community/community.component.ts
```diff
@@ -1,9 +1,11 @@
 import { Component, OnInit, AfterViewInit, ViewChild } from '@angular/core';
 import { Location } from '@angular/common';
-import { Router } from '@angular/router';
+import { ActivatedRoute, ParamMap } from '@angular/router';
 import { CouchService } from '../shared/couchdb.service';
 import { DialogsDeleteComponent } from '../shared/dialogs/dialogs-delete.component';
 import { MatTableDataSource, MatPaginator, MatDialog } from '@angular/material';
+import { switchMap } from 'rxjs/operators';
+declare var jQuery: any;
```
lmmrssa: this is probably not needed here anymore

lmmrssa:
List is not filtered when clicked 'Communities' from the nation list. Also button of 'Communities' does not look like button.

src/app/community/community.component.ts
```diff
@@ -29,7 +30,8 @@ export class CommunityComponent implements OnInit, AfterViewInit {
 
   constructor(
     private couchService: CouchService,
-    private dialog: MatDialog
+    private dialog: MatDialog,
+    private route: ActivatedRoute
   ) {}
 
   ngAfterViewInit() {
```
nirojdyola: Do we need this. There is no code using.

lmmrssa: there is a code but its not of this commit so its not showing.

lmmrssa:
sketch 11

Filter is doing when it match partial of nation url. Do we want it to be like that or complete match?

PR #178, opened 2017-11-14
[#155] Added Manager view (connects #155)

Rupesh87:
screenshot from 2017-11-14 16-21-01

small admin icon in nav bar will redirect to manager dashboard page in which it contain community and nation button from where we can access to nation and community component

### DIFF
--- a/src/app/home/navigation.scss
+++ b/src/app/home/navigation.scss
@@ -17,7 +17,7 @@
   li {
     float: left;
     display: block;
-    width: 12%;
+    width: 11%;
     text-align: center;
   }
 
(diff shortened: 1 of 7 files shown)


Rupesh87:
the work flow has been completed so please check once. Do we want it to be done this way or something different.

src/app/home/navigation.scss
```diff
@@ -17,7 +17,7 @@
   li {
     float: left;
     display: block;
-    width: 12%;
+    width: 11%;
```
nirojdyola: what's the difference between using width 11% and 12% ?

PR #180, opened 2017-11-14
Fix community delete

paulbert:
In #170 I missed a commit to fix the `deleteClick` function in the community component before merging.  This adds in that fix.

Also removing an unnecessary console.log while I was working on this component.

### DIFF
--- a/src/app/community/community.component.ts
+++ b/src/app/community/community.component.ts
@@ -35,9 +35,9 @@ export class CommunityComponent implements OnInit {
   }
 
   deleteClick(community, index) {
-    // The ... is the spread operator. The below sets deleteItem a copy of the community
+    // The ... is the spread operator. The below sets deleteItem a copy of the community.doc
     // object with an additional index property that is the index within the communites array
-    this.deleteItem = { ...community, index };
+    this.deleteItem = { ...community.doc, index };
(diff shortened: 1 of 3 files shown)


src/app/community/community.component.ts
```diff
@@ -29,15 +29,14 @@ export class CommunityComponent implements OnInit {
...
         this.communities = data.rows;
       }, (error) => this.message = 'There was a problem getting Communities');
   }
 
   deleteClick(community, index) {
-    // The ... is the spread operator. The below sets deleteItem a copy of the community
+    // The ... is the spread operator. The below sets deleteItem a copy of the community.doc
     // object with an additional index property that is the index within the communites array
-    this.deleteItem = { ...community, index };
+    this.deleteItem = { ...community.doc, index };
```
Pr0chin: delete works even if we pass community as a parameter as ....community.doc invoke json data of community database.
@paulbert can you elaborate what exactly index does ?? as we are not getting through

paulbert: @Pr0chin The `index` is a good point that I should also change.  This part adds `index` as a property of `deleteItem` so we can later call `communities.splice` to remove the deleted community from the list.  I had overlooked implementing that fully for this component as well as the meetups and courses components.

Issue #183, opened 2017-11-15
More mock data


paulbert:
Similar to the community & nation mock data we should insert mock data for the following tables when we create the database:

* courses
* meetups
* resources

lmmrssa:
We need to have resource files added.
Also we need to link between all mockups

lmmrssa:
looks like we can close this issue now

PR #186, opened 2017-11-20
183 more mock data (connects #183)

Rupesh87:


lmmrssa:
looks like this PR has changes from material-test as well. changes needs to be done from master branch

PR #190, opened 2017-11-21
Change type for docker build (Fixes #189)

### LINKED ISSUE #189: Failed to build angular docker
https://travis-ci.org/ole-vi/planet/builds/305110405#L3336

paulbert:
This built correctly on my machine with `ng build --prod`, and hopefully this will fix the build for the Dockers as well.  I changed it for the two components that had errors as well as two others that have a similar variable in case that would cause these errors in the future.

Not sure exactly why these errors pop up in production mode and not development.  There must be some optimization for Webpack in production that relies on the variable types.

### DIFF
--- a/src/app/community/community.component.ts
+++ b/src/app/community/community.component.ts
@@ -13,7 +13,7 @@ export class CommunityComponent implements OnInit {
   selectedValue = '';
   selectedNation = '';
   nations = [];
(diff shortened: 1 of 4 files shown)


mappuji:
Confirmed https://hub.docker.com/r/treehouses/planet/tags/

PR #191, opened 2017-11-22
Nation validatation for nation Name and Url (connects #191)

razu9861:


design/nations/nation-validators.js
```diff
@@ -0,0 +1,20 @@
+unction (newDoc, oldDoc, userCtx, secObj) {
```
lmmrssa: it should be function 'typo error'

design/nations/nation-validators.js
```diff
@@ -0,0 +1,20 @@
...
+  function require(field) {
+    var errMessage = field + ' is required';
+    if (!newDoc[field]) {
+      throw { forbidden: errMessage };
+    }
+  }
+  function isHex(field) {
```
lmmrssa: why is this validation of hex here?

src/app/nation/nation.component.ts
```diff
@@ -42,20 +42,22 @@ export class NationComponent implements OnInit {
 
   createForm() {
     this.nationForm = this.fb.group({
-      adminName: [ '', Validators.required,
-        // an arrow function is for lexically binding 'this' otherwise 'this' would be undefined
-        ac => this.nationValidatorService.nationCheckerService$(ac)
+      adminName: [ '', Validators.required
```
lmmrssa: it should be on same line or above should be on different lines

src/app/validators/nation-validator.service.ts
```diff
@@ -15,9 +15,9 @@ export class NationValidatorService {
   readonly dbName = 'nations';
 
   constructor(private couchService: CouchService) {}
-    public nationCheckerService$(name: string): Observable<boolean> {
+    public nationCheckerService$(name: string, nationUrl: string): Observable<boolean> {
```
lmmrssa: i don't see this function used anymore.

src/app/validators/nation-validator.service.ts
```diff
@@ -15,9 +15,9 @@ export class NationValidatorService {
...
       const isDuplicate = this.couchService
-      .post(`${this.dbName}/_find`, findOneDocument('name', name))
+      .post(`${this.dbName}/_find`, findAllDocuments('name', name, 'nationUrl', nationUrl))
```
paulbert: This will not work, since `findAllDocuments` only takes two arguments: one `selector` and one `query`.

src/app/validators/nation-validator.service.ts
paulbert: `nationCheckerService` is only passed one `AbstractControl` at a time, so it will not receive both the `name` and `nationUrl`.

src/app/nation/nation.component.ts
```diff
@@ -124,13 +119,8 @@ export class NationComponent implements OnInit, AfterViewInit {
           formdata[ '_id' ] = data.id;
           formdata[ '_rev' ] = data.rev;
           this.nations.data.push(formdata);
-          console.log(this.nations.data);
+          this.nations.data = this.nations.data;
```
lmmrssa: i think it was changed with some other event in one of comment. So why is this here again?

src/app/nation/nation.component.ts
```diff
@@ -146,16 +136,16 @@ export class NationComponent implements OnInit, AfterViewInit {
     const validation = {
       adminName: [ '', Validators.required,
         // an arrow function is for lexically binding 'this' otherwise 'this' would be undefined
-        ac => this.nationValidatorService.nationCheckerService$(ac)
       ],
-      name: [ '', Validators.required ],
-      nationUrl: [ '', Validators.required ],
+      name: [ '', Validators.required, ac => this.nationValidatorService.checkNationExists$(ac) ],
```
lmmrssa: it was supposed to use mango queries selector. but it is still using single field query

src/app/nation/nation.component.ts
```diff
@@ -63,23 +64,18 @@ export class NationComponent implements OnInit, AfterViewInit {
...
-      name: [ '', Validators.required ],
-      nationUrl: [ '', Validators.required ],
-    });
-  }
-
   getNationList() {
     this.couchService.get(this.dbName + '/_all_docs?include_docs=true')
       .then((data) => {
+        const datas = [];
+        for (let i = 0; i < data.rows.length; i++) {
```
lmmrssa: if condition and map can just be joined together to form one map function

src/app/nation/nation.component.ts
Rupesh87: Raju : i kept this code in my pr just to test  sorry i forget to remover it before creating PR

design/nations/nation-validators.json
```diff
@@ -0,0 +1,4 @@
+{
+  "validate_doc_update":
+    "function (newDoc, oldDoc, userCtx, secObj) {\n  function require(field) {\n   var errMessage = field + ' is required';\n    if (!newDoc[field]) {\n      throw { forbidden: errMessage };\n    }\n  }\n  function isHex(field) {\n    var val = newDoc[field];\n    var errMessage = field + ' is not a valid hex';\n\n    if (!/^#[A-F0-9]{6}$/i.test(val)) {\n     throw { forbidden: errMessage };\n    }\n  }\n    if (!newDoc) {\n      require('name');\n      require('adminName');\n      require('nationUrl');\n   }\n}"
```
lmmrssa: this file still has hex function

src/app/nation/nation.component.ts
```diff
@@ -137,15 +141,16 @@ export class NationComponent implements OnInit, AfterViewInit {
     const validation = {
       adminName: [ '', Validators.required,
         // an arrow function is for lexically binding 'this' otherwise 'this' would be undefined
```
lmmrssa: this comment is not supposed to be here

src/app/shared/dialogs/dialogs-form.component.ts
```diff
@@ -23,7 +23,22 @@ export class DialogsFormComponent {
   public validation: any;
   public message: string;
   public modalForm: FormGroup;
+  readonly dbName = 'nations';
```
lmmrssa: this is not supposed to be readonly field as dialog-form could work with any other modal

src/app/nation/nation.component.ts
```diff
@@ -71,7 +73,12 @@ export class NationComponent implements OnInit, AfterViewInit {
       .then((data) => {
         // _all_docs returns object with rows array of objects with 'doc' property that has an object with the data.
         // Map over data.rows to remove the 'doc' property layer
-        this.nations.data = data.rows.map(nation => nation.doc);
+        this.nations.data = data.rows.reduce((nations: any[], nation: any) => {
```
paulbert: My mistake earlier, this is not equivalent to only map, but I think using map & filter will be faster & simpler than reduce:

https://jsperf.com/map-filter-vs-reduce-simple

PR #195, opened 2017-11-23
[#155] Added manager view (connects #155)

Rupesh87:


src/app/manager-dashboard/manager-dashboard.component.ts
```diff
@@ -0,0 +1,15 @@
...
+@Component({
+  template: `
+    <div><a routerLink="/community" i18n mat-raised-button>Communities</a> <a routerLink="/nation" i18n mat-raised-button>Nations</a></div>
+  `
+})
+export class ManagerDashboardComponent implements OnInit {
+
+  constructor() { }
+
+  ngOnInit() {
```
nirojdyola: There is no code in the scope . So we can remove extra line.

lmmrssa: If we are removing ngOnInit then we need to remove implements in the class as well.

lmmrssa: @paulbert As this block is empty shall we remove whole block and `OnInit` implementation as well?

Issue #194, opened 2017-11-23
Add Nation : Fix validator


Pr0chin:

Pr0chin:
![screenshot-2017-11-23 planet learning 1](https://user-images.githubusercontent.com/24644003/33169303-46327118-d06c-11e7-9f5f-36eb735ec3d5.png)
![screenshot-2017-11-23 planet learning 2](https://user-images.githubusercontent.com/24644003/33169340-6fbef326-d06c-11e7-8c22-172ecdf341b3.png)

PR #196, opened 2017-11-24
[#194] Nation validator fixed (connects #194)

Pr0chin:


src/app/nation/nation.component.ts
```diff
@@ -124,7 +111,7 @@ export class NationComponent implements OnInit, AfterViewInit {
           formdata[ '_id' ] = data.id;
           formdata[ '_rev' ] = data.rev;
           this.nations.data.push(formdata);
-          console.log(this.nations.data);
+          this.getNationList();
```
nirojdyola: I think you need to bind the data instead of calling the function.

Issue #197, opened 2017-11-25
Another error introduced while create production artifact


mappuji:
Another error while build prod version.

Reported in https://www.travis-ci.org/ole-vi/planet/builds/305904116#L3065

```
10% building modules 5/8 modules 3 active .../bootstrap/dist/css/bootstrap.min.cssNode#moveTo was deprecated. Use Container#append.
Date: 2017-11-22T17:05:45.732Z
Hash: aebd03020d5fbdf44327
Time: 17528ms
chunk {0} styles.2867e867484647fdb917.bundle.css (styles) 205 kB [initial] [rendered]
chunk {1} polyfills.3bc34265385d52184eab.bundle.js (polyfills) 86 bytes [initial] [rendered]
chunk {2} main.e402deade8b026b7d50e.bundle.js (main) 84 bytes [initial] [rendered]
chunk {3} inline.c72ec803e3bc84a3c0d1.bundle.js (inline) 1.45 kB [entry] [rendered]
ERROR in Error at /ng-app/src/app/shared/dialogs/dialogs-form.component.html(1,18): Property 'onSubmit' does not exist on type 'DialogsFormComponent'.
```

For this problem I'll propose to have git hook when create a commit to prevent problem
in `package.json` add
```
"config": {
    "ghooks": {
      "pre-commit": "ng build --prod",
    }
  },
"devDependencies": {
    "ghooks": "^1.3.2"
}
```

mappuji:
@paulbert @leomaxi

mappuji:
Is it fixed with https://github.com/ole-vi/planet/commit/4382fb82f8d40febaefc9592bef7352dbf181295 ?

Can I closed #227 and #197 ?

paulbert:
Yes, that got it working in Travis so closing this issue.

PR #200, opened 2017-11-28
add githooks for build -- prod

mappuji:
https://github.com/ole-vi/planet/issues/197

How it works:
* when you commit it will try to do `ng buil --prod` if returns with 0 exit code, it will continue to commit, otherwise not.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,10507 @@
+{
+  "name": "ang-bell-app",
+  "version": "0.0.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@angular-devkit/build-optimizer": {
+      "version": "0.0.33",
+      "resolved": "https://registry.npmjs.org/@angular-devkit/build-optimizer/-/build-optimizer-0.0.33.tgz",
+      "integrity": "sha512-qdGAwI3Yd3QIJ4FJsJcAyBVThUVEGKqSm0E3njSVSQkEatjvxHXNDvexBieQDPiUhjYG2Yyobor8nW5EMxPieQ==",
+      "dev": true,
+      "requires": {
+        "loader-utils": "1.1.0",
+        "source-map": "0.5.7",
+        "typescript": "2.4.2",
+        "webpack-sources": "1.1.0"
+      }
+    },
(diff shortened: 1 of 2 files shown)


lmmrssa:
@mappuji I see package-lock.json file being added. Could you remove it. It's not required to be added in commit. Changes is only in package.json

mappuji:
I think it is common practice to put update on lock file. Since it has file about our dependency tree so we don't need to fetch tree again when new people doing `npm install`

paulbert:
@mappuji We do not include a package.lock in the repository because we are using npm version 3.10.10 on our virtual machines.  Package.lock files were introduced with npm version 5.  We can consider upgrading, but that should be done in another issue.

A few other things:
* This solution will not work for those of us--and this should be all of us--using vagrant for our development environment.  We can write a hook that is similar to to the pre-push hook located in `git-hooks` instead.
* Also, do we want this as a pre-commit hook or pre-push?  We made the decision with linting to make it a pre-push hook so developers could commit locally with errors but anything pushed to GitHub would be clean.  We can discuss at the meeting today.

mappuji:
Hi @paulbert 

Thank you for the explanation. I'll delete the lock file.
However for the hooks, I suggest it is a pre-commit hook since we want to make sure no one commit in un-buildable state.

paulbert:
@mappuji Committing an un-buildable state is OK since those changes are only on the developer's local machine.  Pushing is where the problem arises and we will have build failure, and avoiding merging these broken builds with the master is the major goal (it's too bad Travis is passing builds that fail).

It's easier to follow PRs if different blocks of changes are spread across multiple commits with clear commit messages.  If we have to wait 2+ minutes while commits are confirmed it encourages developers to continue to clump all their changes together and write less helpful messages like "Fix changes".  That's why I'd like to have this as a pre push rather than pre commit hook.

mappuji:
@paulbert again, I understand the concern now. Actually my personal default for this kind of hooks is pre-commit, but after reading your explanation I began to understand. After all, you are the one who experienced the development flow.

Could you help me where I can put the hooks? or general direction for that.

paulbert:
Closing this since #227 is working on the same thing

PR #203, opened 2017-11-29
Adding attachment to resources

Rupesh87:


couchdb-setup.sh
```diff
@@ -47,6 +47,24 @@ else
...
+import urllib, json, sys, subprocess
+data = json.load(open('./design/resources/attachment.json'))
+for key in data:
+ id=key['doc_id']
+ image=key['filename']
+ command = 'http://a:a@localhost:2200/resources/'+id
+ response = urllib.urlopen(command)
+ jsondata = json.loads(response.read())
+ rev=jsondata['_rev']
+ putAttachment='curl -v -X PUT http://a:a@localhost:2200/resources/'+id+'/'+image+'?rev='+rev+' --data-binary @'+image+' -H Content-Type:image/jpg'
```
lmmrssa: content-type needs to be dynamic
url of couch should use the constant variable COUCHURL

couchdb-setup.sh
```diff
@@ -47,6 +47,24 @@ else
   COUCHURL=http://$COUCHUSER:$COUCHPASSWORD@$HOST:$PORT
 fi
 
+# Adding attachment to resource
+insert_attachment() {
```
lmmrssa: make function dynamic so that it can take any json file or attach to any db

couchdb-setup.sh
```diff
@@ -47,6 +47,24 @@ else
...
+import urllib, json, sys, subprocess
+data = json.load(open('./design/resources/attachment.json'))
+for key in data:
+ id=key['doc_id']
+ image=key['filename']
+ command = '$COUCHURL/resources/'+id
+ response = urllib.urlopen(command)
+ jsondata = json.loads(response.read())
+ rev=jsondata['_rev']
+ putAttachment='curl -v -X PUT $COUCHURL/resources/'+id+'/'+image+'?rev='+rev+' --data-binary @'+image+' -H Content-Type:image/jpg'
```
paulbert: Let's also make `Content-Type` dynamic so we can add different kinds of files as well

design/resources/resources-mockup.json
```diff
@@ -30,6 +31,7 @@
             "averageRating": ""
```
nirojdyola: averageRating is placed null in all section. We can rate it.

lmmrssa:
looks good but still not sure about location of files. Its uploaded in assets right now.

design/resources/resources-mockup.json
```diff
@@ -30,6 +31,7 @@
...
          },
          {
+            "_id":"0216c7208a72e79b29997d0fd400beac",
```
Pr0chin: IDs are same in all json doc..I think it should be changed.. or not

lmmrssa: ids are not same they look similar but definitely they are different

Pr0chin:
What if we add some audio video attachment for resources ??? @rupesh

Rupesh87:
@razu9861 please review carefully it had been already done

Rupesh87:
@Pr0chin video and audio files are little bit large.

Issue #168, opened 2017-11-07
Code Climate integration


mappuji:
To make us diligent to re-check our code lets have code climate in our system. What do you guys think?

paulbert:
I like the idea.  It looks like Code Climate has TypeScript support in beta, though, and these two options have official support for TypeScript listed on their homepage:
https://codebeat.co/
https://houndci.com/

mappuji:
Let us do some little research about those three options and choose which one to implement.

mappuji:
https://codeclimate.com/github/ole-vi/planet

PR #205, opened 2017-11-29
Fix Code Climate Issues (connects #168)

paulbert:
WIP trying out some fixes for the issues Code Climate has flagged.  Working on the `shared/couchdb.service.ts` first.


paulbert:
There are still duplication issues, but these small changes upgrade the Code Climate "grade" from D to B for that file without having to refactor any of the components that use the CouchService.

Issue #198, opened 2017-11-28
RTL and LTR text


lmmrssa:
We need multi-lingual site, so we need design in both direction of text for (Urdu, Arabic, etc)

paulbert:
This is currently working in a basic sense (no extensive testing)

PR #206, opened 2017-11-30
[#198] Directionality (connects #198)

Pr0chin:


src/app/home/navigation.component.ts
```diff
@@ -7,13 +7,28 @@ import { Router } from '@angular/router';
   selector: 'planet-navigation',
   template: `
     <ul>
+      <li [ngStyle]="{'width': '10%'}">
+        <i *ngIf="selectedIcon"><img [src]="selectedIcon" alt="selectedIcon" style="margin-bottom: -15px"></i>
+        <mat-form-field [ngStyle]="{'width': '60%'}">
+          <mat-select [(ngModel)]="selectedLanguage" (change)="optionSelected($event)" [(value)]="selected">
```
Pr0chin: Value selected needs to be done. When flag is selected it doesn't reflect LTR and RTL direction as expected but works on reload page. So, it is in progress.

src/app/login/login.component.ts
```diff
@@ -21,6 +21,7 @@ export class LoginComponent {
   returnUrl = this.route.snapshot.queryParams['returnUrl'] || '/';
   model = { name: '', password: '', repeatPassword: '' };
   message = '';
+  languageDirection: string = localStorage.getItem('direction');
```
nirojdyola: think this is not needed as direction is binding instead of languageDirection in login component.

src/app/home/navigation.component.ts
```diff
@@ -7,13 +7,28 @@ import { Router } from '@angular/router';
   selector: 'planet-navigation',
   template: `
     <ul>
+      <li [ngStyle]="{'width': '10%'}">
```
lmmrssa: let not do inline styling

src/app/home/navigation.component.ts
```diff
@@ -29,6 +44,26 @@ export class NavigationComponent {
     { link: '', name: 'Feedback' },
   ];
 
+  languages = [
+    { text: 'English', value: 'en', icon: '../../assets/flags/us.png' },
```
lmmrssa: we are using languages array so instead of setting rtl or ltr on condition we can just use it from this array. so there will not be need of condition. Also in future there could be many languages so it need to adjust that as well

src/app/home/navigation.component.ts
Pr0chin: ok :+1:

src/app/home/home.component.ts
```diff
@@ -6,10 +6,14 @@ import { UserService } from '../shared/user.service';
 @Component({
   template: `
     <planet-navigation></planet-navigation>
-    <main class="container">
+    <main class="container" dir="{{languageDirection}}">
       <router-outlet></router-outlet>
     </main>
   `,
```
Rupesh87: extra line

src/app/home/navigation.component.ts
Pr0chin: ok I got it .you mean rtl and ltr should be used as a json value .

Pr0chin:
@paulbert : thanks for your feedback :+1: . Ok I'm on it

PR #151, opened 2017-11-01
New Crowdin translations

olevim:
Enjoy!

PR #207, opened 2017-12-01
[#198 ]Laguage translation with AOT (connects #198)

Pr0chin:
nepali1
nepali

Languages Translation works but it runs with different port ** ng serve --aot --i18nFile=src/locale/messages.fr.xlf --i18nFormat=xlf --locale=fr ** giving specific path.


Pr0chin:
* But arabic it dosen't show any translation. *
* messages..xlf  '/ ' has been changed in to '\' or vice versa as i m running Linux ubuntu in my machine . *

src/main.ts
```diff
@@ -8,4 +8,17 @@ if (environment.production) {
   enableProdMode();
 }
 
-platformBrowserDynamic().bootstrapModule(AppModule);
+// use the require method provided by webpack
+declare const require;
+// we use the webpack raw-loader to return the content as a string
+
+const translations = require(`raw-loader!./i18n/messages.ne.xlf`);
```
lmmrssa: why are these hardcoded

lmmrssa:
why there are src/messages and src/i18n/messages

Pr0chin:
ok i ll remove  src/messages .

src/main.ts
Pr0chin: just to check it works or not as we can make dynamic  later on as per required language translation.

Pr0chin:
@razu9861 i jst put message file in src folder as it is already exists in i18n folder. so I removed that folder.

src/main.ts
lmmrssa: all these are served when you load ng serve with locale mentioned as AOT has defined so why do we need to mention all files and locale here

src/i18n/messages.xlf
```diff
@@ -6,15 +6,15 @@
         <source>
                     Oops!</source>
         <context-group purpose="location">
-          <context context-type="sourcefile">app\page-not-found\page-not-found.component.ts</context>
+          <context context-type="sourcefile">app/page-not-found/page-not-found.component.ts</context>
```
paulbert: I'm also curious why this PR has changed all of the `\` to `/`...

lmmrssa:
@paulbert  could you look after #151 
It's not supposed to be there but to test xlif file from crowdin which only works with master and l10n_master branch, changes has been tested on this old PR. Once you feel changes are good we will create separate PR for it.

PR #209, opened 2017-12-01
6 register/become a member -issue #6

amanuely:


src/app/login/login.component.ts
```diff
@@ -5,77 +5,135 @@ import { Router, ActivatedRoute } from '@angular/router';
...
-    this.router.navigate([ this.returnUrl ]);
-  }
+export class LoginComponent implements OnInit {
+
+ newUser: UserData = new UserData();
+ educationLevels: Array<any>= [ 1, 2, 3, 4, 5, 6 , 7, 8, 9, 11, 12, 'Higher' ];
+ birthmonths: Array<any>= [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 11, 12 ];
+ length: number= new Date().getFullYear();
+ birthyears: number[] = new Array(length);
+ birthdays: number[]= new Array();
```
paulbert: You don't need to use new Array(), you can just assign these to an empty array as `[]`.  Also for `birthyears` you shouldn't assign the length because there isn't really a performance difference with JavaScript.

src/app/login/login.component.ts
```diff
@@ -5,77 +5,135 @@ import { Router, ActivatedRoute } from '@angular/router';
...
-      this.login(this.model);
-    }
-  }
 
-  reRoute() {
-    this.router.navigate([ this.returnUrl ]);
-  }
+export class LoginComponent implements OnInit {
+
+ newUser: UserData = new UserData();
```
paulbert: This is a little redundant since `newUser` will be of type `UserData` since you assigning it as a `new UserData()`

dogi:
@paulbert think yes

there is an interesting new twist:
- olenepal may want to have no login at all
- oleghana is more into optional login
- and ole general want strict login
wonder if can accomodate al of them ;)

anyhow when login then like you suggested

nirojdyola:
You can follow https://material.angular.io/components/form-field/overview for material design

PR #211, opened 2017-12-04
HttpModule to HttpClientModule and removal of bootstrap

razu9861:
screen shot 2017-12-04 at 4 48 32 pm
screen shot 2017-12-04 at 4 51 43 pm

- iam having this eror while doing vagrant up after destroying vagrant


.angular-cli.json
```diff
@@ -19,16 +19,10 @@
...
-        "../node_modules/font-awesome/css/font-awesome.min.css",
         "styles.scss",
         "planet-mat-theme.scss"
       ],
-      "scripts": [
-        "../node_modules/jquery/dist/jquery.min.js",
-        "../node_modules/popper.js/dist/umd/popper.min.js",
-        "../node_modules/bootstrap/dist/js/bootstrap.min.js"
-      ],
+      "scripts": [ ],
```
lmmrssa: if not used then it should be removed

package.json
```diff
@@ -64,6 +64,7 @@
     "ts-node": "~3.2.0",
     "tslint": "~5.3.2",
     "tslint-eslint-rules": "^4.1.1",
-    "typescript": "2.4.2"
+    "typescript": "2.4.2",
+    "webpack": "^3.9.1"
```
lmmrssa: why webpack is added?

razu9861: it says to update webpack/lib/node/NodeTemplatePlugin thats why wepack is added while updating NodeTemplatePlugin

Pr0chin:
I think the things that is needed to be changed are already mentioned  by laxman bro .

package.json
nirojdyola: yes checked with my machine without webpack and it works. So it doesn't need to add.

paulbert:
@razu9861 The HttpModule -> HttpClientModule change was being done in the unit test PR.  Please remove those changes as this task is (1) already finished and merged and (2) very different from the task of removing Bootstrap.

Please update this PR so it only deals with removing Bootstrap.  There is more to do with that including removing all Bootstrap classes from the app that are still left.

Issue #217, opened 2017-12-05
Dialogbox not dismissed in every form


Pr0chin:
Dialog box seems to be appeared in very pages as it is not being dismissed.

Pr0chin:
![dial1.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5860f49dee006e0e0082488f/125516c66e82c728ace21e0d46ec9d9b2fd3c1e5cdf9ad4cae58f3806b1c65267000f72e73d3b529e94422535e5f4eae170c0911f8f88d39baa47e21d40854b088270a36d43703adbf881eba00b8248e4885b1bb834cc8fa70.png)
![dialo2.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5860f49dee006e0e0082488f/125516c66e82c728ace21e0d46ec9d9b2f8dd1ebc3abaf51ab57f5973d047d317c5da77f7586b57ebe147d585d5e4eac430e0b12a9fd882fb4aa2e32db025fb28c341521842f1bbeb98515be04bb228f4a82b6b4894dc7ed2f68.png)

Pr0chin:
This issue should be look forward to very pages that contains dialog box.. like nation, course, resourses.

PR #218, opened 2017-12-05
[#217] Close dialog while clicked outside (connects #217)

razu9861:


src/app/shared/dialogs/dialogs-form.service.ts
```diff
@@ -18,7 +18,7 @@ export class DialogsFormService {
     let dialogRef: MatDialogRef<DialogsFormComponent>;
     dialogRef = this.dialog.open(DialogsFormComponent, {
       width : '600px',
-      hasBackdrop: false
+      hasBackdrop: true
```
nirojdyola: Default it takes true, so doesn;t need to define it .

PR #220, opened 2017-12-06
Update form error messages for Material and translations

paulbert:
Updates the `form-error-messages.component.ts` for use with Material and the i18n translations.

To get the `mat-error` component to display correctly it had to be a direct descendant of `mat-form-field`, so I had to make adjustments to the `courses-add.component` and `dialogs-form.component`.

### DIFF
--- a/src/app/shared/form-error-messages.component.ts
+++ b/src/app/shared/form-error-messages.component.ts
@@ -13,8 +13,7 @@ import { AbstractControl, AbstractControlDirective } from '@angular/forms';
     <span *ngIf="shouldShowError()" i18n>{updateError(), select,
       required {This field is required}
       min {The number cannot be below {{number}}}
-      duplicateCourse {This course already exists}
-      duplicateNation {This nation already exists}
+      duplicate {Value already exists}
       invalidInt {Please enter a number}
       invalidHex {Hex is not valid}
       invalidTime {Time is invalid}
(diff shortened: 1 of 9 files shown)


src/app/shared/form-error-messages.component.ts
```diff
@@ -1,55 +1,50 @@
...
-  styles: [
-    `ul {
-      list-style-type: none;
-      font-size: .8rem;
-      padding-left: .2em;
-    }`
-  ]
+    <span *ngIf="shouldShowError()" i18n>{updateError(), select,
+      required {This field is required}
+      min {The number cannot be below {{number}}}
```
lmmrssa: Here number might not be translated. And some language use different number scripts such as Hindi and Nepali use १ as 1

src/app/shared/form-error-messages.component.ts
```diff
@@ -1,55 +1,50 @@
...
+      duplicateCourse {This course already exists}
+      duplicateNation {This nation already exists}
```
nirojdyola: Both error message are the same . It should be different for Nation name and URL.
screen shot 2017-12-06 at 3 37 44 pm

paulbert: Thanks for catching that.  @lmmrssa and I decided to use a generic duplicate message for all duplication errors at this time to keep things simple.

PR #222, opened 2017-12-06
211 remove bootstrap (connects #211)

razu9861:
I found bootstrap class used in pagenotfound component.

### DIFF
--- a/src/app/home/navigation.component.ts
+++ b/src/app/home/navigation.component.ts
@@ -9,7 +9,7 @@ import { Router } from '@angular/router';
     <ul>
       <li *ngFor="let comp of components"><a [routerLink]="'/' + comp.link">{{comp.name.toUpperCase()}}</a></li>
       <li><a href="#" class="km-logout" (click)="logoutClick()">LOGOUT</a></li>
-      <li><a routerLink="/manager"><i class="fa fa-user" aria-hidden="true"></i></a></li>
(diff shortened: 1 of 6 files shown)


src/app/home/navigation.component.ts
```diff
@@ -9,7 +9,7 @@ import { Router } from '@angular/router';
     <ul>
       <li *ngFor="let comp of components"><a [routerLink]="'/' + comp.link">{{comp.name.toUpperCase()}}</a></li>
       <li><a href="#" class="km-logout" (click)="logoutClick()">LOGOUT</a></li>
-      <li><a routerLink="/manager"><i class="fa fa-user" aria-hidden="true"></i></a></li>
+      <li><a routerLink="/manager"><mat-icon>person</mat-icon></a></li>
```
Pr0chin: Nice but can u resize icon as it has effected in navigation-bar or else there might be material  icon like medium..

razu9861: this PR only deals with removing bootstrap, popper and fontawesome and repllacing fa icon with material icon not ablout styling or designing

Pr0chin: then why did u change code right here navigation.component.ts as it results :
see

razu9861: as far as i was told to keep everything as it is and not to touch designing par or styling part

PR #227, opened 2017-12-06
Pre push hook update for multiple tests

paulbert:
It took 30 minutes to build the app before allowing me to push this.  I'm not sure if we can improve on that, so maybe it would be better to improve our Travis script to output an error when the build fails since that takes less time to run.  In any case please switch to this branch & run `npm run install-hooks` to try it out.

Here's an example of the error output:
image


mappuji:
I'll create prototype to make the builds red when build prod failed.

paulbert:
@mappuji It's been quite a bit faster (builds in 2-3 minutes) than that first time.  I think I may have left an `ng serve` running in the screen the first time which stretches the vagrant's resources.

Now that it's working with a bit more of a reasonable time frame, I think we can merge this so you can focus on the Dockers rather than making this build test run.

mappuji:
I agree @paulbert . Thanks for helping me.

Is it already fix with https://github.com/ole-vi/planet/commit/4382fb82f8d40febaefc9592bef7352dbf181295 ?

paulbert:
I removed the build test from this, but because there were other changes to the hook that will make it easier to add other tests in the future I think we should still merge this.

Issue #202, opened 2017-11-29
make mango queries function to be dynamic


lmmrssa:
change current functions such that you can pass more conditions and fields or other parameters like order, group etc.

PR #228, opened 2017-12-08
[#202] Made function that can take multiple json data/field (connects #202)

Pr0chin:
- [ ] Made function that takes field in one json object as a selector.

### DIFF
--- a/src/app/shared/mangoQueries.ts
+++ b/src/app/shared/mangoQueries.ts
@@ -1,14 +1,6 @@
 // Creates a Mango query to find one document with search based on one field
 export function findOneDocument(selector, query) {
-  return JSON.parse(`
-    {
-      "selector": {
-        "${selector}": "${query}"
-      },
-      "fields": ["${selector}"],
-      "limit": 1
-    }
-  `);
+  return Object.assign(findAllDocuments(selector, query), { limit: 1 });
 }
 
 // Creates a Mango query to find all documents with search based on one field
--- a/src/app/shared/mangoQueries.ts
+++ b/src/app/shared/mangoQueries.ts
@@ -22,10 +22,10 @@ export function findAllDocuments(selector, query) {
   `);
 }
 
-export function findMultiDocuments(selectors, fields, limit, skip) {
+export function findMultiDocuments(selectors, fields, limit=1, skip=0) {
   return JSON.parse(`
     {
-      "selector": ${JSON.stringify(selectors[0])},
+      "selector": ${JSON.stringify(selectors)},
       "fields": ${JSON.stringify(fields)},
       "limit": ${limit},
       "skip": ${skip}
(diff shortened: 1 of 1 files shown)


src/app/shared/mangoQueries.ts
```diff
@@ -21,3 +21,12 @@ export function findAllDocuments(selector, query) {
     }
   `);
 }
+
+export function findMultiDocuments(selector, query) {
```
Rupesh87: can't we make all code in one function

nirojdyola:
Does it test or not ? You can check this function in nation .

src/app/shared/mangoQueries.ts
```diff
@@ -21,3 +21,12 @@ export function findAllDocuments(selector, query) {
...
+  return JSON.parse(`
+    {
+      "selector": { "$and": [{"${selector}": "${query}"}] },
```
lmmrssa: so far it will only take one field and one condition, nothing more

src/app/shared/mangoQueries.ts
Pr0chin: yes we can :+1:

src/app/shared/mangoQueries.ts
Pr0chin: no it takes multiple field ..

lmmrssa: how will it take multiple condition here do have example tested out?

src/app/shared/mangoQueries.ts
```diff
@@ -22,11 +22,21 @@ export function findAllDocuments(selector, query) {
   `);
 }
 
-export function findMultiDocuments(selector, query) {
+export function findMultiDocuments(selectors) {
```
nirojdyola: Satisifies with "AND" condition. But think to change for other operator as well . @lmmrssa , so if yes then we need to send the selectors value as a query format from the requested page.
eg;
selector": {
    "year": {
      "$gt": 2010
    }
  },
if we need >,<, ! operator then this function will not supported. So , for this we need to send the "selectors" value same as of query value.

src/app/shared/mangoQueries.ts
```diff
@@ -21,3 +21,14 @@ export function findAllDocuments(selector, query) {
     }
   `);
 }
+
+export function findMultiDocuments(selectors, fields, limit, skip) {
+  return JSON.parse(`
+    {
+      "selector": ${JSON.stringify(selectors[0])},
```
paulbert: Why are we using the first element in the selectors array?  In general, what is the format we are expecting for the arguments for `findMultiDocuments`?

lmmrssa: it should be json object

lmmrssa:
calling new method will be like 
without limit or skip being optional. It will only return 1 record
`findMultiDocuments({"name": name}, ['name', 'admin_name'])`
with limit and skip you can choose limit or paginate
`findMultiDocuments({"name": name}, ['name', 'admin_name'], 5, 2)`

Issue #231, opened 2017-12-11
Combine course, nation, & resource validators


paulbert:
Currently there is a different validator for courses and nations and there will soon be another for resources to check the uniqueness of a field.  We should instead have one validator to check if a field is unique in the database that can be used across different components.

PR #240, opened 2017-12-13
[#231] combine validators (connects #231)

razu9861:
findmultidocumnets part is left


src/app/shared/validator.service.ts
```diff
@@ -5,34 +5,35 @@ import { Observable } from 'rxjs/Observable';
...
+export class ValidatorService {
 
-  constructor(private couchService: CouchService) {}
+constructor(private couchService: CouchService, private mongoQueries: MongoQueries) {}
 
-  public nationCheckerService$(header: string, name: string): Observable<boolean> {
+  public CheckService$(db: string, field: string, value: string): Observable<boolean> {
     const isDuplicate = this.couchService
-      .post(`${this.dbName}/_find`, findAllDocuments(header, name))
+      .post(`${db}/_find`, this.mongoQueries.findOneDocument(field, value))
```
nirojdyola: Needs to be dynamic like it needs to check which function to use . It may use findOneDocument or findAllDocuments or findMultiDocuments.

src/app/shared/validator.service.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import { switchMap, map } from 'rxjs/operators';
+
+@Injectable()
+export class ValidatorService {
+
+constructor(private couchService: CouchService, private mongoQueries: MongoQueries) {}
+
+  public CheckService$(findType: string, db: string, field: string, value: any): Observable<boolean> {
+    if (findType === 'findOneDocument') {
```
paulbert: For uniqueness we will always use the `findOneDocument` query, so we do not need to use `findType` in this function

src/app/shared/validator.service.ts
```diff
@@ -15,25 +15,34 @@ export class ValidatorService {
...
-    ac: AbstractControl
+    ac
   ): Observable<ValidationErrors | null> {
-    // mangoQueries.findOneDocument(1, 2)
+       // mangoQueries.findOneDocument(1, 2)
     // calls service every .5s for input change
     return timer(500).pipe(
       switchMap(() => {
-        return this.CheckService$(dbName, fieldName, ac.value).pipe(map(res => {
+        return this.CheckService$(findType, dbName, fieldName, ac).pipe(map(res => {
```
lmmrssa: isExist will always be findOne as anyone occurrence will make it duplicate

PR #244, opened 2017-12-13
[#243] Add search and pagination (connects #243)

Pr0chin:
Pagination section is left. so this issue is still in progress.

### DIFF
--- a/src/app/resources/resources.component.html
+++ b/src/app/resources/resources.component.html
@@ -1,9 +1,21 @@
 <!-- FOR DEMO ONLY -->
 <!-- use the async pipe when fetching for database -->
-<button mat-raised-button routerLink="add" color="primary" routerLinkActive="active" i18n>Add Resource</button>
+<div>
+  <button mat-raised-button routerLink="add" color="primary" routerLinkActive="active" i18n>Add Resource</button>
+  <span class="search" [ngStyle]="{'float': 'right'}">
+    <mat-icon>search</mat-icon>
+    <mat-form-field class="font-size-1">
+      <input matInput [(ngModel)]="searchResource" i18n-placeholder placeholder="Type text to search...">
+    </mat-form-field></span>
+</div>
+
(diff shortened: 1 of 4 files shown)


src/app/resources/resources.component.html
```diff
@@ -1,9 +1,21 @@
...
+</div>
+
+<div class="mat-paginator-container" [ngStyle]="{'justify-content': 'flex-start'}">
+ <mat-paginator [pageSize]="pageSize" [pageSizeOptions]="pageSizeOptions" (page)="pageEvent = $event"></mat-paginator>
+</div>
+
 <ng-template ngFor let-item [ngForOf]="resources">
   <!-- Used ng-template insted of spans or divs because I did not want any to add any element -->
-  <div class="list-item">
+  <div class="list-item" *ngIf="item.doc.title.toLowerCase().includes(searchResource.toLowerCase())" >
```
lmmrssa: material has its own filter for filtering data from record array so use that

Pr0chin: well I m not able to use material features as filter is use ur mat-table..as it has not been use in resource.

src/app/resources/resources.component.html
```diff
@@ -18,10 +30,10 @@ <h3>{{ item.doc.title }}</h3>
         </ng-template>
       </div>
       <div class="list-item-action">
-        <a mat-raised-button i18n [routerLink]="['/resources/view', item.doc._id]">View Details</a>
-        <a mat-raised-button i18n [routerLink]="['/resources/wip']">Add To Library</a>
-        <a mat-raised-button i18n [routerLink]="['/resources/wip']">Feedback</a>
-        <a mat-raised-button i18n [routerLink]="['/resources/wip']">Delete</a>
+        <a mat-raised-button color = "primary" i18n [routerLink]="['/resources/view', item.doc._id]">View Details</a>
```
nirojdyola: It should be color="primary" , no space . Could you please check it once ?

Pr0chin:
Had made reference branch #250 .. so Guys you can go through it :+1:

Issue #245, opened 2017-12-13
remove deprecated @angular/http


lmmrssa:
@angular/http has been deprecated so we need to use @angular/common/http
All Modules that are import from this old package needs to be converted to new one.
Eg: jsonp, headers, response

lmmrssa:
closing as it seems to have already been fixed

PR #249, opened 2017-12-14
[#245] deprecated @angular/http removed (connects #245)

razu9861:


src/app/home/home.module.ts
```diff
@@ -17,6 +17,7 @@ import { MaterialModule } from '../shared/material.module';
 import { PlanetDialogsModule } from '../shared/dialogs/planet-dialogs.module';
 import { ManagerDashboardComponent } from '../manager-dashboard/manager-dashboard.component';
 
+
```
lmmrssa: why it has extra line

PR #250, opened 2017-12-14
[#243] Mat Resource View (connects #243 )

Pr0chin:


paulbert:
Made some changes because there was a big chunk of copied code in `style.scss` to simply separate them from the `.list-item` parent which could just as easily be applied to the `mat-row` element.  Please do not duplicate code like this.  If you are having trouble finding a different solution, you can ask someone for help in the PR or on Gitter.

In addition I went through and simplified some of the styles and fixed the delete function so the deleted resource would be removed from the list.

src/app/resources/resources.component.html
```diff
@@ -1,52 +1,56 @@
...
-      <h3>{{ item.doc.title }}</h3>
+<div>
+  <button mat-raised-button routerLink="add" color="primary" routerLinkActive="active" i18n>Add Resource</button>
+</div>
+<div>
+  <mat-form-field>
+    <input matInput (keyup)="applyResFilter($event.target.value)" i18n-placeholder placeholder="Search">
+  </mat-form-field>
+</div>
+<mat-table #table [dataSource]="resources" matSort>
```
nirojdyola: matSort is not need. Could you please remove this ?

src/app/resources/resources.component.html
```diff
@@ -1,52 +1,56 @@
...
-          <div [ngStyle]="{ 'background-color': 'gold', 'height': '8px', 'width': fRating + '%'}"></div>
+        <div class="label">
+          <span class="small">Rating ({{ rating }})</span>
+          <span class="small">Female ({{ fRating }}%)</span>
+          <span class="small">Male ({{ mRating }}%)</span>
         </div>
-        <div class="rating-bar">
-          <div [ngStyle]="{ 'background-color': 'gold', 'height': '8px', 'width': mRating + '%'}"></div>
+        <div class="info">
+          <span> {{ rating }}
```
nirojdyola: should be in a single line. Closing tag of span is in another line , that needs to be fix. Please check it once .

src/app/resources/resources.component.ts
```diff
@@ -63,7 +74,7 @@ export class ResourcesComponent implements OnInit {
       const { _id: resourceId, _rev: resourceRev } = resource;
       this.couchService.delete('resources/' + resourceId + '?rev=' + resourceRev)
```
nirojdyola: 'resources' can be change to this.dbName

src/app/resources/resources.component.html
```diff
@@ -1,52 +1,58 @@
-<!-- FOR DEMO ONLY -->
-<!-- use the async pipe when fetching for database -->
-<button mat-raised-button routerLink="add" color="primary" routerLinkActive="active" i18n>Add Resource</button>
-<ng-template ngFor let-item [ngForOf]="resources">
-  <!-- Used ng-template insted of spans or divs because I did not want any to add any element -->
-  <div class="list-item">
-    <div class="list-item-info">
-      <h3>{{ item.doc.title }}</h3>
+<div>
+  <button mat-raised-button routerLink="add" color="primary" routerLinkActive="active" i18n>Add Resource</button>
```
lmmrssa: routerlinkactive not required

PR #252, opened 2017-12-15
Change CouchService to use observables (fixes #246)

### LINKED ISSUE #246: Change CouchService to use Observables

paulbert:
Converts CouchService and all applicable components to use observables as part of RxJS (reactive programming for JavaScript)

Fixes #246


lmmrssa:
PR is good to go but it will conflict with our other PR so we need to fix conflict before we merge

LKhadka:
I think you are forgetting to unsubscribe from the subscriptions.

paulbert:
@LKhadka From my research you don't need to explicitly unsubscribe from most observables used with Angular.  In particular, the Http & routeParams methods all (or should) unsubscribe on their own.

Just to be safe, I added a `debug` operator which should log (and does in all but a few cases) when subscriptions start and complete.  Based on that I manually added an unsubscribe for the routeParams because it wasn't logging an end point, but still had to manually add a complete function for the subscription to see a confirmation that it is ending.

The one http call that isn't ending is the AuthService route guard.  Since we don't manually add a subscribe for that, I'm not sure where to add an unsubscribe or if we should even worry about it.

LKhadka:
@paulbert oh ya I gotcha Paul. I seem to have confused it with handling subscriptions in the components.

LKhadka:
The subscriptions are fine. I think we need to be unsubscribing on the component level with `ngOnDestroy()` cycle with the `takeUntil()` operator. Something like this https://stackoverflow.com/a/41177163/8705692

paulbert:
@LKhadka Yes I understand that is the recommended method to unsubscribe.  My point is that HTTP methods complete & unsubscribe automatically (no need to unsubscribe) and routeParams subscriptions should do the same (but we'll unsubscribe to be safe since there are concerns they do not).  The debug operator logs when an observable is subscribed to and when it subsequently completes/unsubscribes, confirming completion for all HTTP calls except for the one found in AuthService.

HTTP info: https://stackoverflow.com/questions/35042929/ist-it-necessary-to-unsubscribe-from-observables-created-by-http-methods

routeParams info: https://github.com/angular/angular.io/issues/3003

Issue #260, opened 2017-12-19
Look into linting for HTML and SCSS as well


paulbert:
Note the HTML will have to have at least some support for special Angular attributes and other syntax.

PR #261, opened 2017-12-20
[#260] scss linting (connects #260)

razu9861:
I have tried to implement scss lint rule and  i tried one some scss file, 
I have added https://www.npmjs.com/package/htmlhint-ng2 for HTML lint.


src/app/login/login.scss
```diff
@@ -25,30 +25,29 @@
   }
 
   label {
-    color: white;
+    color: #ffffff;
   }
 
   .ole-logo {
     margin-top:-200px;
```
nirojdyola: no space after :

src/app/login/login.scss
```diff
@@ -2,13 +2,13 @@
 
 .ole-login {
 
-  height:100vh;
-  width:600px;
+  height: 100vh;
+  width: 600px;
```
nirojdyola: I found double space after ":" . Could you please check it once ?

package.json
```diff
@@ -10,7 +10,8 @@
     "v-build": "vagrant ssh -c 'cd /vagrant;ng build'",
     "test": "ng test",
     "v-test": "vagrant ssh -c 'cd /vagrant;ng test'",
-    "lint": "ng lint",
+    "saas-lint": "./node_modules/sass-lint/bin/sass-lint.js -c ./sass-lint.yml -v -q",
```
lmmrssa: it should try to follow same structure as other run works

src/app/login/login.scss
lmmrssa: it should be taken care by linting itself

src/app/courses/request-courses/courses-request.component.ts
```diff
@@ -1,7 +1,6 @@
 import { Component } from '@angular/core';
 
 @Component({
-  templateUrl: 'courses-request.component.html',
-  styleUrls: [ 'courses-request.component.scss' ]
```
lmmrssa: why is this file removed? This issue is supposed to be for linting only.

sass-lint.yml
```diff
@@ -0,0 +1,96 @@
...
+  no-color-keywords: 1
+  no-color-literals: 1
+  no-combinators: 0
+  no-css-comments: 1
+  no-debug: 1
+  no-disallowed-properties: 0
+  no-duplicate-properties: 1
+  no-empty-rulesets: 1
+  no-extends: 0
+  no-ids: 1
```
lmmrssa: why ids and important are not allowed

package.json
```diff
@@ -13,7 +13,7 @@
     "html-lint": "./node_modules/html-linter/bin/html-linter.js --config html-linter.json",
     "sass-lint": "./node_modules/sass-lint/bin/sass-lint.js -c ./sass-lint.yml -v -q",
     "lint": "npm run sass-lint && ng lint",
-    "v-lint": "vagrant ssh -c 'cd /vagrant;ng lint;npm run sass-lint'",
+    "v-lint": "vagrant ssh -c 'cd /vagrant;ng lint;npm run sass-lint; npm run html-lint'",
```
nirojdyola: There is a space before npm run html-lint

paulbert:
@razu9861 I don't think this HTML linting library is the right one for us to use.  It looks like it requires another package just to run it from the cli, and it would be more ideal to use a tool which works with Angular 2/4/5.

From my short research, it seems like the only option for Angular is [htmlhint-ng2](https://github.com/nikolasleblanc/HTMLHint-ng2), but we could do some more digging to see if there is a better option.

razu9861:
@paulbert i have added [htmlhint-ng2](https://github.com/nikolasleblanc/HTMLHint-ng2)

lmmrssa:
@paulbert  could you look after html linting on it

paulbert:
Looking at the output from htmlhint-ng2, it isn't actually doing anything Angular specific, which is annoying.

By running something that doesn't work with Angular syntax we'll have to shut off many useful rules, but it's better than nothing.

PR #263, opened 2017-12-21
6 become member  #6

amanuely:
Please check localhost:3000/register to see the view


src/app/register/register.component.html
```diff
@@ -0,0 +1,82 @@
...
+      <mat-error><planet-form-error-messages [control]="registerForm.controls.middleName"></planet-form-error-messages></mat-error>
+    </mat-form-field>
+    <mat-form-field>
+      <input matInput i18n-placeholder placeholder="Last Name" formControlName="lastName" required>
+      <mat-error><planet-form-error-messages [control]="registerForm.controls.LastName"></planet-form-error-messages></mat-error>
+    </mat-form-field>
+  </mat-grid-list>
+  <mat-grid-list cols="3">
+    <mat-form-field>
+      <input matInput i18n-placeholder placeholder="Login" formControlName="login" required>
```
razu9861: please do check username, and make user name unique, same user name can't be register twice.

nirojdyola: @amanuely You can use ValidatorService(validator.service.ts) to check unique field.

src/app/register/register.component.ts
```diff
@@ -0,0 +1,92 @@
...
+
+@Component({
+  selector: 'app-register',
+  templateUrl: './register.component.html',
+  styleUrls: [ './register.component.scss' ]
+})
+export class RegisterComponent implements OnInit {
+
+  registerForm: FormGroup;
+  educationLevel = [ 1, 2, 3, 4, 5, 6 , 7, 8, 9, 11, 12, 'Higher' ];
```
nirojdyola: 1-12 are number and one is string . Could you please change it all to the same type ?

src/app/register/register.component.ts
```diff
@@ -0,0 +1,92 @@
...
+  RegistrationMsg: String;
+
+  constructor(
+    private couchService: CouchService,
+    private fg: FormBuilder
+    ) {
+    this.createform();
+  }
```
nirojdyola: Could you please check indentation?

src/app/register/register.component.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  ngOnInit() {
```
nirojdyola: No code is used inside this function. If it is not needed , please remove it.

src/app/register/register.component.ts
```diff
@@ -0,0 +1,92 @@
...
+      firstName: [ '', Validators.required ],
+      middleName: [ '', Validators.required ],
+      lastName: [ '', Validators.required ],
+      login: [ '', Validators.required ],
+      email: [ '', Validators.required ],
+      password: [ '', Validators.required ],
+      repeatPassword: [ '', Validators.required ],
+      language: [ '', Validators.required ],
+      phoneNumber: [ '', Validators.required ],
+      birthDay: [ '', Validators.required ],
```
nirojdyola: Don;t we need birthMonth and birthYear ?

src/app/register/register.component.html
```diff
@@ -0,0 +1,82 @@
...
+      <mat-error><planet-form-error-messages [control]="registerForm.controls.password"></planet-form-error-messages></mat-error>
+    </mat-form-field>
+    <mat-form-field>
+      <input  type="password" matInput i18n-placeholder placeholder="Re-enter password" formControlName="repeatPassword" required>
+      <mat-error><planet-form-error-messages [control]="registerForm.controls.repeatPassword"></planet-form-error-messages></mat-error>
+    </mat-form-field>
+  </mat-grid-list>
+  <mat-grid-list cols="3">
+    <mat-form-field>
+      <input matInput i18n-placeholder placeholder="Email" formControlName="email" required>
```
razu9861: the input field email should only take valid email example : example@example.com
Please check  https://material.angular.io/components/input/examples for reference

src/app/register/register.component.html
amanuely: @nirojdyola  I tried to use ValidatorService, but it is not working for me. I followed the example in course_add_component.ts. Could you please have a look at the file and see where I am making mistakes? Thank you!!

nirojdyola: @amanuely It can be used for _users . So what you can do is look createUser function of login component . This function throws error when same login name is added. You can check with browser inspect element, so similarly you can do it here. You can catch this error for login name unique.

amanuely: @nirojdyola  Thank you. That is very useful.

Issue #266, opened 2017-12-22
Profile Page


lmmrssa:
Create a profile page where user can see own profile.
And also able to update with more details
* Name
* Gender
* Language
* Picture, etc

PR #272, opened 2017-12-25
[#266] User profile view (connects #266)

Rupesh87:


src/app/users/user-profile/user-profile.component.html
```diff
@@ -0,0 +1,54 @@
+<h3 i18n>Member Profile</h3>
+<form [formGroup]="editForm" (ngSubmit)="onSubmit()">
+  <mat-grid-list cols="3">
```
paulbert: Please use `<div>` tags instead.  The `mat-grid-list` isn't doing anything without `mat-grid-tile` as child components, and the layout will not be good without doing a lot of work to tweak the grid attributes.

src/app/users/user-profile/user-profile.component.html
```diff
@@ -0,0 +1,54 @@
...
+    <mat-form-field>
+      <input matInput i18n-placeholder placeholder="First Name" formControlName="firstName" required>
```
paulbert: You should use `planet-form-error-messages` here like in the `courses-add` component.

nirojdyola: @Rupesh87  Could you please use error message ?

lmmrssa:
screen shot 2018-01-02 at 10 11 09
Here it seems user can repeat for admin and user.
Username needs to be unique so can this be checked when we create user

paulbert:
Made a few updates.  I didn't foresee us uploading attachments for databases other than `resources`, so my mistake not adding the `prepAttachment` function to a shared service.

I don't want to cause more merge conflicts by fixing the `resources-add.component` to use that shared function, but we should update that once we've merged most of the backlog.

src/app/home/home.module.ts
```diff
@@ -33,10 +31,9 @@ import { ManagerDashboardComponent } from '../manager-dashboard/manager-dashboar
     HomeComponent,
     DashboardComponent,
     NavigationComponent,
-    UsersComponent,
     CommunityComponent,
     NationComponent,
-    ManagerDashboardComponent
+    ManagerDashboardComponent,
```
lmmrssa: why is there ',' at the end of array list

Issue #265, opened 2017-12-22
Template for Library change based on new designs


lmmrssa:
https://drive.google.com/drive/u/3/folders/17A4y0JfF3VMU2eFywLxUQBduY8A_AKuB
https://drive.google.com/drive/u/3/folders/18HbmjgcgssiWrkN0rLaKI5eS-m519JfA

PR #273, opened 2017-12-25
[#265] template view for resources (connects #265)

razu9861:
Tried to implement new design for Resources still more things are left to do.


src/styles.scss
```diff
@@ -46,14 +46,15 @@ body {
...
-    display: grid;
+    display: flex;
     grid-template-columns: auto 140px;
     grid-gap: 2em;
 
     .list-item-info {
       display: flex;
       flex-direction: column;
       justify-content: space-around;
+      width:900px;
```
lmmrssa: let's not have fixed width make it to % or some relative value

src/app/resources/resources.component.ts
```diff
@@ -4,12 +4,22 @@ import { DialogsDeleteComponent } from '../shared/dialogs/dialogs-delete.compone
 import { MatTableDataSource, MatPaginator, MatFormField, MatFormFieldControl, MatDialog, MatDialogRef } from '@angular/material';
 
 @Component({
-  templateUrl: './resources.component.html'
+  templateUrl: './resources.component.html',
+  styles: [ `
+    .mat-cell, .mat-header-cell {
```
paulbert: Please **do not modify existing Material classes**.  This makes the templates difficult to debug and may cause unintended side effects.

src/styles.scss
```diff
@@ -46,14 +46,15 @@ body {
 
   // STYLING FOR TABLE LIST ENTRIES
   .list-item {
-    display: grid;
+    display: flex;
```
paulbert: I think it will be easier to implement this design by using CSS grid rather than flexbox.

src/app/resources/resources.component.html
```diff
@@ -1,25 +1,66 @@
...
+    <mat-icon>library_add</mat-icon><span i18n>Add to Library</span>&nbsp;&nbsp;
+    <mat-icon>feedback</mat-icon><span i18n>Feedback</span>&nbsp;&nbsp;
+    <mat-icon>delete_forever</mat-icon><span i18n>Delete</span> &nbsp;&nbsp;
+    <button mat-mini-fab routerLink="add" ><mat-icon>add</mat-icon></button>
+  </mat-toolbar-row>
+</mat-toolbar>
+
 <mat-table #table [dataSource]="resources">
+  <ng-container matColumnDef="select">
+    <mat-header-cell *matHeaderCellDef>
```
paulbert: There is no header so this can be removed (the `mat-header-row` element is required for all tables, but not `mat-header-cell`)

src/app/resources/resources.component.html
```diff
@@ -1,25 +1,66 @@
...
+    </mat-header-cell>
+    <mat-cell *matCellDef="let row" class="list-item-select">
+      <mat-checkbox></mat-checkbox>
```
paulbert: The checkbox vertical alignment is off after being clicked, possibly due to the styles added.

lmmrssa:
@paulbert there is no final design right now we are just implementing popup actionbar and positioning of items based on the design you mentioned. We will have more specific design this week. If you have any comment of preferences then you can mention that.

src/app/resources/resources.component.html
```diff
@@ -1,25 +1,72 @@
...
+        </mat-checkbox>
+    </mat-cell>
+  </ng-container>
   <ng-container matColumnDef="title">
     <mat-header-cell *matHeaderCellDef></mat-header-cell>
-    <mat-cell *matCellDef="let element" class="list-item-info">
+    <mat-cell *matCellDef="let element">
       <h3>{{element.title}}</h3>
       <div class="list-item-content">
         Author: {{ element.author }},Year: {{ element.year }}, Media: {{ element.Medium }}, Language: {{ element.language }}, Subject/s: {{element.subject}}, Publisher/Attribution: {{ element.Publisher }}, Link to License: {{element.linkToLicense}}, Resource: {{ element.resourceFor }}, Resource Type: {{ element.resourceType }}
```
yaoxing1990: Can you put the content of this paragraph in a form to make it more beautiful?

yaoxing1990:
@razu9861 Please fix travis error

src/app/resources/resources.component.html
```diff
@@ -1,25 +1,72 @@
...
+
+  <mat-toolbar-row class="primary-color font-size-1">
+    <mat-checkbox (change)="$event ? masterToggle() : null"
+      [checked]="selection.hasValue() && isAllSelected()"
+      [indeterminate]="selection.hasValue() && !isAllSelected()" i18n>Click to select all
+    </mat-checkbox>
+    <span class="toolbar-fill"></span>
+    <mat-icon>folder</mat-icon><span i18n>View Details</span>&nbsp;&nbsp;
+    <mat-icon>library_add</mat-icon><span i18n>Add to Library</span>&nbsp;&nbsp;
+    <mat-icon>feedback</mat-icon><span i18n>Feedback</span>&nbsp;&nbsp;
```
yaoxing1990: The above 3 links seems not functional. Are they just for decoration?

src/app/resources/resources.component.html
```diff
@@ -51,8 +98,8 @@ <h3>{{element.title}}</h3>
       </div>
     </mat-cell>
   </ng-container>
-  <mat-header-row *matHeaderRowDef="displayedColumns"></mat-header-row>
-  <mat-row *matRowDef="let row; columns: displayedColumns;" class="list-item"></mat-row>
+  <mat-header-row *matHeaderRowDef="displayedColumns" class="example-header"></mat-header-row>
```
nirojdyola: If class is not used, then it can be remove. Could you please remove it ?

src/app/resources/resources.component.ts
```diff
@@ -76,4 +117,23 @@ export class ResourcesComponent implements OnInit, AfterViewInit {
     };
   }
 
+  deleteSelected() {
+    const DeleteAll = confirm('Are you sure you want to delete?');
```
nirojdyola: @razu9861  I think we can reuse the existing dialog delete componet. So, could you please check it once ? And also for deleting one item , the message is display "Are you sure you want to delete?" I think for deleting one item it should be the same as we have done in other component (Eg; nation) . For deleting All or more than one item , it should display "Are you sure you want to delete all ? " or should display the deleting item name .Please check it once.

paulbert:
I made some changes which we can revert in the last commit to approximate the inner box with the table and gradient background.

Even if we like that, we should redo the implementation so it will apply across the app.

src/app/resources/resources.component.html
```diff
@@ -1,58 +1,104 @@
...
+      <input matInput (keyup)="applyResFilter($event.target.value)" i18n-placeholder placeholder="Type text to search...">
+    </mat-form-field>
+  </mat-toolbar-row>
+</mat-toolbar>
+
+<div class="space-container">
+  <mat-toolbar>
+    <mat-toolbar-row class="primary-color font-size-1">
+      <mat-checkbox (change)="$event ? masterToggle() : null"
+        [checked]="selection.hasValue() && isAllSelected()"
```
paulbert: This seems redundant.  Why is there the check on `selection.hasValue` in addition to `isAllSelected`?

src/styles.scss
```diff
@@ -44,102 +44,121 @@ body {
...
-        margin-bottom: 0.5em;
-      }
+  }
+
+  // the colored stars are placed on top of the uncolored ones
+  // the width is set dynamically with JS
+  .stars {
+    display: inline-block;
+    font-size: 1.5rem;
+    color: gray;
```
paulbert: Please revert this to the Sass variable `$grey`.  We should make sure to use variables for colors so if wee change theme we only need to update the variable values.

src/styles.scss
```diff
@@ -44,102 +44,121 @@ body {
...
+    position: relative;
+    grid-area: stars;
```
paulbert: I included this in the `.list-item-rating .stars` style so if we use this component in another situation we don't need to use this `grid-area` value.  Please remove from here.

src/styles.scss
```diff
@@ -44,102 +44,121 @@ body {
...
-      .list-item-content {
-        margin-bottom: 0.5em;
-      }
+  }
+
+  // the colored stars are placed on top of the uncolored ones
+  // the width is set dynamically with JS
+  .stars {
+    display: inline-block;
+    font-size: 1.5rem;
```
paulbert: Please remove.  With the new design the stars display better without the increased font size.

paulbert:
The white around the paginator at the bottom should not extend to the edges of the screen.  Something is also causing a horizontal scrollbar to appear, which should not be happening.

razu9861:
@paulbert it was caused by the background style position absolute now the style is changed to background position fixed

PR #277, opened 2017-12-26
Fix npm commands & pre push hook

paulbert:
After the Vagrant upgrade the npm commands with `vagrant ssh -c` stopped working.

They all work now instead using `vagrant ssh -- -t`


lmmrssa:
tested few of the scripts and it seems to work

package.json
```diff
@@ -5,16 +5,16 @@
   "scripts": {
     "ng": "ng",
     "start": "LNG=${LNG:-en};ng serve --aot --i18n-file=src/i18n/messages.$LNG.xlf --locale=$LNG --i18n-format=xlf",
-    "v-serve": "vagrant ssh -c 'cd /vagrant;ng serve'",
+    "v-serve": "vagrant ssh -- -t 'cd /vagrant;ng serve'",
     "build": "ng build",
-    "v-build": "vagrant ssh -c 'cd /vagrant;ng build'",
+    "v-build": "vagrant ssh -- -t 'cd /vagrant;ng build --prod'",
```
lmmrssa: we have --prod here but not on "build" why is it so?

paulbert: I added that in because generally we run the build command with the prod flag as a check to make sure there aren't errors when building the app.  I have not run `ng build` without the prod flag at all.

Issue #287, opened 2018-01-03
Multiple language to switch between languages


lmmrssa:
- [x] Make a json file which can be used by installation script and also from the planet application itself
- [ ] File should follow the structure of language
  ```
  short_code: *Format ISO 639-3
  name:
  flag: *Graphical image
  served_url:
  active:
  ```
- [ ] Make a language switch option within application. 
_For now it should just give the option to click but not actually translate, later it will switch to different application AOT based on language_

PR #297, opened 2018-01-05
[#287] Added mockup data (connects #287)

nirojdyola:
mock file has been added.
WIP: css in navigation for language switch and what to display in navigation for language switch is left.


paulbert:
@nirojdyola This issue is only for the menu to switch languages, since with AOT compilation we will switch to other languages by loading a new app.

We will implement changes for RTL languages, if necessary, with a different issue.

lmmrssa:
@nirojdyola  can we not use language database and read detail from json file itself?

src/app/home/navigation.component.ts
```diff
@@ -2,22 +2,42 @@ import { Component, OnInit } from '@angular/core';
 import { UserService } from '../shared/user.service';
 import { CouchService } from '../shared/couchdb.service';
 import { Router } from '@angular/router';
+import * as languages from '../../../design/languages/languages-mockup.json';
```
lmmrssa: this file can go to assets or shared as it is not mockup data

design/languages/languages-mockup.json
```diff
@@ -0,0 +1,30 @@
...
+        "flag": "assets/flags/np.png",
+        "served_url": "",
+        "active": "N"
+    },
+    {
+        "short_code": "fr",
+        "name": "French",
+        "flag": "assets/flags/fr.png",
+        "served_url": "",
+        "active": "N"
```
lmmrssa: Active is supposed to be used to filter the language option which is not done.

src/app/home/navigation.component.ts
```diff
@@ -34,6 +54,16 @@ export class NavigationComponent implements OnInit {
     { link: '', name: 'Feedback' },
   ];
 
+  ngOnInit() {
+    Object.assign(this, this.userService.get());
+    this.languages = (<any>languages);
+  }
+
+  switchLanguage(lang) {
```
lmmrssa: later once we have AOT it should change the url of application
And selected language needs to be set based on the url of application

src/app/home/navigation.component.ts
paulbert: It should go into shared because it will be built into the js files Webpack outputs.  The assets folder is for files that will be included as is in the app.

Issue #310, opened 2018-01-12
Resource list change the rating section based on image below


lmmrssa:
![resource](https://user-images.githubusercontent.com/2909490/34861938-ade24e38-f78f-11e7-847f-0567b3e487ed.jpg)

Just make the rating section to reflect it on list view itself

lmmrssa:
@yaoxing1990 This issue was created just to implement rating section into current list view.
If you are working on the grid view then that will be different issue and view not to replace list view.

PR #311, opened 2018-01-14
[#310] Added grid view (connects #310)

yaoxing1990:


yaoxing1990:
@lmmrssa @paulbert It seems that the inner organization in mat-grid-tile is not properly implemented. Could you give some guidance? 
capture

lmmrssa:
could you point to the design you tried to implement. you are forcing image ratio to change instead it should be image with that ratio not squeezed one. Also you do not have any margin or padding between two sections title and rating.

yaoxing1990:
@lmmrssa I want to implement [Grid List View](https://material.angular.io/components/grid-list/overview), to replace the original [mat table](https://material.angular.io/components/table/overview). And I did not figure out how to give appropriate positions.

lmmrssa:
@yaoxing1990 issue you tried to work on was not meant to implement grid view. It was just supposed to update the rating section of list view.
Grid view will require different issue to work on. Also it will not replace the mat-table view instead it should be two switchable views (Grid/List)

paulbert:
Closing since this was not the intended goal of the issue, and we won't be implementing a grid view in the near future.

Issue #309, opened 2018-01-11
Resource Activity


nirojdyola:
Create Resource Activity 
Data structure:
ratings
{
    user: 
    item: resource/course (even feedback or reply could be rated) (parentId)
    type:
    rate:
    comment:
    time:
}

PR #312, opened 2018-01-15
[#309] Resource Activity stored in DB (connects #309)

nirojdyola:


src/app/resources/view-resources/resources-view.component.ts
```diff
@@ -86,4 +92,17 @@ export class ResourcesViewComponent implements OnInit, OnDestroy {
...
   }
 
+  resource_activity(resource_id, activity) {
+    const datas = {
+      'resource': resource_id,
+      'user': this.userService.get().name,
+      'activity': activity,
+      'time': Date()
+    };
+    this.couchService.post('resource_activities', datas)
```
nirojdyola: same user same resource saved multiple times in DB.

Issue #313, opened 2018-01-15
Create a menubar for notification


lmmrssa:
Notification should show a dropdown
It should popup similar to one shown below profile icon on given image but for notification (bell)
Just keep some sample data resembling notification for now

![dashboard v4dashboard v2 4 copy 5](https://user-images.githubusercontent.com/2909490/34929545-9327a84a-f9ec-11e7-9786-1393293aae38.jpg)

PR #316, opened 2018-01-15
[#313] Notification MenuBar (connects #313)

nirojdyola:


src/app/home/navigation.component.ts
```diff
@@ -11,7 +11,17 @@ import { Router } from '@angular/router';
       <li><a href="#" class="km-logout" (click)="logoutClick()">LOGOUT</a></li>
       <li><a routerLink="/manager"><i class="material-icons">settings</i></a></li>
       <li *ngIf="roles.indexOf('_admin') === -1"><a routerLink="/users/profile/{{name}}"><mat-icon>person</mat-icon></a></li>
+      <li><a href="#"><mat-icon [matMenuTriggerFor]="notification">notifications</mat-icon></a></li>
```
paulbert: When clicked acts as a link to `/#`.  Please remove the link as this should simply open the menu.

src/app/home/navigation.component.ts
```diff
@@ -11,7 +11,17 @@ import { Router } from '@angular/router';
       <li><a href="#" class="km-logout" (click)="logoutClick()">LOGOUT</a></li>
       <li><a routerLink="/manager"><i class="material-icons">settings</i></a></li>
       <li *ngIf="roles.indexOf('_admin') === -1"><a routerLink="/users/profile/{{name}}"><mat-icon>person</mat-icon></a></li>
+      <li><a href="#"><mat-icon [matMenuTriggerFor]="notification">notifications</mat-icon></a></li>
     </ul>
+    <mat-menu #notification="matMenu">
+      <mat-list>
+        <mat-list-item> Notification 1 </mat-list-item>
```
paulbert: Each element should simply be a `mat-menu-item` (see https://material.angular.io/components/menu).  I think you can use a `div` rather than `button` to get the same effect without the odd spacing that the `mat-list` is creating.

Issue #314, opened 2018-01-15
Create sidebar menu


lmmrssa:
Create a menu on sidebar which should have icon for now and look something similar to shown in image below
![dashboard-new2](https://user-images.githubusercontent.com/2909490/34929700-873dcd60-f9ed-11e7-8c30-a76c8eb22866.jpg)

PR #317, opened 2018-01-15
[#314] New navigation (top & sidebar) (connects #314)

paulbert:
I changed the meetups icon because I couldn't find the one with the additional speech bubble in https://material.io/icons/ and it is very similar to the members icon, anyway.  I'm also not sure what the third from bottom (half white/half a box with three lines in it) is intended to link to, so it links back to the home page.

I think we should consider as a next step adding text at least for a minimum screen size if not for all sizes.  I am familiar with the app and I had trouble figuring out what icons should link to which page, so I think it would be difficult for someone new to figure it out.


lmmrssa:
@paulbert PR seems good. I would like to have the icons showing always and the text you mentioned will toggle based on the `menu icon`

lmmrssa:
https://stackblitz.com/angular/vmvlxjylbrj
It changes the size of sidebar

src/app/home/home.component.html
```diff
@@ -0,0 +1,91 @@
+<div class="background"></div>
+<mat-toolbar color="primary" class="header-nav">
+  <button mat-icon-button (click)="toggleNav()"><mat-icon>menu</mat-icon></button>
+  <span>Planet Learning</span>
+  <span class="toolbar-fill"></span>
+  <button mat-icon-button routerLink="/manager"><mat-icon>settings</mat-icon></button>
+  <button mat-icon-button [matMenuTriggerFor]="notificationMenu"><mat-icon>notifications</mat-icon></button>
+  <button mat-icon-button [matMenuTriggerFor]="languageMenu">
+    <img *ngIf="current_flag" src="assets/flags/{{current_flag}}.png" i18n-alt alt="{{current_lang}}" i18n-title
```
paulbert: Just noticed these `i18n-alt` and `i18n-title` tags won't work since the alt and title tags are set by interpolation.  We will need to update that.

PR #321, opened 2018-01-16
[#319] resources rating on view page (connects #319)

razu9861:


src/app/resources/view-resources/resources-view.component.ts
```diff
@@ -17,7 +17,62 @@ import { Subject } from 'rxjs/Subject';
...
+    .rating {
+      text-align: center;
+    }
+    $gray: #aaa;
+
+    body{
+      background: black;
+      color: white;
+    }
+    .container{
```
paulbert: Please stick to Material components and basic styling until we have more specific designs from UX

src/app/resources/view-resources/resources-view.component.ts
```diff
@@ -17,7 +17,62 @@ import { Subject } from 'rxjs/Subject';
...
+      display: grid;
+      max-width: 230px;
+      margin-left: auto;
+    }
+    .rating {
+      text-align: center;
+    }
+    $gray: #aaa;
+
+    body{
```
paulbert: This style affects the entire app.  Do not add styles to app wide tags/component like `body` in a specific component style.

src/app/resources/view-resources/resources-view.component.ts
```diff
@@ -17,7 +17,62 @@ import { Subject } from 'rxjs/Subject';
...
+          margin:16px;
+          padding: 40px;
+          .margin-vertical-8{
+            margin: 8px 0;
+          }
+        }
+      }
+    }
+
+    .star-rating{
```
paulbert: We already have styles in `styles.scss` for styling & the color of the star ratings.

src/app/resources/view-resources/resources-view.component.ts
```diff
@@ -89,6 +102,66 @@ export class ResourcesViewComponent implements OnInit, OnDestroy {
...
+        });
+        let rate_sum = 0;
+        let has_rated = 0;
+        let total_rating = 0;
+        let male_rating = 0;
+        let female_rating = 0;
+        this.ratings.map(rate => {
+          has_rated = (rate.user === this.userService.get().name) ? rate.rate : has_rated;
+          total_rating++;
+          (rate.gender === 'M') ? male_rating++ : female_rating++ ;
```
paulbert: A few problems here:

- The gender will be in `rate.user.gender`, so this will always increment `female_rating`
- The value being stored is `'male'` or '`female`', so it will never equal `'M'`
- We should explicitly check that it equals '`female'` before incrementing `female_rating` in case the user has not filled out their profile.
- Using the ternary operator in this case is not ideal.  It's not intended to be a shorthand if/else with code running in each case, but rather to decide between values to assign to a variable based on a conditional.  This is a good case to use a `switch`, as it makes it clear to someone reading the code that we are running a different block of code based on the value of `rate.user.gender`

src/app/resources/view-resources/resources-view.component.ts
```diff
@@ -89,6 +102,66 @@ export class ResourcesViewComponent implements OnInit, OnDestroy {
       this.pdfSrc = this.sanitizer.bypassSecurityTrustResourceUrl(this.resourceSrc);
     }
     this.couchSrc = this.urlPrefix + resource._id + '/' + filename;
+    // resource rating
+    this.couchService.get('ratings/_all_docs?include_docs=true')
```
paulbert: You should do a mango query to find the specific ratings for this resource rather then fetch all resources & `map` and `filter` over each value.

src/app/resources/view-resources/resources-view.component.ts
```diff
@@ -89,6 +102,66 @@ export class ResourcesViewComponent implements OnInit, OnDestroy {
...
+      rate: [ '' ],
+      comment: [ '' ]
+    };
+    this.dialogsFormService
+      .confirm(title, type, fields, validation, '')
+      .debug('Dialog confirm')
+      .subscribe((res) => {
+        if (res !== undefined) {
+          const datas = {
+            'user': this.userService.get().name,
```
paulbert: We should insert the full user object here.  Adding a `getProfile` method to the `userService` which fetches that information would work.

src/app/resources/view-resources/resources-view.component.ts
```diff
@@ -89,6 +102,66 @@ export class ResourcesViewComponent implements OnInit, OnDestroy {
...
+            'gender': 'M', // gender need to fetch from profile (need to work on admin part)
+            'item': resource_id,
+            'type': 'resource',
+            'rate': res.rate,
+            'comment': res.comment,
+            'time': Date.now()
+          };
+          this.couchService.post('ratings', datas)
```
paulbert: We should avoid nesting observables.  Please refactor.

src/app/resources/view-resources/resources-view.component.ts
```diff
@@ -89,6 +102,66 @@ export class ResourcesViewComponent implements OnInit, OnDestroy {
...
+        });
+        this.resource.rating = rate_sum;
+        this.resource.has_rated = has_rated;
+        this.resource.female_rating = (female_rating / total_rating) * 100;
+        this.resource.male_rating = (male_rating / total_rating) * 100;
+        this.resource.total_rating = total_rating;
+      }, (error) => console.log(error));
+  }
+
+  rate(resource_id) {
```
paulbert: Give this a clearer function name like `openRatingDialog` which describes the action better.

src/app/resources/view-resources/resources-view.component.ts
nirojdyola: Removed.

src/app/resources/view-resources/resources-view.component.ts
nirojdyola: Renamed to openRatingDialog

src/app/resources/view-resources/resources-view.component.ts
nirojdyola: Used mango query

src/app/resources/view-resources/resources-view.component.ts
```diff
@@ -89,6 +102,66 @@ export class ResourcesViewComponent implements OnInit, OnDestroy {
...
     this.couchSrc = this.urlPrefix + resource._id + '/' + filename;
+    // resource rating
+    this.couchService.get('ratings/_all_docs?include_docs=true')
+      .subscribe((data) => {
+        this.ratings = data.rows.map(ratings => {
+          return ratings.doc;
+        }).filter(rt  => {
+          return rt['type'] === 'resource' && rt['item'] === resource._id;
+        });
+        let rate_sum = 0;
```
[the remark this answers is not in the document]
nirojdyola: All the variable has been changed.

src/app/resources/view-resources/resources-view.component.ts
nirojdyola: That means getProfile need to fetch his/her data from _users document , so Should I query it when user is set in auth-guard.service.ts.
After 
this.userService.set(res.userCtx);

paulbert: Yes, do a get request for the full document from the _users database.  The auth-guard service runs on every route change, so if we run it from there we should only request profile data if the user name has changed.

paulbert:
Note there is a change on `nations.component` because there were some unused parameters on the `dialogs-form.service` which I removed.

src/app/resources/view-resources/resources-view.component.ts
```diff
@@ -18,8 +21,9 @@ import { UserService } from '../../shared/user.service';
       height: 80vh;
       border: none;
     }
-  ` ]
+  ` ],
```
lmmrssa: extra comma at end

src/app/resources/view-resources/resources-view.component.ts
```diff
@@ -91,7 +97,79 @@ export class ResourcesViewComponent implements OnInit, OnDestroy {
...
+          userRating, totalCount });
+      }, error => console.log(error));
+  }
+
+  openRatingDialog(resource_id) {
+    const title = 'Rating';
+    const type = 'rating';
+    const fields =
+      [
+        { 'label': 'Rate', 'type': 'radio', 'name': 'rate', 'placeholder': 'Your Rating', 'required': false },
```
lmmrssa: can we add some extra parameter so it says its rating and form only change for rating not all radio

src/app/shared/dialogs/dialogs-form.component.html
```diff
@@ -1,23 +1,38 @@
...
-      <mat-form-field *ngIf="field.type == 'selectbox'" class="full-width">
+      <mat-form-field *ngIf="field.type === 'selectbox'" class="full-width">
         <select mat-select *ngFor="let options of field.options">
           <option value="{{option.value}}">{{option.text}}</option>
         </select>
         <mat-error><planet-form-error-messages [control]="modalForm.controls[field.name]"></planet-form-error-messages></mat-error>
       </mat-form-field>
+
+      <!-- Star Rating -->
+      <mat-form-field class="rating" *ngIf="field.type === 'radio'">
```
lmmrssa: we need to check if it is only rating not for all radio

src/app/shared/dialogs/dialogs-form.component.html
```diff
@@ -1,23 +1,38 @@
...
+      </mat-form-field>
     </div>
   </mat-dialog-content>
   <mat-dialog-actions>
     <button color="warn" type="button" mat-raised-button (click)="dialogRef.close()" i18n>Cancel</button>&nbsp;
     <button type="submit" color="primary" mat-raised-button i18n>Submit</button>
   </mat-dialog-actions>
 </form>
+
+
```
lmmrssa: extra line

Issue #320, opened 2018-01-16
Notification


nirojdyola:
Notificaton on 
- user register
- new feedback

PR #322, opened 2018-01-16
[#320] Notification (connects #320)

nirojdyola:
Also need to check for admin register notification.


nirojdyola:
Since navigation page is no more, will create new PR

razu9861:
Since navigation page is no more, will create new PR  #344

Issue #324, opened 2018-01-17
db-init docker not working


mappuji:
Need to change how we run `couchdb-setup.sh`

from `./couchdb-setup.sh` to `sh ./couchdb-setup.sh` since we are change the shebang https://github.com/ole-vi/planet/commit/1f4b1dbb8225b82f9c99be098695c8c8a916f634

<img width="1240" alt="screen shot 2018-01-17 at 2 14 32 pm" src="https://user-images.githubusercontent.com/11813607/35030206-c72ae6d4-fb90-11e7-9607-79a87502ccfe.png">
<img width="1098" alt="screen shot 2018-01-17 at 2 13 35 pm" src="https://user-images.githubusercontent.com/11813607/35030209-c8e8275c-fb90-11e7-9e3f-1d529479063d.png">

PR #326, opened 2018-01-17
~Use sh in our couchdb setup to make db-init docker in alpine can use it~ Add bash instead

mappuji:
Links to #324 

@Rupesh87 was changing the shebang of `couchdb-setup.sh` and it breaks our `db-init` docker since it doesn't has bash (it is an alpine linux container).

Is this shebang not sufficient in our vagrant? @dogi @paulbert

### DIFF
--- a/docker/db-init/Dockerfile
+++ b/docker/db-init/Dockerfile
@@ -1,7 +1,7 @@
 FROM node:6.11-alpine
 
 RUN apk update ;\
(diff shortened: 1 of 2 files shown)


mappuji:
Hi @paulbert I think I can make the script `sh` compliant at make it work also inside vagrant, let me fix.

PR #329, opened 2018-01-17
redirect user to edit profile after registration

amanuely:
@lmmrssa

### DIFF
--- a/src/app/login/login-form.component.ts
+++ b/src/app/login/login-form.component.ts
@@ -25,7 +25,7 @@ export class LoginFormComponent {
     if (this.createMode) {
       this.createUser(this.model);
     } else {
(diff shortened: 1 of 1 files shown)


src/app/login/login.component.ts
```diff
@@ -95,4 +95,10 @@ export class LoginComponent implements OnInit {
         this.reRoute();
       }, (error) => this.message = 'Username and/or password do not match');
   }
+  loginAndEditProfile({ name, password }: {name: string, password: string}) {
```
paulbert: Instead of creating a new function which is almost identical to the `login` function, let's refactor the `login` function to take an optional parameter which when set will redirect to the user's update page.

Issue #332, opened 2018-01-18
Course list page similar to resource list page


lmmrssa:
- [ ] Add Title and back
- [ ] Header bar
- [ ] Add selection

PR #334, opened 2018-01-19
[#332] Course made similar to resources (connects #332)

BasnetSamriddhi:


src/app/courses/courses.component.html
```diff
@@ -1,32 +1,62 @@
...
+          </mat-checkbox>
+        </mat-cell>
+      </ng-container>
+
+      <ng-container matColumnDef="title">
+        <mat-header-cell *matHeaderCellDef mat-sort-header i18n> Title </mat-header-cell>
+        <mat-cell class="km-course-title" *matCellDef="let element"> {{element.courseTitle}} </mat-cell>
+      </ng-container>
+      <ng-container matColumnDef="action">
+        <mat-header-cell *matHeaderCellDef i18n> Action </mat-header-cell>
```
nirojdyola: Instead of this action , we can create menu as same as done in resource section.

src/app/courses/courses.component.ts
```diff
@@ -2,22 +2,32 @@ import { Component, OnInit, AfterViewInit, ViewChild } from '@angular/core';
...
 @Component({
-  templateUrl: './courses.component.html'
+  templateUrl: './courses.component.html',
+  styles: [ `
+    .space-container {
+      margin: 64px 30px;
+    }
+` ]
 })
-export class CoursesComponent implements OnInit, AfterViewInit {
```
nirojdyola: Remove extra line.

src/app/courses/courses.component.ts
```diff
@@ -2,22 +2,32 @@ import { Component, OnInit, AfterViewInit, ViewChild } from '@angular/core';
...
   deleteDialog: any;
 
   constructor(
     private couchService: CouchService,
-    private dialog: MatDialog
+    private dialog: MatDialog,
+    private location: Location
   ) { }
 
   ngOnInit() {
```
nirojdyola: If the function is not using, then we can remove it . So, Could you please check  it once ?

lmmrssa: @nirojdyola  there is actually a code inside if you look after the whole file

src/app/courses/courses.component.html
```diff
@@ -1,32 +1,97 @@
 <div class="mat-elevation-z8 courses">
```
paulbert: Please remove the `mat-elevation-z8` class

src/app/courses/courses.component.html
```diff
@@ -1,32 +1,97 @@
...
   <p>{{message}}</p>
```
paulbert: We should move this, or possibly remove it, since the margin of the `p` tag is adding extra space at the top.

src/app/courses/courses.component.html
```diff
@@ -1,32 +1,95 @@
...
-          <button mat-raised-button color="warn" (click)="deleteClick(element)" i18n> Delete </button>
-      </mat-cell>
-    </ng-container>
-    <mat-header-row *matHeaderRowDef="displayedColumns"></mat-header-row>
-    <mat-row *matRowDef="let row; columns: displayedColumns;"></mat-row>
-  </mat-table>
-  <mat-paginator #paginator [pageSize]="10" [pageSizeOptions]="[5, 10, 20]" i18n></mat-paginator>
 </div>
+
+
```
lmmrssa: this is a extra line

PR #340, opened 2018-01-22
[#339] Add course page (connects #339)

razu9861:


src/app/courses/add-courses/courses-add.component.ts
```diff
@@ -17,7 +17,18 @@ import { MatFormField, MatFormFieldControl } from '@angular/material';
 
 
 @Component({
-  templateUrl: 'courses-add.component.html'
+  templateUrl: 'courses-add.component.html',
+  styles: [ `
+    /* Consider using space-container app wide for route views */
```
lmmrssa: why do we have this style here?

src/app/courses/add-courses/courses-add.component.html
```diff
@@ -1,102 +1,109 @@
...
+      <div>
+        <button type="submit" mat-raised-button color="primary" i18n>Submit</button>
+        <button type="button" mat-raised-button color="warn" (click)="cancel()" i18n>Cancel</button>
+      </div>
+    </form>
+  </div>
+</div>
 <!--FOR DEBUGGING (remove once view is final)-->
-<div>
+<!-- <div>
```
lmmrssa: km-course-valid is used by testcase so it needs to be there. If its removed test also needs to be updated.

src/app/courses/add-courses/courses-add.component.html
```diff
@@ -1,102 +1,109 @@
...
-      <mat-select i18n-placeholder placeholder="Course Leader" formControlName="courseLeader">
-        <mat-option *ngFor="let memb of members" value={{memb?.id}}> {{memb?.doc.firstName}} {{memb?.doc.middleName}} {{memb?.doc.lastName}}</mat-option>
-      </mat-select>
-    </mat-form-field>
-    <mat-form-field>
-      <input matInput i18n-placeholder placeholder="Method" formControlName="method">
-    </mat-form-field>
-  </div>
+<div class="space-container">
+  <mat-toolbar class="primary-color font-size-1" i18n>Add new Course</mat-toolbar>
```
lmmrssa: it has changed the title can we use the same title or atleast all words to have same case sensitivity.

src/app/home/home.component.ts
```diff
@@ -65,7 +65,7 @@ export class HomeComponent implements OnInit, AfterViewInit {
   }
 
   backgroundRoute() {
-    const routesWithBackground = [ 'resources' ];
+    const routesWithBackground = [ 'resources', 'courses' ];
```
paulbert: In case #334 takes longer than this, please add `courses/add`

Issue #348, opened 2018-01-24
Create a message section


lmmrssa:
Create a message section that will display message regarding any request error
Record being inserted, updated or any other events

PR #356, opened 2018-01-26
[#348] Message section added(connects #348)

BasnetSamriddhi:


src/app/app.module.ts
```diff
@@ -6,23 +6,23 @@ import { AppComponent } from './app.component';
 import { PageNotFoundComponent } from './page-not-found/page-not-found.component';
 
 import { AppRoutingModule } from './app-router.module';
-
 import { CouchService } from './shared/couchdb.service';
 import { AuthService } from './shared/auth-guard.service';
 import { UserService } from './shared/user.service';
 import { ValidatorService } from './validators/validator.service';
-
+import { MatSnackBarModule } from '@angular/material/snack-bar';
```
nirojdyola: The material module should be define in material file . Could you please remove this and define it in Materail file.

src/app/app.module.ts
```diff
@@ -6,12 +6,10 @@ import { AppComponent } from './app.component';
 import { PageNotFoundComponent } from './page-not-found/page-not-found.component';
 
 import { AppRoutingModule } from './app-router.module';
-
```
paulbert: Please leave these spaces in.  We do not need these changes in the history.

Issue #354, opened 2018-01-25
Start replacing icons


lmmrssa:
We have few icons inside https://drive.google.com/drive/u/3/folders/0Bw7aA5bLT2P9WVBLV0tUalAzcms
We shall start replacing our application with these icons

PR #357, opened 2018-01-29
[#354] icon change (connect #354)

MacNew:
only one suitable icon match for the project


mappuji:
Can we just `git reset @~` then only add the file you change @MacNew ?

razu9861:
This PR is closed, reference New Pr of this issue #361

Issue #368, opened 2018-02-01
Design User Profle  page


razu9861:
![v2 43users-profile](https://user-images.githubusercontent.com/24647983/35661364-8214f3b2-0709-11e8-86d1-4f720e1f6d02.png)

PR #370, opened 2018-02-02
[#368] modified User Profile section(connects #368)

BasnetSamriddhi:


src/app/users/users-profile/users-profile.component.html
```diff
@@ -1,53 +1,73 @@
...
+        </ng-container>
+        <ng-template #noAttachment>
+          <img class="user-profile" src="../assets/image.png">
+        </ng-template>
+      </mat-grid-tile>
+      <mat-grid-tile>
+        <mat-list>
+          <mat-list-item>
+            <h3 matLine i18n>Full Name</h3>
+            <h2 mat-line i18n><b>{{userDetail.firstName}} {{userDetail.middleName}} {{userDetail.lastName}}</b></h2>
```
nirojdyola: I think i18n is not needed in dynamic section , could you please remove it and check it another section too ?

src/app/home/home.component.ts
```diff
@@ -65,7 +65,7 @@ export class HomeComponent implements OnInit, AfterViewInit {
   }
 
   backgroundRoute() {
-    const routesWithBackground = [ 'resources', 'courses' ];
+    const routesWithBackground = [ 'resources', 'courses', 'users' ];
```
lmmrssa: this is not looking good for member list

src/app/users/users-profile/users-profile.component.ts
```diff
@@ -125,4 +138,12 @@ export class UsersProfileComponent implements OnInit {
     };
   }
 
+  goBack() {
+    this.location.back();
+  }
+
+  searchFilter(filterValue: string) {
```
lmmrssa: why do we have this filter?

src/styles.scss
```diff
@@ -202,7 +202,8 @@ body {
   }
 
   .user-profile {
-    border-radius: 5px;
-    width: 125px;
+    border-radius: 50%;
+    width: 100%;
+    height: 100%;
```
lmmrssa: it is not required, we will have to either force the ratio of image or crop if we want to keep image circular

src/app/users/users-profile/users-profile.component.ts
```diff
@@ -6,12 +6,24 @@ import { UserService } from '../../shared/user.service';
...
+  templateUrl: './users-profile.component.html',
+  styles: [ `
+    .space-container {
+      margin: 64px 30px;
+    }
+    .view-container {
+      background-color:#fff;
+      padding: 8%;
+    }
+    mat-list-item h3{
```
paulbert: Please create a `.primary-color` class in `styles.scss` which uses the `$primary` variable so if we change our color scheme this will change with it.

src/app/users/users-profile/users-profile.component.ts
```diff
@@ -125,4 +134,8 @@ export class UsersProfileComponent implements OnInit {
     };
   }
 
+  goBack() {
+    this.location.back();
```
lmmrssa: change route to go back to dashboard or user list based on user role

src/app/users/users-update/users-update.component.html
```diff
@@ -1,62 +1,70 @@
...
+            </mat-option>
+          </mat-select>
+          <mat-error><planet-form-error-messages [control]="editForm.controls.level"></planet-form-error-messages></mat-error>
+        </mat-form-field>
+        <label>File Upload:</label>
+        <input type="file" (change)="bindFile($event)" />
+      </div>
+      <div>
+        <button type="submit" mat-raised-button color="primary" i18n>Update</button>
+        <button type="button" mat-raised-button color="warn" (click)="cancel()" i18n>Cancel</button>
```
lmmrssa: cancel and goback on topbar should do the same thing so we can use same function

src/app/users/users-update/users-update.component.ts
```diff
@@ -95,6 +104,10 @@ export class UsersUpdateComponent implements OnInit {
     this.location.back();
   }
 
+  goBack() {
```
lmmrssa: goback() and cancel() above it should be one function and it should take to profile view

src/app/users/users-update/users-update.component.ts
```diff
@@ -9,22 +9,34 @@ import { Observable } from 'rxjs/Observable';
...
-
+  roles: string[] = [];
   constructor(
     private fb: FormBuilder,
     private couchService: CouchService,
     private route: ActivatedRoute,
-    private router: Router
+    private router: Router,
+    private userService: UserService
+
```
lmmrssa: extra line

src/styles.scss
```diff
@@ -212,7 +212,11 @@ body {
   }
 
   .user-profile {
-    border-radius: 5px;
-    width: 125px;
+    border-radius: 50%;
+    width: 100%;
+  }
+
+  .text-color{
```
lmmrssa: it should change to something meaningful

lmmrssa:
We can move Full name to the headerbar or else remove login from detail to full name as header bar already has space for login/fullname
can we shift the positioning of field to look like
Login DOB
Email Phone
Language Level

For gender we can show icon of gender next to name on header bar or if no image added we can use the default image based on gender

paulbert:
I don't like the change to use the gender icon rather than text.  With both male/female icons like on the resources list it's easy to understand, but with one on its own it's not readily apparent what the icon means.

Also the last commit I made changes these tags because the Material CSS for `mat-list-item` was overriding most of the CSS for `h3` & `h2` tags so using those was confusing.

Issue #365, opened 2018-01-31
Add url (current route) to document when adding feedback


paulbert:
Also, if possible, include screenshot of page as an attachment.

paulbert:
@razu9861 @nirojdyola When you are able to work on the screenshot part, we may be able to use the screenshot method on the browser utility from protractor, which we are already including in the `package.json` for e2e testing.

paulbert:
See [here](https://github.com/angular/protractor/blob/master/docs/faq.md#how-can-i-get-screenshots-of-failures) for protractor documentation on taking screenshots (this writes to the file system, we'd just want to add it to CouchDB as an attachment)

PR #371, opened 2018-02-02
[#365] Current Url to feedback (connects #365)

razu9861:
brainstroming on taking screenshot, 
i have found some packages like html2canvas, if all agrees we can add package for screenshot

### DIFF
--- a/src/app/feedback/feedback.directive.ts
+++ b/src/app/feedback/feedback.directive.ts
@@ -5,6 +5,9 @@ import { FormsModule } from '@angular/forms';
 import { CouchService } from '../shared/couchdb.service';
 import { Validators } from '@angular/forms';
 import { DialogsFormService } from '../shared/dialogs/dialogs-form.service';
+import { Router,  ActivatedRoute, NavigationEnd } from '@angular/router';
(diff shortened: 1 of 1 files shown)


src/app/feedback/feedback.directive.ts
```diff
@@ -33,7 +34,7 @@ export class FeedbackDirective {
     private userService: UserService,
     private couchService: CouchService,
     private dialogsFormService: DialogsFormService
-  ) {}
+  ) { }
```
nirojdyola: Nothing is different before and after in code. Please check it once.

src/app/feedback/feedback.directive.ts
```diff
@@ -6,6 +6,7 @@ import { CouchService } from '../shared/couchdb.service';
 import { Validators } from '@angular/forms';
 import { DialogsFormService } from '../shared/dialogs/dialogs-form.service';
 
+
```
nirojdyola: Extra line . Please check it once.

paulbert:
We should use `router.url` from Angular rather than the document.

Issue #358, opened 2018-01-30
Fix login_activity on browser refresh when session is still active


paulbert:
We either need to treat this as a new session for the `login_activity` database or fetch the info from the CouchDB to repopulate `sessionRev` and `sessionId` in `user.service.ts`

PR #374, opened 2018-02-05
[358] Login Activity Maintain on Refresh Page (connects #358)

nirojdyola:


src/app/shared/user.service.ts
```diff
@@ -71,14 +71,15 @@ export class UserService {
   endSessionLog() {
     if (this.sessionId === undefined) {
       return this.couchService.post(this.logsDb + '/_find', findDocuments(
-              { 'user': this.get().name, 'logout_time': 0 },
+              { 'user': this.get().name },
               [ '_id', '_rev', 'login_time' ],
-              1
+              [ { '_id': 'desc' } ]
```
lmmrssa: _id will not be generated such that it will be larger as time pass.
So we could use last login_time instead

nirojdyola: Yes, but couch db is not allowing me to do it. 
screen shot 2018-02-13 at 9 17 31 am

Issue #384, opened 2018-02-07
Convert location.back to specific route


paulbert:
Using location.back for the back buttons can cause an issue where it takes you back to the login page.  Let's convert all uses of this to navigate up the "app tree" (for example `/courses/add` to `/courses` then `/courses` to `/`) rather than use location.back.

PR #387, opened 2018-02-08
[#384] Convert location.back to specific route

lucy811:


src/app/courses/courses.component.ts
```diff
@@ -6,6 +6,7 @@ import { PlanetMessageService } from '../shared/planet-message.service';
 import { MatCheckboxModule } from '@angular/material/checkbox';
 import { SelectionModel } from '@angular/cdk/collections';
 import { Location } from '@angular/common';
```
nirojdyola: If location is not using anymore, then we can remove import section as well. And also can remove from constructor.

Issue #352, opened 2018-01-24
Dashboard View


lmmrssa:
Start filling dashboard view
- [x] Add the center block with different blocks

nirojdyola:
![Dashboard V4Dashboard V2.4.jpg](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/58aeb9eb257d931a000df4b2/125516c66e82c728ace21e0d46cc95892b808ca89dedff75f079f187214428723b00e31d22cbb76af3006b1300051bff054a5556f9f0dd2ff3e53a76db5a5de4d36c5f74c46b45faed861ebb13a632c91edbe2e3d7158abd7f37ed998d4f341f00198ff75830a69d4dc051f359ac.jpg)

Should have 4 blocksas shown in image.
Let's focus on design now.

PR #393, opened 2018-02-13
[#352] New dashboard view

paulbert:


paulbert:
This is ready for some reviews.  Since we don't have detail pages for courses or meetups, clicking those goes to the full list, and since pulling a full user list would require more changes for non-admins I'm leaving myTeams with example items.

paulbert:
@razu9861 What version of Firefox is that?  It's displaying correctly on the current version.

@nirojdyola Yes, there are a few warnings about the css.  It's because the grid areas are set in a different file.  The grid displays correctly and the app builds with no errors, so I think we can still move forward with this as long as it is otherwise OK.

lmmrssa:
@razu9861 if this error comes up on other versions as well then will try to create new issue for it

Issue #378, opened 2018-02-07
Community Filter


nirojdyola:
![screen shot 2018-02-07 at 11 37 08 am](https://user-images.githubusercontent.com/7495421/35900785-9fbae338-0bcb-11e8-8ac8-814bf3f63fe9.png)

If you check into snapshot, MadagascarBeLL and accepted is filtered but ht.np is also displaying in the list.

PR #396, opened 2018-02-13
[#378] CommunityFilter (connects #378)

nirojdyola:


paulbert:
Simplified it somewhat, but this still needs a reduction in cognitive complexity.

Issue #388, opened 2018-02-09
Access restricted to admins


nirojdyola:
![screen shot 2018-02-09 at 12 39 56 pm](https://user-images.githubusercontent.com/7495421/36015553-d280f0c2-0d66-11e8-99e0-6d167812ba4b.png)
On the first setup that means admin is also restricted on viewing users. Please fix the issue.

lmmrssa:
It looks like its not getting the admin user copy being created on the first instance. But later on it seems to work. So need to look further on getuser

PR #399, opened 2018-02-13
[#388] admin restriction fixed(connects #388)

BasnetSamriddhi:


lmmrssa:
closing it as it seems like it has already been fixed

Issue #402, opened 2018-02-14
Different users are created with same name (No case sensitive)


MacNew:
Not checking case sensitive  between registered and new user. Different users are created with same name.

PR #407, opened 2018-02-15
[#402] Userwithsame name problem solve (connects #402)

MacNew:


nirojdyola:
It doesn't give any error message when I gave some special characters as a login name .
screen shot 2018-02-15 at 2 20 50 pm

src/app/login/login-form.component.ts
```diff
@@ -51,30 +51,30 @@ export class LoginFormComponent {
...
-      { 'name': name, 'password': password, 'roles': [], 'type': 'user', 'isUserAdmin': false })
+      this.couchService.put('_users/org.couchdb.user:' + name.toLowerCase(),
+      { 'name': name.toLowerCase(), 'password': password, 'roles': [], 'type': 'user', 'isUserAdmin': false })
         .subscribe((data) => {
-          this.message = 'User created: ' + data.id.replace('org.couchdb.user:', '');
+          this.planetMessageService.showMessage('User created: ' + data.id.replace('org.couchdb.user:', ''));
           this.welcomeNotification(data.id);
           this.login(this.model, true);
-        }, (error) => this.message = '');
+        }, (error) => this.planetMessageService.showMessage(error.error.reason));
```
paulbert: Let's replace this message with something user friendly.  `There was an error in creating a new user` would be better

src/app/login/login-form.component.ts
```diff
@@ -41,8 +42,7 @@ export class LoginFormComponent {
     };
     this.couchService.post('notifications', data)
       .subscribe((res) => {
-        console.log(res);
-      }, (error) => this.message = 'Error');
+      }, (error) => this.planetMessageService.showMessage('Error'));
```
paulbert: Let's not pass an error message from this to the message service.  If there is an error, it will replace the user created message and confuse the person into thinking something went wrong creating the user.

src/app/login/login-form.component.ts
```diff
@@ -51,30 +51,30 @@ export class LoginFormComponent {
...
         return this.userService.newSessionLog();
       })).subscribe((res) => {
         if (isCreate) {
-          this.router.navigate( [ 'users/update/' + name ]);
+          this.router.navigate( [ 'users/update/' + name.toLowerCase() ]);
         } else {
           this.reRoute();
         }
-      }, (error) => this.message = 'Username and/or password do not match');
+      }, (error) => this.planetMessageService.showMessage(error.error.reason));
```
paulbert: Same as above, but let's pass in `Username and/or password do not match` since that will be the most common error response.

src/app/app.module.ts
```diff
@@ -11,18 +11,19 @@ import { CouchService } from './shared/couchdb.service';
...
   imports: [
-    BrowserModule, AppRoutingModule, HttpClientModule, BrowserAnimationsModule
+    BrowserModule, AppRoutingModule, HttpClientModule, BrowserAnimationsModule, MaterialModule
   ],
   declarations: [
     AppComponent, PageNotFoundComponent
   ],
   providers: [
-    CouchService, AuthService, UserService, ValidatorService
+    CouchService, AuthService, UserService, ValidatorService, PlanetMessageService
```
paulbert: We should remove this service from other module provider arrays since it only needs to appear in one.

Issue #392, opened 2018-02-13
Full screen of pdf on resource view not working as expected


lmmrssa:
Full screen of pdf is expected to take the whole screen of the application.
But right now it is just switching its mode using the same space.
![screen shot 2018-02-13 at 09 56 09](https://user-images.githubusercontent.com/2909490/36133211-565cb972-10a4-11e8-92f3-d9454000ce8b.png)
![screen shot 2018-02-13 at 09 55 57](https://user-images.githubusercontent.com/2909490/36133213-585515ee-10a4-11e8-80fe-a89657e53f69.png)

PR #410, opened 2018-02-15
[#392] Added fullscreen view for PDF(Connects #392)

BasnetSamriddhi:
The full screen view for PDF is working only in Firefox, other browsers are not supporting it.


paulbert:
On the tablets this solution automatically downloads the PDF rather than showing it in the browser frame.  The previous solution also does not work on the tablets, though.  I'm leaning towards merging this and keeping the issue open with these notes.

Issue #394, opened 2018-02-13
login redirect to configuration


lmmrssa:
When user is logged in and application is requested to 'login' url, it seems like it is redirecting to admin configuration. 
So please redirect to dashboard.
![screen shot 2018-02-13 at 10 41 10](https://user-images.githubusercontent.com/2909490/36134077-96df83de-10aa-11e8-8ec8-7e527944e138.png)

PR #411, opened 2018-02-15
[#394] Logged In  Redirect on Configuration (connects #394)

nirojdyola:


src/app/login/login.component.ts
```diff
@@ -25,7 +25,7 @@ export class LoginComponent implements OnInit {
     if (!environment.test) {
       this.checkAdminExistence().subscribe((noAdmin) => {
         // false means there is admin
-        if (noAdmin) {
+        if (noAdmin && !this.checkAuthSession()) {
```
lmmrssa: this should check for is_logged in or has session. and redirect to dashboard. No need to check for admin exist if already logged in

src/app/login/login.component.ts
```diff
@@ -43,4 +43,11 @@ export class LoginComponent implements OnInit {
     );
   }
 
+  checkAuthSession() {
+    return this.couchService.get('_session')
```
paulbert: This does not work as you intended: it will simply return the Observable which evaluates to `true`.  Please remember that callbacks in asynchronous functions (promises, observables, etc) don't push their return value up to a surrounding function.

PR #414, opened 2018-02-16
[#413] password value checked (connects #413)

MacNew:


src/app/login/login-form.component.html
```diff
@@ -1,12 +1,12 @@
-<form (ngSubmit)="onSubmit()" #loginForm="ngForm" class="login-form">
+<form (ngSubmit)="onSubmit()" class="login-form" [formGroup]="userForm">
   <mat-form-field class="example-full-width">
-    <input matInput [(ngModel)]="model.name" i18n-placeholder placeholder="Username" name="name" />
+    <input matInput [(ngModel)]="model.name" i18n-placeholder placeholder="Username" name="name" formControlName = "name"/>
```
lmmrssa: ngmodel is not required if you are using reactive form

lmmrssa: Also there is not validator added

src/app/login/login-form.component.ts
```diff
@@ -3,18 +3,27 @@ import { CouchService } from '../shared/couchdb.service';
...
     private route: ActivatedRoute,
-    private userService: UserService
-  ) { }
+    private userService: UserService,
+    private formBuilder: FormBuilder
+  ) {
+       this.userForm = this.formBuilder.group({
+       name: [ '', Validators.required ],
+       password: [ '', Validators.required ],
+       repeatPassword: [ '', Validators.required ]
```
lmmrssa: is this working for both login and registration case?

src/app/login/login-form.component.ts
```diff
@@ -23,9 +32,17 @@ export class LoginFormComponent {
 
   onSubmit() {
     if (this.createMode) {
-      this.createUser(this.model);
+      if (this.userForm.valid) {
+        this.createUser(this.model);
+      } else {
+        this.message = 'please, fill up required credential';
```
lmmrssa: this should come from reactive form validator not as separate message.
Also there is not event triggering invalidate message show

src/app/login/login-form.component.ts
```diff
@@ -49,21 +80,17 @@ export class LoginFormComponent {
     this.router.navigate([ this.returnUrl ]);
   }
 
-  createUser({ name, password, repeatPassword }: {name: string, password: string, repeatPassword: string}) {
-    if (password === repeatPassword) {
+  createUser(name: string, password: string, repeatPassword: string) {
```
paulbert: This does not need to change.  Previous version was using object destructuring (which you can read up about in the [Typescript docs](https://www.typescriptlang.org/docs/handbook/variable-declarations.html)) allowing us to simply pass the object to the function rather than each property as a separate argument.

src/app/login/login-form.component.ts
```diff
@@ -49,21 +80,17 @@ export class LoginFormComponent {
...
-          this.login(this.model, true);
+          this.login(this.userForm.value.name, this.userForm.value.password, true);
         }, (error) => this.message = '');
-    } else {
-      this.message = 'Passwords do not match';
-    }
   }
 
-  login({ name, password }: {name: string, password: string}, isCreate: boolean) {
+  login(name: string, password: string, isCreate: boolean) {
```
paulbert: Same as above.  No need to change this from the previous destructured version.

src/app/login/login-form.component.ts
```diff
@@ -3,30 +3,61 @@ import { CouchService } from '../shared/couchdb.service';
...
+         control.markAsTouched({ onlySelf: true });
+         });
+      }
     } else {
-      this.login(this.model, false);
-    }
+        if (this.userForm.value.password !== '' && this.userForm.value.name !== '') {
+          this.login(this.userForm.value.name, this.userForm.value.password, false);
+        } else {
+          Object.keys(this.userForm.controls).forEach(field => {
```
paulbert: Repetitive block (same as lines 46-49)

src/app/login/login-form.component.ts
```diff
@@ -3,30 +3,61 @@ import { CouchService } from '../shared/couchdb.service';
...
+      } else {
+         Object.keys(this.userForm.controls).forEach(field => {
+         const control = this.userForm.get(field);
+         control.markAsTouched({ onlySelf: true });
+         });
+      }
     } else {
-      this.login(this.model, false);
-    }
+        if (this.userForm.value.password !== '' && this.userForm.value.name !== '') {
```
paulbert: If we are using reactive forms, you should use that validation.  Please review the [documentation](https://angular.io/guide/reactive-forms) for a solution.

src/app/login/login-form.component.html
```diff
@@ -1,12 +1,21 @@
-<form (ngSubmit)="onSubmit()" #loginForm="ngForm" class="login-form">
+<form (ngSubmit)="onSubmit()" class="login-form" [formGroup]="userForm">
   <mat-form-field class="example-full-width">
-    <input matInput [(ngModel)]="model.name" i18n-placeholder placeholder="Username" name="name" />
+    <input matInput  i18n-placeholder placeholder="Username" name="name" formControlName = "name"/>
+    <mat-error>
+      <planet-form-error-messages class="km-coursetitle-errormessage" [control]="userForm.controls.name">
+      </planet-form-error-messages>
+  </mat-error>
```
paulbert: Please fix the indentation on this file.

lmmrssa:
@paulbert there was no validation done. so people were able to login or sign up without password.
So we are changing login form to reactive to add validation

src/app/login/login-form.component.html
```diff
@@ -1,12 +1,21 @@
...
+    <input matInput  i18n-placeholder placeholder="Password" name="password" type="password" formControlName = "password"/>
+    <mat-error> 
+    <planet-form-error-messages class="km-coursetitle-errormessage" [control]="userForm.controls.password">
+    </planet-form-error-messages>
+    </mat-error>
   </mat-form-field>
   <mat-form-field class="example-full-width" *ngIf="createMode">
-    <input matInput [(ngModel)]="model.repeatPassword" i18n-placeholder placeholder="Repeat Password" name="repeatPassword" type="password" />
+    <input matInput  i18n-placeholder placeholder="Repeat Password" name="repeatPassword" type="password" formControlName = "repeatPassword" />
+    <mat-error><planet-form-error-messages class="km-coursetitle-errormessage" [control]="userForm.controls.repeatPassword"></planet-form-error-messages></mat-error>
```
lmmrssa: could we not use km-coursetitle class

Issue #409, opened 2018-02-15
Star Rating design 


nirojdyola:
Works fine in Chrome, but not in mozilla. Some circle is displaying over star.
And also when I click upto 5 star, it is not displaying at all.
![screen shot 2018-02-15 at 2 51 16 pm](https://user-images.githubusercontent.com/7495421/36248527-bbee18d0-122f-11e8-8144-5c6636598e91.png)

paulbert:
The problem is the `display: inline` that was necessary to make sure the yellow stars overlapped is causing Firefox to render the incorrect icon, with incorrect sizing and spacing.

This isn't newbie friendly.  The cause of this odd behavior is unclear and the solution will not be simple.

PR #417, opened 2018-02-19
[#409] Star Rating Design (connects #409)

nirojdyola:
Snapshot of Chrome and Mozilla respectively.
screen shot 2018-02-19 at 11 09 59 am

### DIFF
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -118,7 +118,7 @@ body {
       top: 0;
       left: 0;
       overflow: hidden;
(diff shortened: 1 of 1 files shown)


paulbert:
Don't need the browser prefix for flexbox (see https://caniuse.com/#search=flex).  I also noticed the alignment of the gold stars is a little off on the tablets (unrelated to this issue), but it's otherwise working on the tablets as well.

PR #428, opened 2018-02-22
packagejson file update

Rupesh87:
@all Due to angular new version there was an error so we updated package.json file. Please have a look....


nirojdyola:
The problem was : 
screen shot 2018-02-22 at 11 29 09 am

Issue #429, opened 2018-02-22
unexpected div bar on user section 


razu9861:
![screen shot 2018-02-22 at 3 52 10 pm](https://user-images.githubusercontent.com/24647983/36532286-32595b46-17b8-11e8-8dd2-081109970058.png)

PR #430, opened 2018-02-22
[#429] unexpected white div bar removed (connects #429)

razu9861:


src/app/users/users.component.html
```diff
@@ -95,4 +95,4 @@ <h1 i18n>All Users</h1>
   </mat-table>
   <mat-paginator #paginator [pageSize]="10" [pageSizeOptions]="[5, 10, 20]"></mat-paginator>
 </div>
-<div class="view-container">{{message}}</div>
+<div class="view-container" *ngIf="!isUserAdmin">{{message}}</div>
```
lmmrssa: I see the whitespace out there but it would have been better to change the style of message section or instead of view-container making it something else.

Issue #433, opened 2018-02-23
Connection between nation and community


Rupesh87:

lmmrssa:
First phase task will be setting up planet to link with nation/center
- [ ] Setup configuration of planet will sync up to selected nation/center
- [ ] Admin during setup will be created as user on nation/center
- [ ] User will have different type and link to its related planet
- [ ] Approve/Reject of community will also activate or deactivate related user

PR #440, opened 2018-02-26
[#433] connects nation community (connects #433)

Rupesh87:


src/app/home/home.component.html
```diff
@@ -29,10 +29,10 @@
     <p class="menu-item-text" *ngIf="notification.doc.status==='unread' && notification.doc.link ===''">
       {{notification.doc.message}} {{notification.doc.time | date: 'MMM d, yyyy'}}
     </p>
-    <p *ngIf="notification.status!=='unread' && notification.doc.link">
```
paulbert: These changes from the `Duplicate notification` commit should be removed as they are for a different PR.  Please make sure to begin branches from the `master` branch.

Rupesh87:
@paulbert  @dogi when connecting between nation and community we use the same data in configurations and communityregistrationrequests databases. Do we want to replicate data in between or make them different entities with different ID?

paulbert:
@Rupesh87 I think different entities.  That way we avoid conflicts if the uuid assigned locally to the configuration is the same as one that already exists on the nation in communityregistrationrequests.  An unlikely scenario, but it's easy for us to avoid.

Also, please make sure to fix linting errors before pushing:

image

src/app/community/community.component.ts
```diff
@@ -84,11 +66,38 @@ export class CommunityComponent implements OnInit, AfterViewInit {
...
+          .subscribe((data) => {
+            this.updateRev(data, this.communities.data);
+            this.editDialog.close();
+          }, (error) => this.editDialog.componentInstance.message = 'There was a problem accepting this community');
+      }
+      if (change === 'accept') {
+        community.registrationRequest = 'accepted';
+        this.couchService.get('_users/_all_docs?include_docs=true')
+          .subscribe((data) => {
+            data.rows.map(data => {
```
paulbert: This block includes 4 levels of nested CouchDB requests/subscribes, 3 of which are nested in a `map`.

1. Please **do not** loop over even 1 couch request.  Use [_bulk_docs](http://docs.couchdb.org/en/2.1.1/api/database/bulk-api.html#db-bulk-docs) instead.
2. Do not nest subscribes.  Create different functions and/or use the [switchMap](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-switchMap) rxjs operator to clean this up.

mappuji:
hints:

src/app/shared/couchdb.service.ts
```diff
@@ -14,8 +14,8 @@ export class CouchService {
     return Object.assign({}, this.defaultOpts, opts) || this.defaultOpts;
   }
 
-  private couchDBReq(type: string, db: string, opts: any, data?: any) {
-    const url = this.baseUrl + db;
+  private couchDBReq(type: string, db: string, opts: any, data?: any, parentLink?: any) {
```
paulbert: We should include parentLink in the opts variable and adjust the rest of the code to match.  I'd rather we have one optional parameter for the methods available on the CouchService.

src/app/configuration/configuration.component.ts
```diff
@@ -99,23 +104,32 @@ export class ConfigurationComponent implements OnInit {
...
-          }, (error) => this.message = '');
-          const config = Object.assign({}, this.configurationFormGroup.value, this.contactFormGroup.value);
-          this.couchService.post('configurations', config).subscribe(() => {
+      const configuration = Object.assign({ registrationRequest: 'pending' },
+        this.configurationFormGroup.value, this.contactFormGroup.value);
+      const userDetail = { 'name': this.loginForm.value.username,
+            'password': this.loginForm.value.password,
+            roles: [],
+            'type': 'user',
+            'isUserAdmin': true,
```
paulbert: isUserAdmin should only be set on the local user, not the parent user

src/app/configuration/configuration.component.ts
```diff
@@ -99,23 +104,32 @@ export class ConfigurationComponent implements OnInit {
...
+            'phoneNumber': this.contactFormGroup.value.phoneNumber
+          };
+      forkJoin([
+        this.couchService.put('_node/nonode@nohost/_config/admins/' + this.loginForm.value.username, this.loginForm.value.password),
+        this.couchService.put('_users/org.couchdb.user:' + this.loginForm.value.username, userDetail),
+        this.couchService.post('configurations', configuration),
+        this.couchService.post('communityregistrationrequests', configuration, {}, configuration.parent_domain)
+      ]).debug('Sending request to parent planet').subscribe((data) => {
+        userDetail['request_id'] =  data[3].id;
+        this.couchService.put('/_users/org.couchdb.user:' + this.loginForm.value.username,
```
paulbert: Please use switchMap rather than nesting observables.

src/app/shared/couchdb.service.ts
Rupesh87: @paulbert httpRequest is using opts so we used another optional (parentLink). If we are going to include parentLink inside opts then we have to remove parentLink while requesting. so which approach we should follow???

paulbert:
A few things noted today:
- User needs to be logged out of both child and parent when logout clicked.
- User needs to renew session with parent.  Otherwise viewing parent resources will not work after session resets (10 minutes).
- We should handle the situation when the parent is not accessible (someone has access to the nation because they are on the same network, but that network does not have access to the internet & center).  We will also need to allow the user to log into the center when it becomes available.
- Password changes need to be synced with all accounts (need to handle the above case for this as well)

paulbert:
Just raising a question: do we want to add a document to `login_activities` of the parent (nation) in addition to the child?

We'll have to make some adjustments if so, notably the ids will not match so if we sync the two dbs we will have duplicate sessions recorded.  Also we should have a field that stores the nation/community name in the session so we know what the user logged into.

Rupesh87:
Yes, we think we need to maintain login activities in parent  environment as well. But it is not necessary to do that. So, what’s your plan? Should we remove the login activities to the parent planet ?

Issue #437, opened 2018-02-26
isDownloadable Field for Resource Add


nirojdyola:
isDownloadable field needs to be add in the resource add section. You can give any meaningful name as well. This field needs to be checkbox. If the box has been checked, then the file can be downloadable in  the resource view section.

PR #442, opened 2018-02-26
[#437] added downloadable field in resource

BasnetSamriddhi:


src/app/resources/resources-add.component.ts
```diff
@@ -46,7 +46,7 @@ export class ResourcesAddComponent implements OnInit {
   file: any;
   resourceForm: FormGroup;
   readonly dbName = 'resources'; // make database name a constant
-
+  ischecked = false;
```
paulbert: isChecked is part of formGroup, so this here is unnecessary and should be removed

src/app/resources/resources.component.ts
```diff
@@ -193,4 +193,10 @@ export class ResourcesComponent implements OnInit, AfterViewInit {
     this.router.navigate([ '/' ]);
   }
 
+  download(dataFile) {
+    const a = document.createElement('a');
```
paulbert: We should not directly modify the DOM unless absolutely necessary.  Angular should handle all DOM manipulation otherwise we may see unintended side effects.

src/app/resources/view-resources/resources-view.component.ts
```diff
@@ -182,4 +181,10 @@ export class ResourcesViewComponent implements OnInit, OnDestroy {
     this.fullView = this.fullView === 'on' ? 'off' : 'on';
   }
 
+  download(dataFile) {
+    const a = document.createElement('a');
```
paulbert: Same here.  You can simply create the link on the template to download the file.

src/app/resources/view-resources/resources-view.component.html
```diff
@@ -9,6 +9,10 @@
     <button mat-button (click)="toggleFullView()" i18n>
       {fullView, select, on {Show} off {Hide}} details
     </button>
+    <span class="toolbar-fill"></span>
+    <button mat-button (click)="download(resource)" [disabled]="!resource.ischecked">
```
paulbert: For buttons with an icon use `mat-icon-button` rather than `mat-button`

src/app/resources/view-resources/resources-view.component.html
```diff
@@ -9,6 +9,10 @@
     <button mat-button (click)="toggleFullView()" i18n>
       {fullView, select, on {Show} off {Hide}} details
     </button>
+    <span class="toolbar-fill"></span>
+    <a href={{resourceSrc}} mat-icon-button download="{{resource.filename}}" [disabled]="!resource.isDownloadable">
```
paulbert: Setting the download attribute changes the file name to be saved locally, so by setting it to what the file is already named you aren't doing anything.

src/app/resources/resources.component.html
```diff
@@ -60,6 +60,10 @@ <h3 class="header"><a *ngIf="!nationName" [routerLink]="['/resources/view', elem
             <mat-icon>delete_forever</mat-icon>
             <span i18n>Delete</span>
           </button>
+          <a href={{resourceSrc}} mat-menu-item (click)="downloadFile(element)" download [disabled]="!element.isDownloadable">
```
paulbert: Again, you don't need a click handler.  Just set the href to the url for the file to download.

This does, however, highlight the point that the filename property isn't set for our mock data.  I'll make a new issue for that.

Issue #438, opened 2018-02-26
Notification View List


nirojdyola:
Please create the Notification view list. We don;t have view for the given list. You can take example from github notification view list or any site.

PR #443, opened 2018-02-26
[#438] notification view list (connects #438)

MacNew:


src/app/home/home.component.html
```diff
@@ -21,21 +21,12 @@
...
-    </p>
-    <p *ngIf="notification.status!=='unread' && notification.doc.link">
-      <a href="{{notification.doc.link}}">{{notification.doc.message}} {{notification.doc.time | date: 'MMM d, yyyy'}}</a>
-    </p>
-    <p *ngIf="notification.status!=='unread' && notification.doc.link===''">
-      {{notification.doc.message}} {{notification.doc.time | date: 'MMM d, yyyy'}}
-    </p>
-  </span>
+  <mat-list *ngFor="let notification of notifications">
+    <mat-list-item (click)="readNotification(notification.doc)" *ngIf="notification.doc.status=='unread'" class="menu-item-text">
```
nirojdyola: This has been changed in https://github.com/ole-vi/planet/pull/432. So could you please look into the new changes.

paulbert: Yes, please rebase and use version from master

lmmrssa:
after rebasing can we change this to open new list view when someone clicks on more notification. Instead of taking to new view on each item click. Then we can limit notification on dropdown to some numbers (probably 10)

src/app/notification-view/notification-view.component.ts
```diff
@@ -0,0 +1,18 @@
...
+})
+export class NotificationViewComponent implements OnInit {
+  message: string;
+  constructor(
+    private route: ActivatedRoute
+  ) {}
+
+  ngOnInit() {
+    this.route.queryParams.subscribe(params => {
+        this.message = params['message'];
```
paulbert: Indentation is off, and after switching to use a route parameter you'll need to grab the notification info from the database.

src/app/notification-view-list/notification-view-list.component.ts
```diff
@@ -0,0 +1,60 @@
...
+    const user_id = 'org.couchdb.user:' + this.userService.get().name;
+    this.couchService.get('notifications/_all_docs?include_docs=true')
+      .subscribe((data) => {
+        let cnt = 0;
+        data.rows.sort((a, b) => 0 - (new Date(a.doc.time) > new Date(b.doc.time) ? 1 : -1));
+        this.notifications = data.rows.map(notifications => {
+          if (notifications.doc.status === 'unread') {
+            cnt ++;
+          }
+          return notifications;
```
paulbert: Variable should be singular and you should return `notifications.doc` and adjust the rest accordingly

src/app/notification-view-list/notification-view-list.component.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  getNotification() {
+    const user_id = 'org.couchdb.user:' + this.userService.get().name;
+    this.couchService.get('notifications/_all_docs?include_docs=true')
+      .subscribe((data) => {
+        let cnt = 0;
+        data.rows.sort((a, b) => 0 - (new Date(a.doc.time) > new Date(b.doc.time) ? 1 : -1));
+        this.notifications = data.rows.map(notifications => {
+          if (notifications.doc.status === 'unread') {
+            cnt ++;
```
paulbert: Extra space

src/app/notification-view-list/notification-view-list.component.ts
```diff
@@ -0,0 +1,60 @@
...
+        }).filter(nt  => {
+          return nt.doc['user'] === user_id;
+        });
+        this.notifications['count_unread'] =  cnt;
+      }, (error) => console.log(error));
+  }
+
+  readNotification(notification) {
+    const update_notificaton =  { ...notification, 'status': 'read' };
+    this.couchService.put('notifications/' + notification._id, update_notificaton).subscribe((data) => {
```
paulbert: Only need to update if it was previously unread

src/app/notification-view-list/notification-view-list.component.ts
```diff
@@ -0,0 +1,62 @@
+import { Component, OnInit } from '@angular/core';
+import { UserService } from '../shared/user.service';
+import { CouchService } from '../shared/couchdb.service';
+
+@Component({
+  template: `
+    Your Notifications
```
razu9861: Is this notification message or just a text, if it is text to display then please add translations

src/app/home/home-router.module.ts
```diff
@@ -18,7 +18,8 @@ const routes: Routes = [
       { path: 'community', component: CommunityComponent },
       { path: 'community/:nation', component: CommunityComponent },
       { path: 'resources', loadChildren: '../resources/resources.module#ResourcesModule' },
-      { path: 'meetups', loadChildren: '../meetups/meetups.module#MeetupsModule' }
+      { path: 'meetups', loadChildren: '../meetups/meetups.module#MeetupsModule' },
+      { path: 'notificationviewlist', component: NotificationViewListComponent }
```
nirojdyola: Instead of such a big path name notificatonviewlist, we can use notification/view .

src/app/home/home.component.html
```diff
@@ -21,12 +21,16 @@
 </mat-menu>
 <!--Notification dropdown menu-->
 <mat-menu #notificationMenu="matMenu" [overlapTrigger]="false">
-  <span class="menu-item-notification" mat-menu-item *ngFor="let notification of notifications" (click)="readNotification(notification.doc)">
+  <a routerLink="{{notification.doc.link}}"
```
paulbert: Are there instances where we would not have this link go to the 'notification/view' route?  If the menu links always lead to the 'notification/view', we should just have that here.

src/app/notification-view-list/notification-view-list.component.ts
```diff
@@ -0,0 +1,62 @@
...
+        <p [ngClass]="{'menu-item-text':notification.status==='unread'}">
+          <a routerLink="{{notification.link}}">
+            {{notification.message}} {{notification.time | date: 'MMM d, yyyy'}}
+          </a>
+        </p>
+      </mat-list-item>
+    </mat-list>
+  `,
+  styles: [ `
+    .menu-item-text {
```
paulbert: Please remove this style from this file and `home.scss`.

Add this style to `style.scss` below the `.accent-color` definition and use it instead of `menu-item-text` in the two locations:
```
.primary-text-color {
  color: $primary;
}
```

src/app/notification-view-list/notification-view-list.component.ts
```diff
@@ -0,0 +1,62 @@
...
+        data.rows.sort((a, b) => 0 - (new Date(a.doc.time) > new Date(b.doc.time) ? 1 : -1));
+        this.notifications = data.rows.map(notifications => {
+          if (notifications.doc.status === 'unread') {
+            cnt ++;
+          }
+          return notifications.doc;
+        }).filter(nt => {
+          return nt['user'] === user_id;
+        });
+        this.notifications['count_unread'] =  cnt;
```
paulbert: `this.notifications` is an array.  Now you are converting it into an object.  You should instead just add a property on the class for `countUnread` (for consistency, please use camelCase variable names in the JavaScript) and keep `this.notifications` as an array.

src/app/notification-view-list/notification-view-list.component.ts
paulbert: There is still an extra space here.  It should be `cnt++`

src/app/notification-view-list/notification-view-list.component.ts
```diff
@@ -0,0 +1,62 @@
...
+      .subscribe((data) => {
+        let cnt = 0;
+        data.rows.sort((a, b) => 0 - (new Date(a.doc.time) > new Date(b.doc.time) ? 1 : -1));
+        this.notifications = data.rows.map(notifications => {
+          if (notifications.doc.status === 'unread') {
+            cnt ++;
+          }
+          return notifications.doc;
+        }).filter(nt => {
+          return nt['user'] === user_id;
```
paulbert: Please only use bracket notation when necessary (i.e. when accessing an object property dynamically with a string variable).  In this case, use dot notation.

Also please use `userId` rather than `user_id`.

src/app/notification-view-list/notification-view-list.component.ts
```diff
@@ -0,0 +1,62 @@
...
+  constructor(
+    private couchService: CouchService,
+    private userService: UserService
+    ) { }
+
+  ngOnInit() {
+    this.getNotification();
+  }
+
+  getNotification() {
```
paulbert: Rename this to `getNotifications` so it's clearer we're getting many rather than one

src/app/notification-view-list/notification-view-list.component.ts
```diff
@@ -0,0 +1,62 @@
...
+    const user_id = 'org.couchdb.user:' + this.userService.get().name;
+    this.couchService.get('notifications/_all_docs?include_docs=true')
```
paulbert: This is a good case to use a [Mango Query with the _find endpoint](http://docs.couchdb.org/en/2.1.1/api/database/find.html).  That way we can pull filtered and sorted data rather than filter and sort from the client.

src/app/notification-view-list/notification-view-list.component.ts
```diff
@@ -0,0 +1,62 @@
...
+import { CouchService } from '../shared/couchdb.service';
+
+@Component({
+  template: `
+    <p i18n>Your Notifications</p>
+    <mat-list role="list" *ngFor="let notification of notifications">
+      <mat-list-item (click)="readNotification(notification)">
+      <mat-divider></mat-divider>
+        <p [ngClass]="{'menu-item-text':notification.status==='unread'}">
+          <a routerLink="{{notification.link}}">
```
paulbert: See my other question.  Right now the only notification link goes to this list, so this goes nowhere.  If we will have notifications lead to other pages, this is fine, but if not then this does not need to be a link.

src/app/home/home-router.module.ts
```diff
@@ -1,11 +1,11 @@
 import { NgModule } from '@angular/core';
 import { Routes, RouterModule } from '@angular/router';
-
 import { DashboardComponent } from '../dashboard/dashboard.component';
 import { HomeComponent } from './home.component';
 import { CommunityComponent } from '../community/community.component';
 import { NationComponent } from '../nation/nation.component';
 import { ManagerDashboardComponent } from '../manager-dashboard/manager-dashboard.component';
+import { NotificationViewListComponent } from '../notification-view-list/notification-view-list.component';
```
paulbert: This can simply be `NotificationComponent` if it's the only one.

src/app/home/home-router.module.ts
```diff
@@ -18,7 +18,8 @@ const routes: Routes = [
       { path: 'community', component: CommunityComponent },
       { path: 'community/:nation', component: CommunityComponent },
       { path: 'resources', loadChildren: '../resources/resources.module#ResourcesModule' },
-      { path: 'meetups', loadChildren: '../meetups/meetups.module#MeetupsModule' }
+      { path: 'meetups', loadChildren: '../meetups/meetups.module#MeetupsModule' },
+      { path: 'notification/view', component: NotificationViewListComponent }
```
paulbert: Similar to the above, the path can simply be 'notifications' if it's the only notification component.

Issue #454, opened 2018-02-28
admin user is kinda not getting any data ...


dogi:
![screen shot 2018-02-28 at 1 15 26 am](https://user-images.githubusercontent.com/13684059/36772896-15a22c82-1c25-11e8-8eb9-ab04a33f3806.png)

paulbert:
Do you mean that the information inputted during configuration does not get added to the admin profile?  If so, then yes I agree this is an issue that should be fixed.  First, Middle, Last Name, Email, & Phone Number should all be populated from the configuration.

![login_configuration](https://user-images.githubusercontent.com/9203229/36798266-2a41ef80-1c78-11e8-94e9-61c608082dab.png)

paulbert:
Reopening.  The commit is only working when a role has been deleted using the "X" in the pill.

paulbert:
Hmm...I think I reopened the wrong issue based on that last comment.  Closing.

PR #465, opened 2018-03-01
[#454] Fix admin user is kinda not getting any data 

lucy811:


src/app/configuration/configuration.component.ts
```diff
@@ -101,7 +101,9 @@ export class ConfigurationComponent implements OnInit {
         .subscribe((data) => {
           this.couchService.put('_users/org.couchdb.user:' + this.loginForm.value.username,
           { 'name': this.loginForm.value.username, 'password': this.loginForm.value.password, roles: [], 'type': 'user',
-          'isUserAdmin': true }).subscribe((data1) => {
+            'isUserAdmin': true, 'firstName': this.contactFormGroup.value.firstName, 'middleName': this.contactFormGroup.value.middleName,
+            'lastName': this.contactFormGroup.value.lastName, 'email': this.contactFormGroup.value.email,
+            'phoneNumber': this.contactFormGroup.value.phoneNumber }).subscribe((data1) => {
             this.message = 'Admin created: ' + data1.id.replace('org.couchdb.user:', '');
```
nirojdyola: We already implemented PlanetMessageService to show the message. So , we can use it here too.

Issue #457, opened 2018-02-28
Hide manager menu from learner


dogi:
![screen shot 2018-02-28 at 1 53 51 am](https://user-images.githubusercontent.com/13684059/36774047-774f08d8-1c2a-11e8-9f11-e62e6face7ff.png)

wonder generally how we do with that menu,
but what I know so far that a learner should see it

PR #466, opened 2018-03-02
[#457] hide usersetting from users connects (#457)

razu9861:
while setting up configuration its creating user without role, and it is adding isUserAdmin= true, therefore while checking roles its giving empty array[], thats why i have checked isUserAdmin for this issue. while creating user the value for isUserAdmin is false. so the manager setting button will be visible only if isUserAdmin is true

### DIFF
--- a/src/app/manager-dashboard/manager-dashboard.component.ts
+++ b/src/app/manager-dashboard/manager-dashboard.component.ts
@@ -1,12 +1,26 @@
-import { Component } from '@angular/core';
+import { Component, OnInit } from '@angular/core';
+import { UserService } from '../shared/user.service';
 
 @Component({
   template: `
-    <div><a routerLink="/community" i18n mat-raised-button>Communities</a> <a routerLink="/nation" i18n mat-raised-button>Nations</a></div>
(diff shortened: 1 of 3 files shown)


nirojdyola:
We also need to block the non-admin from the manager address as well. I mean to say , ,when I put 
http://localhost:3000/manager in address bar it is opening, so we also need to check that one.

screen shot 2018-03-05 at 10 19 03 am

src/app/manager-dashboard/manager-dashboard.component.ts
```diff
@@ -1,12 +1,26 @@
...
+export class ManagerDashboardComponent implements OnInit {
+  isUserAdmin = false;
+  displayDashboard = true;
+  message = '';
 
-  constructor() { }
+  constructor( private userService: UserService) { }
+
+  ngOnInit() {
+    Object.assign(this, this.userService.get());
```
paulbert: I think I encouraged something not so good by using this earlier.  Now that we have many properties from `this.userService.get()`, we shouldn't apply all of them to our classes.  Instead, you should simply assign `this.isUserAdmin  = this.userService.get().isUserAdmin`

Issue #458, opened 2018-02-28
Refactor template CSS & HTML


paulbert:
The below layout has become standard for many pages on our app, but since it was initially only implemented on `resources.component` there is some code duplication in CSS and HTML across the app.  We should refactor the implementation to reduce this duplication.

![resources](https://user-images.githubusercontent.com/9203229/36809173-31bce2ac-1c95-11e8-9c35-89d7645a9058.png)

paulbert:
Fixed in #467

PR #467, opened 2018-03-02
[#458] removed duplicate codes(connets #458)

BasnetSamriddhi:


src/styles.scss
```diff
@@ -239,5 +239,25 @@ body {
     height: 50px;
     padding: 3px;
   }
+  
+  .space-container {
+    margin: 64px 30px;
+  }
+
+  .mat-column-select {
```
paulbert: These column sizes are actually specific to those components, so should stay as part of the component styles.

src/styles.scss
```diff
@@ -239,5 +239,25 @@ body {
...
+
+  .mat-column-select {
+    max-width: 44px;
+  }
 
+  .mat-column-action, .mat-column-rating{
+    max-width: 225px;
+  }
+  
+  .tableAnchor a:hover {
```
paulbert: We should use a more meaningful class name and kebab-case rather than camelCase for consistency.  Maybe `.primary-link-hover` unless I think of something better tonight

src/styles.scss
```diff
@@ -239,5 +239,25 @@ body {
...
+  
+  .space-container {
+    margin: 64px 30px;
+  }
+
+  .mat-column-select {
+    max-width: 44px;
+  }
 
+  .mat-column-action, .mat-column-rating{
```
paulbert: Please add a space before the bracket.

src/styles.scss
```diff
@@ -239,5 +239,25 @@ body {
...
+    max-width: 225px;
+  }
+  
+  .tableAnchor a:hover {
+    color:$primary;
```
paulbert: Please add a space after the colon.

PR #469, opened 2018-03-02
[#468] pop up error message (connects #468)

MacNew:


src/app/login/login-form.component.ts
```diff
@@ -69,7 +69,11 @@ export class LoginFormComponent {
           this.planetMessageService.showMessage('User created: ' + data.id.replace('org.couchdb.user:', ''));
           this.welcomeNotification(data.id);
           this.login(this.userForm.value, true);
-        }, (error) => this.message = '');
+        }, error => {
+          if (error.error.error === 'conflict') {
+            this.planetMessageService.showMessage('User name already exists, please, register with a different user name');
```
lmmrssa: We can either have general message not checking error condition or we can modify this message to look clean

Issue #456, opened 2018-02-28
can't log out - what can I do now?


dogi:
version 0.0.14

I went nuclear and restarted the whole services,
cause that was the only thing left ...

but next time I would <3 to know what to do
so I can help the effort to find and fix this bug

PR #472, opened 2018-03-05
[#456] Users logout fixed (connects #456)

razu9861:


src/app/shared/auth-guard.service.ts
```diff
@@ -17,10 +17,7 @@ export class AuthService {
       .get('_session', { withCredentials: true })
       .pipe(switchMap((res: any) => {
         if (res.userCtx.name) {
-          // If user already matches one on the user service, do not make additional call to CouchDB
-          if (res.userCtx.name === this.userService.get().name) {
-            return of(true);
-          }
+          console.log("Res", res.userCtx.name)
```
paulbert: This code stops unnecessary GET requests to the CouchDB when we already have the user information.  Please modify this to not run setProfile if it is already set

mappuji:
hints:

Issue #400, opened 2018-02-13
production docker with multiple ip addresses


dogi:
We want to access the app from different ip address and it should work every where.

```
root@ole:~/planet/docker# git diff
diff --git a/docker/rpi-planet.yml b/docker/rpi-planet.yml
index 9d3c5b6..55b608a 100644
--- a/docker/rpi-planet.yml
+++ b/docker/rpi-planet.yml
@@ -13,10 +13,10 @@ services:
   planet:
     image: treehouses/planet:rpi-latest
     ports:
-      - "80:80"
+      - "3000:80"
     environment:
       - HOST_PROTOCOL=http
-      - DB_HOST=127.0.0.1
+      - DB_HOST=192.168.2.1
       - DB_PORT=2200
     depends_on:
       - couchdb
root@ole:~/planet/docker# ifconfig |grep 192
        inet 192.168.0.107  netmask 255.255.255.0  broadcast 192.168.0.255
        inet 192.168.2.1  netmask 255.255.255.0  broadcast 192.168.2.255
```

Possible solution proposed by @mappuji is redirect request from other IP to main IP

mappuji:
you can use

```
window.location.host
```

PR #475, opened 2018-03-05
[#400] dynamic origin host

mappuji:
Solve #400

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,10 @@
 # BeLL apps Angular reboot prototype **Planet**
 
-Project to create a prototype for a reboot of the BeLL apps using Angular4 & CouchDB.
+Project to create a prototype for a reboot of the BeLL apps using Angular5 & CouchDB2.
 
 ## To work on this
 
-The only prerequisite is Vagrant. If you don't know about it, please do some research and try it. After cloning the repository, run `vagrant up` in the console (In this phase you should already know about vagrant). Once it's done installing the virtual machine it'll automatically start compiling the app.  After about 10 seconds, you can open the app at `localhost:3000`
-
-Please create an issue if the vagrant base image not available in the future for some reason.
+The only prerequisite is Vagrant. If you don't know about it, please do some research and try it. After cloning the repository, run `vagrant up` in the console. Once it's done installing the virtual machine it'll automatically start compiling the app.  After about 10 seconds, you can open the app at `localhost:3000`.
 
 ## Project guidelines
 
--- a/README.md
+++ b/README.md
@@ -4,13 +4,15 @@ Project to create a prototype for a reboot of the BeLL apps using Angular4 & Cou
 
 ## To work on this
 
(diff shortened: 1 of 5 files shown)


docker/planet/docker-entrypoint.sh
```diff
@@ -10,7 +10,13 @@ else
   PROTOCOL="http"
 fi
 
-sed -i -e "s#planet-db-host#$PROTOCOL://$DB_HOST#g" /usr/share/nginx/html/main*
+if [ "$ON_RPI" = "true" ]
```
lmmrssa: does docker-entrypoint runs on every request? As it seems to be using window.location which is supposed to be from browser

README.md
```diff
@@ -1,16 +1,16 @@
 # BeLL apps Angular reboot prototype **Planet**
 
-Project to create a prototype for a reboot of the BeLL apps using Angular4 & CouchDB.
+Project to create a prototype for a reboot of the BeLL apps using Angular5 & CouchDB2.
 
 ## To work on this
 
-The only prequisite is Vagrant.  After cloning the repository, run `vagrant up` in the console.  Once it's done installing the virtual machine it'll automatically start compiling the app.  After about 10 seconds, you can open the app at `localhost:3000`
+The only prerequisite is Vagrant. If you don't know about it, please do some research and try it. After cloning the repository, run `vagrant up` in the console. Once it's done installing the virtual machine it'll automatically start compiling the app.  After about 10 seconds, you can open the app at `localhost:3000`.
```
lmmrssa: NOTE: CPU or BIOS should have virtulization (VT-x) enabled, Parallelization is not enough.
Also for windows Pro Hypervisor should not be enabled

src/environments/environment.prod.ts
```diff
@@ -2,5 +2,5 @@ export const environment = {
   production: true,
   test: false,
   // Change this to Docker address
-  couchAddress: window.location.hostname + ':planet-db-port/'
+  couchAddress: window.location.protocol + '//' + window.location.hostname + ':planet-db-port/'
```
dogi: @mappuji think we need to keep changing this line ... you remove this later

dogi:
@mappuji  look also on my outdated comment of `src/environments/environment.prod.ts

README.md
dogi: thanks @mappuji and @lmmrssa but since non of this changes in `README.md` have nothing to do with the feature of this PR please ignore that for right now

mappuji: gettext

mappuji:
@dogi could you try again, I add another commit

src/environments/environment.prod.ts
mappuji: We change it dynamically via shell script @dogi

mappuji:
@lmmrssa could you also take a look? Basically we need to add `MULTIPLE_IPS=true` in the environment variable of docker-compose file

Issue #476, opened 2018-03-06
a re-enabled user is nearly loosing all the old assigned roles


dogi:
original roles:
![screen shot 2018-03-05 at 10 44 41 pm](https://user-images.githubusercontent.com/13684059/37013308-32d38dfe-20c7-11e8-99ca-9a1b468fdefc.png)

disabled:
![screen shot 2018-03-05 at 10 44 54 pm](https://user-images.githubusercontent.com/13684059/37013321-42fc3d70-20c7-11e8-8d57-748cb98605bc.png)

re-enabled:
![screen shot 2018-03-05 at 10 45 06 pm](https://user-images.githubusercontent.com/13684059/37013366-7a4169cc-20c7-11e8-96eb-54a0e391e3bf.png)

lmmrssa:
I understand the concern of disabling and enabling user losing the roles. But @dogi the solution will only work when you are enabling and disabling without navigating page (i.e accidental disable)
Later it will not save state. Instead couldn't we approach with giving confirmation before we disable saying that `all roles will be removed so do you wan to continue?`

lucy811:
The latest commit only work on without navigating page.

lucy811:
#601

PR #484, opened 2018-03-06
[#476] Fix a re-enabled user is nearly loosing all the old assigned roles

lucy811:


lmmrssa:
I understand the concern of disabling and enabling user losing the roles. But @dogi the solution will only work when you are enabling and disabling without navigating page (i.e accidental disable)
Later it will not save state. Instead couldn't we approach with giving confirmation before we disable saying that `all roles will be removed so do you wan to continue?`

Issue #473, opened 2018-03-05
Disable add icon of library, courses and meetups if user role is not defined


MacNew:

paulbert:
Note after closing the old PR:

The app should be reacting to the security settings on the CouchDB.  We should turn off buttons if the user does not have access to the database by checking database security & if the user has write access to the database.

paulbert:
Duplicate of #3023

PR #485, opened 2018-03-07
[#473] disable add icon button of those users, where role is not defined (connects #473 )

MacNew:


lmmrssa:
instead of creating variables and putting condition on error of fetch, we shall look after roles of user and just add that condition on add button.
Also we can then stop couchDB call as well to prevent from unnecessary request, when user as inactive

src/app/shared/auth-guard.service.ts
```diff
@@ -40,13 +40,14 @@ export class AuthService {
...
   // For login route will redirect to main app if there is an active session
   canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean> {
-    return this.couchService.get('_session', { withCredentials: true }).pipe(map(res => {
-      if (res.userCtx.name) {
-        this.router.navigate([ '' ]);
-        return false;
-      }
-      return true;
-    }));
+    if (route.url.toString() === 'login') {
```
paulbert: We do not need to check the route here, because this only applies to the routes we place it on in the router files.

src/app/shared/auth-guard.service.ts
```diff
@@ -40,13 +40,14 @@ export class AuthService {
...
+      return this.couchService.get('_session', { withCredentials: true }).pipe(map(res => {
+        if (res.userCtx.name) {
+          this.router.navigate([ '' ]);
+          return false;
+        }
+        return true;
+      }));
+    }
   }
-
```
paulbert: Please put this space back in, there should be no changes to this file in the PR.

src/app/shared/role-guard-service.ts
```diff
@@ -0,0 +1,20 @@
...
+import { UserService } from './user.service';
+import { Router, CanActivate, ActivatedRouteSnapshot, RouterStateSnapshot } from '@angular/router';
+
+@Injectable()
+export class RoleService {
+
+  constructor(private couchService: CouchService, private userService: UserService, private router: Router) {}
+
+  canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): any {
+    if (route.url.toString() === 'add') {
```
paulbert: We do not need to check the route here, because this only applies to the routes we place it on in the router files.

src/app/shared/role-guard-service.ts
```diff
@@ -0,0 +1,20 @@
...
+      if (this.userService.get().roles.length !== 0 || this.userService.get().isUserAdmin) {
```
paulbert: As I wrote in my previous feedback, this approach will cause us a lot of headaches if we change security settings on different databases.  We should base route access off the security that are currently on CouchDB, not hardcoded in the app like this.

mappuji:
hints:

Issue #477, opened 2018-03-06
Remove Object.assign(this, this.userService.get())


paulbert:
We should replace `Object.assign(this, this.userService.get());` and only assign the necessary properties from the userService to components.

I had introduced this earlier and realize it's not a good idea because we're assigning a bunch of properties to the component classes implicitly, some with very generic names.  Those properties could get overwritten in the future easily.

PR #487, opened 2018-03-07
[477] removed ObjectAssign (connects #477)

BasnetSamriddhi:


lmmrssa:
this issue is removing objectassign so that it does not create variables of all fields of user. As we are changing can't we just create a variable which will use details of user instead of assigning each one to `this` again

PR #488, opened 2018-03-07
Add filenames to resources mock data (fixes #460)

### LINKED ISSUE #460: Add filenames to resources mock data
The resources mock data do not have `filename` fields.  Please set these to the matching filenames in the attachment json file.

They can be found in the /design/resources directory.

YangLou1:
fileName fields has been added to resources mock data according to corresponding filenames in the attachment json file.

### DIFF
--- a/design/resources/resources-mockup.json
+++ b/design/resources/resources-mockup.json
@@ -29,6 +29,7 @@
             "year": "2014",
             "title": "Image Resource",
             "averageRating": "1",
+            "fileName": "image.png",
             "mediaType": "image",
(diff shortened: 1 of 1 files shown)


design/resources/resources-mockup.json
```diff
@@ -29,6 +29,7 @@
             "year": "2014",
             "title": "Image Resource",
             "averageRating": "1",
+            "fileName": "image.png",
```
nirojdyola: We have used **filename** instead of **fileName** in resources-add.component.ts
So, could you plesae change it to filename ?

YangLou1: @nirojdyola Thanks for reminding me. I have changed them to filename.

PR #491, opened 2018-03-08
[#490] Using Directives to make a username as lower-case (connects #490)

MacNew:


lmmrssa:
couldn't we use some masking or validation to prevent from adding capital username instead of forcing it on save.
Also what issue is pointing and what the solution is done does not seem to relate eachother. So may be we should change the issue title

nirojdyola:
Yes, you can refer this link https://medium.com/@firecube/angular-2-directives-to-format-text-as-you-type-e2c8e3ca09d8

src/app/login/login.module.ts
```diff
@@ -7,13 +7,14 @@ import { LoginRouterModule } from './login-router.module';
...
 
 @NgModule({
   imports: [
-    LoginRouterModule, FormsModule, CommonModule, MaterialModule, ReactiveFormsModule, PlanetFormsModule
-  ],
+    LoginRouterModule, FormsModule, CommonModule, MaterialModule, ReactiveFormsModule, PlanetFormsModule  ],
   declarations: [
-    LoginComponent, LoginFormComponent, ConfigurationComponent
+    LoginComponent, LoginFormComponent, ConfigurationComponent, LowerCase
+
```
nirojdyola: Extra space.Could you please remove extra space ?

src/app/login/login.module.ts
```diff
@@ -7,13 +7,14 @@ import { LoginRouterModule } from './login-router.module';
 import { MaterialModule } from '../shared/material.module';
 import { PlanetFormsModule } from '../shared/planet-forms.module';
 import { LoginFormComponent } from './login-form.component';
+import { LowerCase } from './login-form.directive'
 
 @NgModule({
   imports: [
-    LoginRouterModule, FormsModule, CommonModule, MaterialModule, ReactiveFormsModule, PlanetFormsModule
-  ],
+    LoginRouterModule, FormsModule, CommonModule, MaterialModule, ReactiveFormsModule, PlanetFormsModule  ],
```
nirojdyola: No changes before and after.

src/app/login/login-form.directive.ts
```diff
@@ -0,0 +1,12 @@
+import { Directive, ElementRef } from '@angular/core';
```
nirojdyola: Can we make this directive as a service or any global file , so that we can reuse this .

src/app/login/login-form.directive.ts
```diff
@@ -0,0 +1,12 @@
...
+
+@Directive({
+    selector: '[lowercase]'
+})
+export class LowerCase {
+  constructor(public el: ElementRef) {
+    this.el.nativeElement.onkeypress = (evt) => {
+      (evt.which < 65 || evt.which > 90) ? "" : evt.preventDefault();
```
nirojdyola: it only prevents key code. Actually we need to change it to lowecase, needn't we ?

src/app/shared/lowercase.ts
```diff
@@ -0,0 +1,25 @@
+import {Directive, Input, Output, EventEmitter, OnInit} from '@angular/core';
+
+@Directive({
+  selector: '[lowercase]',
+  host: {
+    '[value]': 'lowercase',
```
paulbert: Let's use HostListener & HostBinding rather than the host property of the decorator.

See: https://angular.io/guide/styleguide#hostlistenerhostbinding-decorators-versus-host-metadata

Issue #492, opened 2018-03-08
User roles added to unselected users


paulbert:
On `users.component.ts`

PR #496, opened 2018-03-09
[#492] User roles added to unselected users

lucy811:


nirojdyola:
Tested on NO role user
1. Click lock/ unlock button
2. learner role assigned
3. Now , click cross button in learner(means removing learner role from the user)
4. click on lock/unlock button, nothing happen(role will not assign back again)

lucy811:
In step 3, call the deleteRole( ) method, the user.roles.length === 0 (not undefined) at this time, 
In step 4, call the addRole(user) method, because this.selectedRolesMap.get(user.name) !== undefined, so selectedRolesArray won’t get [‘learner’].

paulbert:
@nirojdyola Thanks for testing and finding that.  It's a separate issue, so I'll make one for it.

Issue #504, opened 2018-03-13
Add resources to myLibrary


paulbert:
Selecting resources & clicking "Add to Library" or opening the three dot menu of a resource and clicking "Add to Library" should update the user's document adding the resource id to an array in a myLibrary property for the user.

This should consider the case that there isn't a myLibrary property, and add it to the new array.

PR #509, opened 2018-03-15
[#504] Added resources to myLibrary (connects #504)

Rupesh87:


nirojdyola:
@paulbert I think the data needs to be insert in **shelf** database instead of **_users**. In Bell-App, we have saved the data in separate document named : **shelf**

mappuji:
hints:

src/app/resources/resources.component.ts
```diff
@@ -187,4 +206,44 @@ export class ResourcesComponent implements OnInit, AfterViewInit {
...
+        }),
+        // If there are no matches, CouchDB throws an error
+        // User has no "shelf", and it needs to be created
+        catchError(err => {
+          // Observable of continues stream
+          return of({ rev: {}, resourceIds: [] });
+        }),
+        switchMap(data => {
+          const resourceIds = resourceIdArray.concat(data.resourceIds).reduce(this.dedupeShelfReduce, []);
+          return this.couchService.put('shelf/' + this.userService.get()._id,
```
lmmrssa: it is using user_id as shelf id. we should leave shelf id and just add user_id as another key

Issue #508, opened 2018-03-15
Resource Feedback


nirojdyola:
Please add feedback for resource. When Feedback button clicks , it needs to add feedback from the user.

PR #512, opened 2018-03-15
[#508] Feedback added to resource(connects #508)

BasnetSamriddhi:


lmmrssa:
we have feedback directive so same can be called without creating new form. We can just add some of the details pre-filled on the form

src/app/feedback/feedback.directive.ts
```diff
@@ -20,13 +20,15 @@ export class Feedback {
   source: string;
   url: string;
   messages: Array<Message>;
+  Resource_title: string;
```
nirojdyola: I think we need to define it in such a way that it supports to all. Suppose for course feedback , it will not valid. So , could you please check it once ?

mappuji:
Please add karma test if possible https://martinfowler.com/bliki/TestDrivenDevelopment.html

src/app/feedback/feedback.module.ts
```diff
@@ -0,0 +1,10 @@
+import { NgModule } from '@angular/core';
```
paulbert: If there were many imports in this module it would make sense to do this, but rather than create a module with only FeedbackDirective in it you should simply declare the FeedbackDirective in the ResourcesModule

BasnetSamriddhi: @paulbert  as I am adding the directive in two modules (Home module and Resource module ) I am getting the following error, could you suggest me a better way of adding it multiple modules. 
directive_error

paulbert: Right, forgot about that part.  I still do not think it's good to have such a small module, so let's add this directive to the DialogsModule instead.

src/app/resources/resources.component.html
```diff
@@ -52,7 +52,7 @@ <h3 class="header"><a *ngIf="!nationName" [routerLink]="['/resources/view', elem
             <mat-icon>library_add</mat-icon>
             <span i18n>Add to Library</span>
           </a>
-          <a mat-menu-item [routerLink]="['/resources/wip']">
+          <a mat-menu-item planetFeedback [value]="[{'state' : 'resources', 'item' : element._id}]" i18n-title title="Feedback">
```
paulbert: `value` does not need to be an array.  Please just send in an object.

src/app/feedback/feedback.directive.ts
```diff
@@ -45,7 +47,8 @@ export class FeedbackDirective {
         ...feedbackInfo,
         openTime: Date.now(),
         messages: [ startingMessage ],
-        url: this.router.url
+        url: this.router.url,
+        ...this.value[0]
```
paulbert: Once this is changed to an object you can remove `[0]`

src/app/feedback/feedback.directive.ts
```diff
@@ -27,6 +27,7 @@ export class Feedback {
 })
 export class FeedbackDirective {
   message: string;
+  @Input() value: any;
```
paulbert: - `value` is a vague variable name that could confuse because it also has specific meaning on certain HTML elements.  Please call this `feedbackOf`
- To avoid having to check if this is undefined, you can give it a default value of `{}` by just assigning it: `@Input() feedbackOf: any = {}`

src/app/feedback/feedback.directive.ts
```diff
@@ -37,6 +38,7 @@ export class FeedbackDirective {
 
   addFeedback(post: any) {
     this.message = '';
+    this.value = (this.value !== undefined) ? this.value : ' ';
```
paulbert: Setting a default means you can remove this line

PR #518, opened 2018-03-16
fix hardcoded db init

mappuji:
@lmmrssa 

fix the problem of dynamic db init address


docker/rpi-planet-dev.yml
```diff
@@ -11,7 +11,7 @@ services:
     depends_on:
       - couchdb
     environment:
-      - COUCHDB_HOST=http://couchdb:598
```
mappuji: thanks

Issue #511, opened 2018-03-15
Profile Image Upload View


nirojdyola:
Usually the user forget which file image was uploaded, so I think we need profile image review(shown in 2) in profile edit/add page 
Eg;
1. ![screen shot 2018-03-15 at 2 44 03 pm](https://user-images.githubusercontent.com/7495421/37453591-0fc70c12-2830-11e8-9f19-c56dffd210b0.png)

And we can use some design part like 
2. ![screen shot 2018-03-15 at 2 41 10 pm](https://user-images.githubusercontent.com/7495421/37453608-1a233f5a-2830-11e8-9360-3f549af37d60.png)

PR #521, opened 2018-03-19
[#511] Profile Image Upload View

lucy811:


nirojdyola:
screen shot 2018-03-19 at 11 30 44 am

1. new registration
2. Already there is image which is wrong.
3. **File Upload** is not necessary. You can label it  as Select image  button and can trigger the input file 
4. Change button is not working ,it must allow the user to change the image.

For example , you can check it in http://keenthemes.com/preview/metronic/theme/admin_2/components_bootstrap_fileinput.html

src/app/users/users-update/users-update.component.html
```diff
@@ -58,13 +58,26 @@
...
         </mat-form-field>
-        <label>File Upload:</label>
-        <input type="file" (change)="bindFile($event)" />
       </div>
       <div>
         <button type="submit" mat-raised-button color="primary" i18n>Update</button>
         <button type="button" mat-raised-button color="warn" (click)="goBack()" i18n>Cancel</button>
       </div>
     </form>
+    <div class="profile-upload-view">
```
paulbert: Rather than create new styles for this, we should use the same styles as the image for UsersProfileComponent

src/app/users/users-update/users-update.component.html
```diff
@@ -58,13 +58,26 @@
...
+      <img class="profile-upload-image" src="">
+      <br>
```
paulbert: A `<br>` tag should not be used in this case.  It has a specific use for breaking lines of text in poetry & addresses.  Instead you could wrap the button in a block level (div) element.

src/app/users/users-update/users-update.component.ts
```diff
@@ -104,7 +128,28 @@ export class UsersUpdateComponent implements OnInit {
     this.router.navigate([ '/users/profile', this.user.name ]);
   }
 
-  bindFile(event) {
+  previewFile(event) {
+    const preview = <HTMLImageElement>document.querySelector('.profile-upload-image');
```
paulbert: This is not a good way to change the image src in Angular.  Instead you should bind the value to a variable on the class, the html would be like `[src]="previewSrc"`

src/app/users/users-update/users-update.component.html
lucy811: "profile-upload-view" style keep image and button at the right of form, I change "profile-upload-image" style to "profile-image-large"(the same style as the image for usersProfileComponent).

lucy811:
input[type=file] in <button> doesn't work on firefox, so I use <label> instead.

Issue #523, opened 2018-03-19
login/register - undefined lowercase


razu9861:
![screen shot 2018-03-19 at 11 04 46 am](https://user-images.githubusercontent.com/24647983/37579331-82ce8980-2b35-11e8-9ab2-c86c23e8ddac.png)

PR #525, opened 2018-03-19
[#523] login undefined lowercase fixed

MacNew:


paulbert:
So I removed that binding because I thought it wasn't necessary, but it looks like it needed a few more changes.

- Now to implement this directive you only need to add the `planetLowercase` attribute to the input tag, no need for the `[(lowercase)]=...`
- Part of the reason for removing this was that the two-way binding was not implemented correctly. (see the @Output eventEmitter in the [documentation](https://angular.io/guide/template-syntax#two-way-binding---)
- This also removes the `toLowerCase()` uses immediately before the user is created.  Since we already force the input value to lowercase, we do not need this.  Removing them will make updating easier if we decide to allow uppercase letters in the future.
- Lastly, I realized that this directive was not added in the configuration.  That meant the admin could create a user name with uppercase letters and then would not be able to log in because the login form only allowed lowercase.

Issue #524, opened 2018-03-19
Profile Login Field


nirojdyola:
Please check the profile of any user, and check the login field. It should not be editable as login name is already taken . That's why we must not allow the user to change it again .

PR #527, opened 2018-03-19
[#524] make user login name disabled (connects #524)

razu9861:


src/app/users/users-update/users-update.component.ts
```diff
@@ -55,8 +55,7 @@ export class UsersUpdateComponent implements OnInit {
       firstName: [ '', Validators.required ],
       middleName: '',
       lastName: [ '', Validators.required ],
-      name: '',
-      login: [ { value: '', disabled: true }, Validators.required ],
+      name: [ { value: '', disabled: true }, Validators.required ],
```
lmmrssa: this field is actually not required as we are appending form value over this.user which already has the name in it.

src/app/users/users-update/users-update.component.html
```diff
@@ -3,7 +3,10 @@
 </mat-toolbar>
 
 <div class="space-container">
-  <mat-toolbar class="primary-color font-size-1" i18n>Edit Your Profile</mat-toolbar>
+  <mat-toolbar class="primary-color font-size-1">
+    <span>{{user.name}}</span><span class="toolbar-fill"></span>
+    <span class="margin-lr-3">Edit Profile</span>
```
lmmrssa: this code is not required and also will not look nice as it is just text which shows in the place used for button

Issue #528, opened 2018-03-19
Should notify user that password sucessfully changed


MacNew:

paulbert:
Closing as this was fixed in #549

PR #530, opened 2018-03-19
[#528] notify password change (connect #528)

MacNew:


src/app/users/users-profile/users-profile.component.ts
```diff
@@ -68,7 +70,9 @@ export class UsersProfileComponent implements OnInit {
       return of({ ok: false, reason: 'Error changing password' });
     })).subscribe((res) => {
       if (res.ok === true) {
-        // TODO: Should notify user that password successfully changed or that there was an error
+        this.planetMessageService.showAlert('Password sucessfully changed');
+      } else {
+        this.planetMessageService.showAlert('There was an error');
```
lmmrssa: this could show res.reason on message as it already has message return from `return of({ ok: false, reason: 'Error changing password' });`

MacNew: OK

Issue #531, opened 2018-03-19
Multiple selection value in Resource form is string instead of array (for previously saved resources in DB)


adepojubosun:
![screenshot from 2018-03-19 10-53-19](https://user-images.githubusercontent.com/7988475/37603788-c6b05a18-2b65-11e8-9efd-5152cc7f7ace.png)

![screenshot from 2018-03-19 11-33-04](https://user-images.githubusercontent.com/7988475/37605572-da9279cc-2b69-11e8-9a60-794e17334e5d.png)

**_Error: Value must be an array in multiple-selection mode_**
This error is thrown if you attempt to assign a value other than null, undefined, or an array to a <mat-select multiple>. 

Default resources in database need clarification. All multiple selection values in the database are string values instead of array. 
This issue is required to complete issue #498

paulbert:
Good catch, we should change this in the mock data in /design/resources

lmmrssa:
change "resourceFor": "Learner" to array

Issue #498, opened 2018-03-12
Edit resource after creating


paulbert:
Add a "manage" link in the menu which leads to an "AddResource" component, but for updating values.

PR #532, opened 2018-03-19
[#498] Edit/Manage resource after creating (connects #498)

adepojubosun:
To update resources already available in couchdb, issue #531 needs to be solved


src/app/resources/resources-add.component.ts
```diff
@@ -51,6 +54,19 @@ export class ResourcesAddComponent implements OnInit {
...
   ngOnInit() {
     this.userDetail = this.userService.get();
+    // update resource url check
+    if (this.route.snapshot.url[0].path === 'update') {
+      this.couchService.get('resources/' + this.route.snapshot.paramMap.get('id'))
+      .subscribe((data) => {
+        this.revision = data._rev;
+        this.id = data._id;
+        console.log('this', data);
+        this.resourceForm.patchValue(data);
```
adepojubosun: Can only patch newly created resources. see issue #531

adepojubosun:
**TODO : Check for file reupload and replace instances of new file uploaded**

.gitignore
```diff
@@ -21,7 +21,7 @@
 .vscode/*
```
lmmrssa: this code should have ignored all other files inside .vscode so it should not have been required to add extra lines.

src/app/resources/resources-add.component.ts
```diff
@@ -30,19 +30,23 @@ export class ResourcesAddComponent implements OnInit {
   media: string[];
   openWith: string[];
   resourceType: string[];
-  currentDate = new Date();
+  currentDate = new Date(); // might be forced to change date format using "toISOString().split('T')[0]"
```
lmmrssa: we were using timestamp in milliseconds so we could format it on rendering as required. So can we use same concept here

lmmrssa: We were using timestamp in milliseconds to save so that we could format it on rendering as required. Can we use same over here as well?

lmmrssa:
@adepojubosun could you make sure you do not commit from vagrant :)

src/app/resources/resources-add.component.ts
```diff
@@ -77,19 +95,19 @@ export class ResourcesAddComponent implements OnInit {
       addedBy: '',
       openUrl: [],
       openWhichFile: '',
-      isDownloadable: ''
+      isDownloadable: '',
```
paulbert: Extra comma

src/app/validators/validator.service.ts
```diff
@@ -38,25 +38,25 @@ constructor(private couchService: CouchService) {}
...
-    courseId: string
+    Id: string
   ): Observable<ValidationErrors | null> {
     return timer(500).pipe(
       switchMap(() => this.couchService.post(
         `${dbName}/_find`,
         findDocuments(
-          { 'courseTitle' : ac.value },
-          [ '_id', 'courseTitle' ]
+          { fieldName : ac.value },
```
paulbert: This object will have a key equal to "fieldName", not the value of the variable `fieldName`.  You need to surround fieldName in brackets (see [Computed property names](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Computed_property_names))

ArashDai:
the download buttons don't work perfectly when trying to download the attachments

lmmrssa:
This PR has been quite a old one. So either we can start new PR or need to rebase this PR.

Issue #536, opened 2018-03-21
Profile Image in Top Bar


nirojdyola:
After profile image is uploaded, please load image in top bar as well.

PR #538, opened 2018-03-21
[#536] Profile Image in Top Bar (connects #536)

nirojdyola:
screen shot 2018-03-21 at 10 18 45 am

@paulbert How to get value in real time ? I mean to say, when I change the profile Image , then it needs to update the top bar image as well. 
I checked **BehaviorSubject**, we can update **this.user** value in user service. So, please give me some suggestion  to follow

### DIFF
--- a/src/app/home/home.component.html
+++ b/src/app/home/home.component.html
@@ -10,7 +10,10 @@
     <img *ngIf="current_flag" src="assets/flags/{{current_flag}}.png" alt="{{current_lang}}"
       title="{{current_lang}}" />
   </button>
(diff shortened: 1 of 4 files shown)


paulbert:
It was easier to just implement than explain, so you can look at my commit for more detail.

Basically we needed to have a Subject & Observer to subscribe to on the UserService.  The Subject triggers any subscribed observables after UserService.set() runs, so components will then get the new information.

In addition we needed to update the UserService after the user updates their profile (this was an oversight when we implemented that).

Issue #543, opened 2018-03-22
travis collision? instead of waiting


dogi:
```
$ ssh -o StrictHostKeyChecking=no -p 22 travis@kraken.ole.org 'bash -s' -- < ./.travis/deploy_rpi.sh --branch="$TRAVIS_BRANCH" --commit="$TRAVIS_COMMIT" --pull="$TRAVIS_PULL_REQUEST" --duser="$DOCKER_USER" --dpass="$DOCKER_PASS" --gtag="$TRAVIS_TAG" --image=planet
Warning: Permanently added 'kraken.ole.org,204.9.221.20' (ECDSA) to the list of known hosts.
fatal: could not create work tree dir 'planet-81616070'.: File exists
The command "ssh -o StrictHostKeyChecking=no -p 22 travis@kraken.ole.org 'bash -s' -- < ./.travis/deploy_rpi.sh --branch="$TRAVIS_BRANCH" --commit="$TRAVIS_COMMIT" --pull="$TRAVIS_PULL_REQUEST" --duser="$DOCKER_USER" --dpass="$DOCKER_PASS" --gtag="$TRAVIS_TAG" --image=planet" exited with 128.
```

mappuji:
Since we move entirely to #820

PR #547, opened 2018-03-24
[#543] implement flock and refactor shell script

mappuji:
implement flock and refactor shell script


mappuji:
Sorry need to close this one since it contaminated with other changes.

PR #548, opened 2018-03-24
[#537] Resign and admission added(connects #537)

BasnetSamriddhi:


src/app/courses/courses.component.html
```diff
@@ -41,13 +41,13 @@ <h3 class="header">{{element.courseTitle}}</h3>
...
           <mat-menu #resourceMenu="matMenu">
-            <a mat-menu-item routerLink="add">
+            <a *ngIf="element.resign" mat-menu-item routerLink="add">
               <mat-icon>lock_open</mat-icon>
               <span i18n>Open</span>
             </a>
-            <a mat-menu-item routerLink="add">
-              <mat-icon>clear</mat-icon>
-              <span i18n>Resign</span>
+            <a mat-menu-item (click)="toggleResign(element,element._id)">
```
nirojdyola: **element** contains _id,_rev and other fields . So , do we need two parameters here ?

src/app/courses/courses.component.ts
```diff
@@ -114,4 +114,20 @@ export class CoursesComponent implements OnInit, AfterViewInit {
     this.courses.data.forEach(row => this.selection.select(row));
   }
 
+  toggleResign(course, courseId) {
```
nirojdyola: Could not we use only one parameter, since course contains course's id .

src/app/dashboard/dashboard.component.ts
```diff
@@ -49,4 +50,13 @@ export class DashboardComponent implements OnInit {
       return response.rows.map((item) => ({ ...item.doc, title: item.doc[titleField], link: linkPrefix + (addId ? item.id : '') }));
     }));
   }
+
+  filterResigned(res) {
```
lmmrssa: @paulbert  I think we should go with mango query to get the list of courses or library. And not try to filter it on user side. It will be fetching lot of data when we will only be showing few.

src/app/dashboard/dashboard.component.ts
```diff
@@ -44,9 +45,21 @@ export class DashboardComponent implements OnInit {
...
+      }));
+    }
+
+  myCourse() {
+    return this.couchService.post(`courses/_find`,
+      findDocuments({ 'members':  {'$in': [
+        this.userService.get()._id
+        ]} }, 0 ))
+      .pipe(
+        map((resopnse => {
```
paulbert: Should be `response`, and extra set of parantheses

src/app/dashboard/dashboard.component.ts
```diff
@@ -34,8 +35,8 @@ export class DashboardComponent implements OnInit {
     this.getData('resources', { linkPrefix: 'resources/view/', addId: true }).subscribe((res) => {
       this.data.resources = res;
     });
-    this.getData('courses', { linkPrefix: 'courses', titleField: 'courseTitle' }). subscribe((res) => {
-      this.data.courses = res;
+    forkJoin(this.myCourse()).subscribe((res) => {
```
paulbert: Should be `this.myCourse().subscribe...`

src/app/dashboard/dashboard.component.ts
```diff
@@ -24,6 +24,7 @@ import { map } from 'rxjs/operators';
 })
 export class DashboardComponent implements OnInit {
   data = { resources: [], courses: [], meetups: [] };
+  courseArray = [];
```
paulbert: Not needed

src/app/dashboard/dashboard.component.ts
```diff
@@ -1,9 +1,9 @@
 import { Component, OnInit } from '@angular/core';
-
+import { findDocuments } from '../shared/mangoQueries';
 import { UserService } from '../shared/user.service';
 import { CouchService } from '../shared/couchdb.service';
-
-import { map } from 'rxjs/operators';
+import { forkJoin } from 'rxjs/observable/forkJoin';
```
paulbert: Once forkJoin is removed from courses observable, don't need to import

PR #552, opened 2018-03-27
Add arrow_back icon in communities list (fixes #551)

### LINKED ISSUE #551: Add arrow_back icon with routerLink manager in community.component.html
When a user click arrow_back icon in Communities List page, redirect the user to manager router link.

untitled

YangLou1:


nirojdyola:
Yes, it is correct way of adding arrow icon. But, the design is not correct .
screen shot 2018-03-27 at 9 08 21 am
It should be like this:
screen shot 2018-03-27 at 9 09 04 am

YangLou1:
@nirojdyola I have made the suggested change to the design of Communities List page, please take a look.

paulbert:
@YangLou1 It appears there was some miscommunication from the meeting last Monday.  I had said that you should not fully modify the entire template, but only make the "Communities List" on the top into a `mat-toolbar`.

The intent was to limit the updates to just the portion of the template which is relevant to the issue.  By changing more of the template, more changes will now be needed to fix this PR.

The build is failing, as you can see in [Travis](https://travis-ci.org/ole-vi/planet/jobs/360499610#L1834).  The development mode used by the `ng serve` command will ignore some errors that will cause the production build to fail.

This is because the communities table doesn't use a text input for its filter, it uses the two dropdowns.  You should remove the input (circled in red), and move the dropdowns (circled in yellow) in its place:
inkedcommunity_li

Issue #560, opened 2018-04-03
Create user shelf when creating user


paulbert:
We should create a document on the `shelf` database when creating a user with the same _id as the newly created user document.

PR #567, opened 2018-04-03
[#560] Create user shelf  while creating user (connects #560)

razu9861:


src/app/login/login-form.component.ts
```diff
@@ -70,6 +70,9 @@ export class LoginFormComponent {
           this.planetMessageService.showMessage('User created: ' + data.id.replace('org.couchdb.user:', ''));
           this.welcomeNotification(data.id);
           this.login(this.userForm.value, true);
+          this.couchService.put('shelf/' + data.id, { })
```
nirojdyola: subscribe under subscribe. Can we manage it ?

src/app/configuration/configuration.component.ts
```diff
@@ -106,7 +106,9 @@ export class ConfigurationComponent implements OnInit {
             'isUserAdmin': true, 'firstName': this.contactFormGroup.value.firstName, 'middleName': this.contactFormGroup.value.middleName,
             'lastName': this.contactFormGroup.value.lastName, 'email': this.contactFormGroup.value.email,
             'phoneNumber': this.contactFormGroup.value.phoneNumber }).subscribe((data1) => {
-             this.planetMessageService.showMessage('Admin created: ' + data1.id.replace('org.couchdb.user:', ''));
+              this.couchService.put('shelf/' + data1.id, { })
```
nirojdyola: subscribe under subscribe. Can we manage it ? And also don't use data1 , we can use other meaningful variable like result , res and other . Please check it once .

Issue #577, opened 2018-04-06
Use allDocs method of couch service instead of .get method for fetching data from CouchDb


MacNew:
Reference can be taken from [https://github.com/ole-vi/planet/tree/505-all-docs-function](url) #505

PR #581, opened 2018-04-06
[#577] Using allDocs instead of get method (Connect #577)

MacNew:


MacNew:
Replace allDocs methods on these file
home.component.ts
nation.component.ts
resources.service.ts
user.service.ts
users.component.ts

PR #583, opened 2018-04-06
Configure center address for docker

paulbert:
Allows us to create production versions of planet with different center/earth address quickly.


lmmrssa:
@paulbert  can we also make this change on rpi-planet.yml

PR #599, opened 2018-04-09
Adjust the layout of nation component (fixes 559)

YangLou1:


src/app/home/home.component.ts
```diff
@@ -82,7 +82,7 @@ export class HomeComponent implements OnInit, AfterViewInit, OnDestroy {
   // Used to swap in different background.
   // Should remove when background is finalized.
   backgroundRoute() {
-    const routesWithBackground = [ 'resources', 'courses', 'feedback', 'users', 'meetups', 'requests' ];
+    const routesWithBackground = [ 'resources', 'courses', 'feedback', 'users', 'meetups', 'requests', 'nation', 'community'];
```
nirojdyola: Do we need to add community here for the layout of nation ? Could you please check it once ?

YangLou1: @nirojdyola nation.component.html shows 'Nation' or 'Community' depending on parentType === 'nation'. When it shows 'Community', should we add community to adjust the page's layout to be consisted as well?

paulbert: The routes changed a little, so if you use 'associated' instead of 'nation' and 'community' it will work.

PR #601, opened 2018-04-09
[#476] a re-enabled user is nearly loosing all the old assigned roles

lucy811:
The latest commit only work without navigating page. In this pull request, I make this function work with navigating page.


lmmrssa:
When creating new vagrant setup of this branch. Login was giving 'Incorrect username/password' error but still allowing to go inside. But when trying to logout it gives following errors.

screen shot 2018-04-10 at 08 26 04
screen shot 2018-04-10 at 08 26 18

lucy811:
I tried the latest master branch (not my branch), created an admin and submitted. It was giving ‘There was an error creating planet’ error but still allowing the admin to go inside.

src/app/users/users.component.ts
```diff
@@ -107,11 +108,17 @@ export class UsersComponent implements OnInit, AfterViewInit {
     const tempUser = { ...user, roles: [ ...user.roles ] };
     tempUser.roles.splice(index, 1);
     this.selectedRolesMap.set(tempUser.name, tempUser.roles);
+    if (tempUser.roles.length === 0) {
+      tempUser.isUserLocked = true ;
```
paulbert: Extra space before semicolon (also on line 114)

src/app/users/users.component.ts
```diff
@@ -166,6 +173,7 @@ export class UsersComponent implements OnInit, AfterViewInit {
           user.roles.push(role);
           const res: any = responses.find((response: any) => response.id === user._id);
           user._rev = res.rev;
+          this.initializeData();
```
paulbert: Do not need to rerun initialize data as we are updating the necessary fields on the user documents

src/app/users/users.component.ts
```diff
@@ -107,11 +108,17 @@ export class UsersComponent implements OnInit, AfterViewInit {
...
+      tempUser.isUserLocked = true ;
+    } else {
+      tempUser.isUserLocked = false ;
+    }
     delete tempUser.selected;
     this.couchService.put('_users/org.couchdb.user:' + tempUser.name, tempUser).subscribe((response) => {
       console.log('Success!');
       user.roles.splice(index, 1);
       user._rev = response.rev;
+      this.initializeData();
```
paulbert: Do not need to rerun initialize data as we are updating the necessary fields on the user documents

lmmrssa:
@lucy811 that issue is caused by not being able to connect with parent planet which is a different issue. And I agree with @paulbert that keeping roles will cause issue as couchDB works on roles for permission. Else we will have to update the validation and auth design doc to deny for isUserLocked.

src/app/users/users.component.ts
```diff
@@ -120,22 +118,30 @@ export class UsersComponent implements OnInit, AfterViewInit {
...
-    if (this.selectedRolesMap.get(user.name) === undefined || this.selectedRolesMap.get(user.name).length === 0) {
-      selectedRolesArray = [ 'learner' ];
-    }
-    const tempUser = { ...user, roles: [ ...selectedRolesArray ] };
-    this.couchService.put('_users/org.couchdb.user:' + tempUser.name, tempUser).subscribe((response) => {
-      console.log('Success!');
-      this.initializeData();
+    let oldRoles: string[] = [];
+    let selectedRole: string[] = [];
+    this.couchService.get('_users/org.couchdb.user:' + user.name).subscribe((data) => {
```
paulbert: We already have the user information from `initializeData()` stored on the component and passed in as the `user` variable, so we do not need to query the CouchDB before getting it.

src/app/users/users.component.ts
```diff
@@ -120,22 +118,30 @@ export class UsersComponent implements OnInit, AfterViewInit {
...
+      oldRoles = data.oldRoles;
+      if ( oldRoles.length === 0 ) {
+        selectedRole = [ 'learner' ];
+      } else {
+        selectedRole = oldRoles;
+      }
+      const tempUser = { ...user, roles: [ ...selectedRole ] };
+      this.couchService.put('_users/org.couchdb.user:' + tempUser.name, tempUser).subscribe((response) => {
```
paulbert: Based on my other feedback, there should only be one couchService request, but for future reference:

We should not nest observable subscribes.  We can use the rxjs operator `switchMap` ([documentation](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-switchMap)) to chain observables in a sequence if needed.

src/app/users/users.component.ts
```diff
@@ -102,12 +101,15 @@ export class UsersComponent implements OnInit, AfterViewInit {
     });
   }
 
-  deleteRole(user: any, index: number) {
+  deleteRole(user: any, index: number, position: number) {
     // Make copy of user so UI doesn't change until DB change succeeds
-    const tempUser = { ...user, roles: [ ...user.roles ] };
+    let tempUser = { ...user, roles: [ ...user.roles ] };
```
lmmrssa: i'm not getting code as it seems to update tempUser roles from user.roles which should have already been there when you make tempUser equals to user

src/app/users/users.component.ts
```diff
@@ -102,12 +101,15 @@ export class UsersComponent implements OnInit, AfterViewInit {
     });
   }
 
-  deleteRole(user: any, index: number) {
+  deleteRole(user: any, index: number, position: number) {
     // Make copy of user so UI doesn't change until DB change succeeds
-    const tempUser = { ...user, roles: [ ...user.roles ] };
+    let tempUser = user;
```
paulbert: Please keep this the same.  We need to create a deep copy of the user object so the view does not get updated before the HTTP request to CouchDB is successful.

src/app/users/users.component.ts
```diff
@@ -102,12 +101,15 @@ export class UsersComponent implements OnInit, AfterViewInit {
...
+  deleteRole(user: any, index: number, position: number) {
     // Make copy of user so UI doesn't change until DB change succeeds
-    const tempUser = { ...user, roles: [ ...user.roles ] };
+    let tempUser = user;
     tempUser.roles.splice(index, 1);
-    this.selectedRolesMap.set(tempUser.name, tempUser.roles);
     delete tempUser.selected;
+    if (tempUser.roles.length === 0) {
+      tempUser = { ...tempUser, oldRoles: [] };
+      this.allUsers.data.forEach(row => this.updateOldRole(row, position, tempUser));
```
paulbert: We should not update the `users` variable or the `this.allUsers` array until the `subscribe`, and we don't need to iterate over the entire array since the `users` variable still refers back to the object in that array.

src/app/users/users.component.ts
paulbert: @lmmrssa This is a key part of JavaScript: deep vs. shallow copy. `...` is one way for doing a shallow copy of an object/array.  In this situation since the `roles` property is also an object/array we need to also use `...` on that object/array to create a copy.

Issue #603, opened 2018-04-10
Adding key value to array


paulbert:
The `notifications` array in `home.component.ts` has a value assigned to an object-like key in the `getNotification()` method.  We should not do this, so this should be refactored.

PR #607, opened 2018-04-10
[#603] Do not add key value to array (connect #603)

MacNew:


src/app/home/home.component.ts
```diff
@@ -142,9 +142,9 @@ export class HomeComponent implements OnInit, AfterViewInit, OnDestroy {
           }
           return notifications;
         }).filter(nt  => {
-          return nt['user'] === user_id;
+          return nt.user === user_id;
         });
-        this.notifications['count_unread'] =  cnt;
+        this.notifications.count_unread = cnt;
```
lmmrssa: if i'm correct. Notifications are called by couchdb alldocs which will return array not object. So this.notifications will be assigned by normal array of documents. And this code is still trying to mix array and object under notifications.

paulbert: Yes, this is correct.  Thinking about this more, we do not even use this count in the component.  We should remove it until we do.

src/app/home/home.component.ts
```diff
@@ -30,7 +30,7 @@ export class HomeComponent implements OnInit, AfterViewInit, OnDestroy {
   current_flag = 'en';
   current_lang = 'English';
   sidenavState = 'closed';
-  notifications = [];
+  notifications = { count_unread: 0 };
```
paulbert: This is adding to the problem, as you later assign an array to this.notifications.  It should just be initialized as an array.

src/app/home/home.component.ts
```diff
@@ -134,23 +134,31 @@ export class HomeComponent implements OnInit, AfterViewInit, OnDestroy {
...
-          if (notifications.status === 'unread') {
-            cnt ++;
-          }
-          return notifications;
+         return Object.keys(notifications).map(key => {
+            return { [ key ] : notifications[ key ] };
+          });
         }).filter(nt  => {
-          return nt.user === user_id;
+          return nt[ 2 ].user === user_id;
```
lmmrssa: now you have started to add number as key. I think the problem is not clear with you.

paulbert:
I updated the title, which was maybe part of the confusion.  In the comment on the issue it says we should not have key values added to an array.  The title referred to the problem, not the task to be done.

lmmrssa:
@paulbert @MacNew  created new PR https://github.com/ole-vi/planet/pull/612 as this PR seems to have taken different approaches. So closing this PR.

PR #612, opened 2018-04-11
[#603] remove object and array mixed

lmmrssa:
Added new variable to assign notification unread count. But still i don't see it was used anywhere. Will keep that as next issue to work on as this issue doesn't mention about showing unread count.


paulbert:
First off, we should stick to camelCase for variable names (see #620).  Second, since we don't use the count in the app at all, I think we should just remove this code until we do.

Issue #613, opened 2018-04-11
Meetup Details


nirojdyola:
More details need to add in meetup list.
For example, we need to add start date-end date, etc.

PR #614, opened 2018-04-11
[#613] Add meetup details on meetups.component.html (connect #613) 

MacNew:


src/app/meetups/meetups.component.html
```diff
@@ -70,6 +70,16 @@ <h3 class="header">
         </mat-cell>
       </mat-cell>
     </ng-container>
+    <ng-container matColumnDef="action">
+      <mat-header-cell *matHeaderCellDef i18n> Information </mat-header-cell>
+      <mat-cell *matCellDef="let element" class="list-content-menu">
+        <div>
+          <p>{{"Date: " + (element.startDate | date: 'fullDate') + " - " + (element.endDate | date: 'fullDate') + " at " + element.startTime + ' - ' + element.endTime}}</p>
```
nirojdyola: @paulbert Should we use translation here ? Since it includes both dynamic and static value .

paulbert: We cannot without a change because it is all in an interpolation ({{...}}) block.  To use translation it needs to be broken up like this:
```
<p><span i18n>Date:</span>{{ ' ' + (element.startDate | date: 'fullDate') + ' - ' + (element.endDate | date: 'fullDate') + ' '}}<span i18n>at</span>{{ ' ' + element.startTime + ' - ' + element.endTime}}</p>
```
Notice that I keep the spaces in the interpolation statements.  Leaving a trailing space would risk the translator remove it for the other languages.

The same change is needed for the next two lines.

src/app/meetups/meetups.component.ts
```diff
@@ -20,12 +20,15 @@ import { Subject } from 'rxjs/Subject';
...
+    .mat-column-action {
+      max-width: 225px;
+    }
   ` ]
 })
 export class MeetupsComponent implements OnInit, AfterViewInit, OnDestroy {
 
   meetups = new MatTableDataSource();
-  displayedColumns = [ 'select', 'title' ];
+  displayedColumns = [ 'select', 'title', 'action' ];
```
lmmrssa: this section is not used for action instead it is used for detail

lmmrssa:
code required rebase also it had few unnecessary codes

PR #616, opened 2018-04-11
[#541] making learner readonly for nations (connects #541)

lmmrssa:


lmmrssa:
added the function in validation rule to not allow user to update. As validation is called only on new entry and update it will invalidate on write @paulbert could you review shall we go this way

design/nations/nation-validators.js
```diff
@@ -5,6 +5,24 @@ function (newDoc, oldDoc, userCtx, secObj) {
...
   }
+  function canWrite() {
+    var CAN_WRITE = true;
+    var readOnly = ['learner'];
+    if (!userCtx.isUserAdmin) {
+      if (userCtx.roles.length < 1) {
+        CAN_WRITE = false;
+      } else {
+        for (var i = 0; i < userCtx.roles.length; i++) {
+          if (~ readOnly.indexOf(userCtx.roles[i])) {
```
paulbert: This is quicker to understand with === -1

design/nations/nation-validators.js
```diff
@@ -5,6 +5,24 @@ function (newDoc, oldDoc, userCtx, secObj) {
...
+            CAN_WRITE = false;
+          }
+        }
+      }
+    }
+    if(!CAN_WRITE)
```
paulbert: Even for a one line block, it's better (more readable, doesn't risk someone writing another line & popping that statement out of the if clause) to have brackets.

design/nations/nation-validators.js
```diff
@@ -5,6 +5,24 @@ function (newDoc, oldDoc, userCtx, secObj) {
       throw { forbidden: errMessage };
     }
   }
+  function canWrite() {
+    var CAN_WRITE = true;
+    var readOnly = ['learner'];
+    if (!userCtx.isUserAdmin) {
```
paulbert: I don't think user profile properties are on a userCtx object.

lmmrssa: had tested it and it was working by that code.

design/nations/nation-validators.js
lmmrssa: oops yes i had noticed it but forgot to push that change :) thanks for catching

Issue #563, opened 2018-04-03
Update dashboard view


paulbert:
Something between:

![image](https://user-images.githubusercontent.com/9203229/38225329-69c6455c-36c2-11e8-84c6-bfb5900cbc54.png)

and

![image](https://user-images.githubusercontent.com/9203229/38225298-4c2bd34a-36c2-11e8-89ac-33bedd5937ab.png)

PR #622, opened 2018-04-12
[#563] Update dashboard view (connects #563)

paulbert:
I can revert to the white/grey alternating for the lists, but since some of the designs have used small lines to divide instead tried that to see how it looks.

The left menu is just a placeholder for now, and on tablets it doesn't fit (it is also scrollable, however):

screenshot_2018-04-11-20-32-36

### DIFF
--- a/src/app/_variables.scss
+++ b/src/app/_variables.scss
@@ -32,6 +32,9 @@ $screen-sm: 780px;
 
 // Maximum full height of view container (avoiding vertical overflow)
 $view-container-height: calc(100vh - 352px);
+// Height of dashboard
+$dashboard-height: calc(100vh - 128px);
+$dashboard-tile-height: calc((#{$dashboard-height} - 3rem) / 4);
 
 /*
  * Old Planet Material Colors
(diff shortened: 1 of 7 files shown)


lmmrssa:
There has not been change on login page but the travis is failing on login test. Not sure why is that.

screen shot 2018-04-12 at 12 00 25

paulbert:
I didn't have time to look into this too deeply last night.  I'm confused why the tests are failing on Travis without changing those pages.  The tests pass locally, so this is difficult to debug.

paulbert:
Saving feedback: myLibrary, etc. should not be same size as "Planet Learning"

paulbert:
Looking at this more, the myLibrary, etc. and Planet Learning are both set to the normal Material header font size.  We should increase the size of Planet Learning since it is our main header rather than have to change a bunch of headers in our other views.  See #852

lmmrssa:
arrows has been replaced by scrollbar but arrow function are still not removed.

lmmrssa:
@paulbert I found most of codes from tile are not required so I have tried to remove them all please go through changes once and see any of those code has any need to be there? I have tested view and it looks nice.

paulbert:
Looks good, thanks!

PR #637, opened 2018-04-12
[#606] Request member to join the meetup (connect #606)

MacNew:


MacNew:
@paulbert   I am not able to fix Paginator. Please provide me suitable suggestion.

src/app/meetups/invitation-meetups/meetups-invitation.service.ts
```diff
@@ -0,0 +1,19 @@
+import { Observable } from 'rxjs/Rx';
```
razu9861: i think this can go in meetup.service.ts, all code form this file can be handle from one service rather than creating multiple service in same folder

Issue #625, opened 2018-04-12
Adding new resource - date format too long 


Tille88:
When adding a resource, the non-editable datestamp that is generated could be a lot shorter, while still being flexible in terms of locale.

<img width="593" alt="screen shot 2018-04-07 at 22 04 41" src="https://user-images.githubusercontent.com/16757941/38667064-adae88d8-3e6a-11e8-88b6-410fe218b1ad.png">

Tille88:
Looks great (also see below for reference to change), 

Since I'm new though, why is the tests failing? I guess it's not custom to give official approval unless theres a green tick?

<img width="674" alt="screen shot 2018-04-13 at 14 33 53" src="https://user-images.githubusercontent.com/16757941/38722322-caf2f04c-3f27-11e8-99ac-6db975e6bfe9.png">

PR #644, opened 2018-04-13
[#625] resource article date shortend (connects #625)

razu9861:
screen shot 2018-04-13 at 11 07 06 am

### DIFF
--- a/src/app/resources/resources-add.component.html
+++ b/src/app/resources/resources-add.component.html
@@ -75,7 +75,7 @@
           </mat-select>
         </mat-form-field>
         <mat-form-field>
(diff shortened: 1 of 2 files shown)


src/app/resources/resources-add.component.html
```diff
@@ -75,7 +75,7 @@
           </mat-select>
         </mat-form-field>
         <mat-form-field>
-          <input matInput [(ngModel)]="currentDate" formControlName="articleDate" i18n-placeholder placeholder="Article Date" readonly>
+          <input matInput value="{{resourceForm.value.articleDate | date}}" i18n-placeholder placeholder="Article Date" readonly>
```
nirojdyola: We have used date: 'fullDate' on most of the component. So, could we do same on everywhere ?

PR #647, opened 2018-04-14
[Open Discussion] Better configuration management

mappuji:
I know this commit not going to work, but I need some discussion regarding this.

I'm inspired by https://12factor.net where every config should be ably configured via an environment variable. It is very important because our app served packaged in Docker container.

Now it is done via replacing some placeholder with `sed` command, and worse we are `sed`-ing the minified version of the artifact. Is there is a way to have a `config.json` or whatever that we can inject via volume mount in Docker or by `sed`-ing the config file (`config.json`) rather than `sed`-ing the minified version which very risky.

Please help with your opinion @paulbert and @dogi and @lmmrssa and Thank you 🙇 

Notes:
The changes I've made will not work, because environment variable will be set in our server and not readable in client's machine.

### DIFF
--- a/docker/planet/docker-entrypoint.sh
+++ b/docker/planet/docker-entrypoint.sh
@@ -1,24 +1,3 @@
 #!/bin/sh
 
-if [ -z "$HOST_PROTOCOL" ]
-then
-  PROTOCOL="http"
-elif [ "$HOST_PROTOCOL" = "https" ]
-then
-  PROTOCOL="https"
-else
-  PROTOCOL="http"
-fi
-
-if [ "$MULTIPLE_IPS" = "true" ]
-then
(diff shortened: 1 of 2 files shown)


paulbert:
The best solution I've been able to find for applying system environment variables to Angular CLI environment is outlined here: https://medium.com/@natchiketa/angular-cli-and-os-environment-variables-4cfa3b849659

It's not too far from what we do.  This solution writes the entire production Angular CLI environment with a node script before running `ng build`.  I think we could simply `sed` the environment.prod.ts, then build (rather than the other way around) to avoid possible problems that arise from editing the compiled & minified JS output from Webpack.

mappuji:
as per discussion, we will be using the old way but will use a more explicit placeholder like `placeholder-db-host`

mappuji:
Close this PR and will create a new one.

Issue #653, opened 2018-04-16
Refactor dialogs-form.component.html


MacNew:
It seems select box of dialogs-form.component.html is incorrect. For loop  (*ngFor) should be on option tag  not on select tag. There is no formConrolName property on mat-select tag.And make other required changes only for selectbox

PR #654, opened 2018-04-16
[#653] refactor dialogs-form.component.html (connects #653)

MacNew:
Now we can generate select text field like this
const fields = [ 
  {
        'label': 'label',
        'type': 'selectbox',
        'option': [ 'option1', 'option2', 'option3' ],
        'name': 'my name',
        'placeholder': 'my placeholder',
        'required': true
}
]

### DIFF
--- a/src/app/shared/dialogs/dialogs-form.component.html
+++ b/src/app/shared/dialogs/dialogs-form.component.html
@@ -10,8 +10,8 @@
 
       <!-- selectbox -->
       <mat-form-field *ngIf="field.type === 'selectbox'" class="full-width">
-        <mat-select placeholder='{{field.placeholder}}' formControlName="{{field.name}}" required="{{field.required}}">
-          <mat-option *ngFor ="let option of field.option" [value]="option">{{option}}</mat-option>
+        <mat-select placeholder="{{field.placeholder}}" formControlName="{{field.name}}" required="{{field.required}}">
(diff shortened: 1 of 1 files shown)


src/app/shared/dialogs/dialogs-form.component.html
```diff
@@ -10,9 +10,9 @@ <h1 mat-dialog-title><span><mat-icon>create</mat-icon>{{title}}</span></h1>
 
       <!-- selectbox -->
       <mat-form-field *ngIf="field.type === 'selectbox'" class="full-width">
-        <select mat-select *ngFor="let options of field.options">
-          <option value="{{option.value}}">{{option.text}}</option>
-        </select>
+        <mat-select placeholder='{{field.placeholder}}' formControlName="{{field.name}}" required="{{field.required}}">
```
lmmrssa: why do we require "{{}}" and also i see ' and " used

src/app/shared/dialogs/dialogs-form.component.html
```diff
@@ -10,9 +10,9 @@ <h1 mat-dialog-title><span><mat-icon>create</mat-icon>{{title}}</span></h1>
...
+          <mat-option *ngFor ="let option of field.option" [value]="option">{{option}}</mat-option>
```
lmmrssa: assumption here is that value and label both will be same which might not be the case

PR #658, opened 2018-04-16
Rename databases (fixes #573)

### LINKED ISSUE #573: Database naming: communityregistrationrequests & nations
Both of these databases could store communities or nations depending on whether it is Earth or a nation.  I think it would make more sense to name them as follows:

- communityregistrationrequests -> registrationRequests
- nations -> childPlanets

YangLou1:


couchdb-setup.sh
```diff
@@ -92,8 +92,8 @@ curl -X PUT $COUCHURL/_global_changes
 curl -X PUT $COUCHURL/meetups
 curl -X PUT $COUCHURL/resources
 curl -X PUT $COUCHURL/courses
-curl -X PUT $COUCHURL/nations
-curl -X PUT $COUCHURL/communityregistrationrequests
+curl -X PUT $COUCHURL/childPlanets
+curl -X PUT $COUCHURL/registrationRequests
```
lmmrssa: could we keep the name convention and make it 'snake_case'

paulbert: Ok, let's go with `child_planets` and `registration_requests`

paulbert:
Rebasing to keep this up to date.

This is on hold because I haven't had the time to set up local center -> nation -> community for testing (which I would like to do to make sure this doesn't break any of the connections).  The changes look like they should work, though!

paulbert:
Closing.  Now that we've deployed with the other naming I don't think we should change this.

PR #662, opened 2018-04-17
Update database field name to keep consistent camelcase

lmmrssa:


paulbert:
Looks good.  I just see one field in the resources mockup: `need_optimization`.  Does that need to stay as it is or can we change it as well?

lmmrssa:
@paulbert regarding need_optimization, couldn't figure out is it added there as it will have attachment and need_optimization be part of couchdb like file_type or was it just kept there by us for some form input in old BeLL System. That's why have kept it for now.

PR #672, opened 2018-04-18
[#666] make course steps sortable

lmmrssa:


paulbert:
#666 is not about sorting the steps, but allowing the user to change the order of the steps.  No need to add the sorting package.

lmmrssa:
@paulbert when we sort and save it will update the order of the step and we are using order of step as step order right now. Other way is to put move up/down button or input field to mention position of the step.

paulbert:
Closing this in favor of #688

Issue #668, opened 2018-04-18
Taking a course


paulbert:
We need to build a component for taking a learner through the course steps.  Let's start with just showing the title & description of each step, then we can move on to adding a resources and examinations.

paulbert:
#677 is a good start for this.  Now we should add slide show like views where the user has next/previous buttons and one step is shown at a time.

PR #677, opened 2018-04-18
[#668] display step on course view

lmmrssa:


src/environments/environment.ts
```diff
@@ -6,6 +6,6 @@
 export const environment = {
   production: false,
   test: false,
-  couchAddress: 'http://127.0.0.1:2200/',
+  couchAddress: 'http://192.168.0.56:2200/',
```
lmmrssa: looks like this is the wrong file commited

paulbert: Whoops...fixing now

Issue #665, opened 2018-04-18
Attach a resource to a course step


paulbert:
The entire resource document should be stored as part of the course step

PR #681, opened 2018-04-19
[#665] Add resource to course step (connects #665)

paulbert:
Related to #665


lmmrssa:
This is the error i'm seeing when i try to attach resource.
screen shot 2018-04-19 at 17 43 52

Issue #633, opened 2018-04-12
Snackbars in UI: behaviour seems not to be in line with expected urgency of message


Tille88:
Course deleted snackbar is shown much longer than I would expect. 
This may be useful functionality for some cases, but maybe create different urgency levels of snackbars and color code/make them disappear/be animated according to urgency level could perhaps be useful. 
Having deleted course snackbar be open and maybe not even seen until way later, could confuse more than help.

<img width="251" alt="screen shot 2018-04-07 at 22 19 13" src="https://user-images.githubusercontent.com/16757941/38669697-227ced48-3e71-11e8-90d8-5d72f20edcee.png">

BasnetSamriddhi:
@Tille88  If you look into the code there are two snackbars created, one that will show up only for few seconds where as the other one has to be removed manually by the user. The one you have mentioned is the later one and once you click the cross ( X ) it will go away.

nirojdyola:
Yes that's make a sense. We need to put some duration . @BasnetSamriddhi , Please check it once .

lmmrssa:
There are already other similar issues #787 so closing it

PR #685, opened 2018-04-19
[#633] snackbar behaviour reworked(connects #633)

BasnetSamriddhi:
In this PR I have changed the duration of alert snackbar which needed manual removal and the color of the respected snacbars


BasnetSamriddhi:
I have just worked using just one change, if approved I will be making changes to other files as well.

lmmrssa:
I see the changes addresses issue upto some extend. But the main problem here is that after certain action is taken like move to next page or when next snackbar shows up, old snackbar needs to be gone. Snackbar staying there when page or context is not relevant is the main issue. @Tille88 is that correct?

BasnetSamriddhi:
@paulbert I have tried on removing the snackbar from appearing on new page. Could you please have a look at it and I have implemented this only in course section only

Tille88:
@lmmrssa @BasnetSamriddhi 
Yes, I meant something like that. It is a deeper question exactly what kind of behavior (and later on styling) that would be good for different messages. 
Having it react to a change of route would also be behavior that would be okay - the main thing was that it was weird to have it stick around for ten minutes unless closed manually. 
Making sure it gets replaced by any new snackbar would of course be required, anything else would be confusing I think...

Good first iteration. An improvement in my view, but exact behaviour I think is something that maybe you should wait for a few people's opinion. I think both could work for this particular event.

paulbert:
I'm going to close this.  It basically makes the two methods equivalent.  Instead of changing `showAlert` we should be using it consistently for errors or unexpected behavior only.

PR #692, opened 2018-04-20
styling profile image update fixes #634

### LINKED ISSUE #634: New user photo upload not resized
When creating a new user, only when picking a small photo it seems to make sense size-wise.
When uploading a large image it takes up the full screen as seen below. Rescaling to something consistent but small may make sense in this case.

screen shot 2018-04-12 at 16 55 07

ArashDai:
This branch is just to fix the profile image becoming large after uploading a new profile image. I implemented it on my 564-ProfileImageCrop branch to test it before moving it to this new branch. There seems to be some kind of issue on this branch, I was unable to upload a new profile image using the old method. Also we should make this (and probably other parts of the app) more responsive, but that might warrant its own issue. 
screenshot from 2018-04-19 17-37-01

### DIFF
--- a/src/app/users/users-update/users-update.component.html
+++ b/src/app/users/users-update/users-update.component.html
@@ -65,7 +65,7 @@
       </div>
     </form>
     <div>
-      <img class="profile-image-large" [src]="previewSrc" >
+      <img class="profile-image-update" [src]="previewSrc">
       <div>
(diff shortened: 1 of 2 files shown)


ArashDai:
@Tille88 Yea I was thinking about the cropping tool as well, I was thinking it would be nice to have the cropping tool in a hidden div, that appears when clicking an 'update profile' button?

paulbert:
@Tille88 @ArashDai Good thoughts, I think we'll make an issue about improving this layout further tonight

Issue #682, opened 2018-04-19
Limit course description length


paulbert:
![course_long_description](https://user-images.githubusercontent.com/9203229/38972594-9689a520-436f-11e8-943f-67a2e2946ec2.png)

After 3 or 4 lines should end with an ellipsis `...` (CSS can do this with `text-overflow: ellipsis;`)

lmmrssa:
max-height: 3.1em;
line-height: 1em;
overflow: hidden;
display: -webkit-box;
-webkit-line-clamp: 3;
seems to do it but it only works for -webkit browsers

paulbert:
Fixed by limiting height and scrolling on overflow for description

PR #695, opened 2018-04-20
[#682] course list description ellipsis (connects #682)

lmmrssa:
text-overflow does not seem to work as expected. Also other style works only for webkit so similar feature added by content ::after ::before

### DIFF
--- a/src/app/courses/courses.component.ts
+++ b/src/app/courses/courses.component.ts
@@ -25,6 +25,42 @@ import { Subject } from 'rxjs/Subject';
     .mat-column-action {
       max-width: 225px;
     }
+    /* styles for '...' */
+    .content-ellipsis {
+      overflow: hidden;
+      position: relative;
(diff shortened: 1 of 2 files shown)


src/app/courses/courses.component.ts
```diff
@@ -24,6 +24,42 @@ import * as constants from './constants';
...
+      content: '';
+      position: absolute;
+      right: 0;
+      width: 1em;
+      height: 1em;
+      margin-top: 0.2em;
+      background: white;
+    }
+    /* ellipsis only works for webkit 
+    .content-ellipsis {
```
paulbert: Should this still be in here?  It overwrites or duplicates some of the styles above.

paulbert:
I'm going to close this.  I think it's better to do a max characters (and try to split by words) that limit it to 3-4 lines rather than have an ellipsis that's outside the div dimensions and not aligned with the text.

Issue #694, opened 2018-04-20
Course step reorder - same step should be open


paulbert:
https://www.useloom.com/share/f6d0188e6d994b0ca6f929ee23277341

I think we need to assign each step an "id" (only needs to be unique across the course & does not need to be saved to the database) to use in the trackBy function instead of the index.

PR #700, opened 2018-04-20
[#694] course step reorder (connects #694)

MacNew:


src/app/courses/add-courses/courses-step.component.ts
```diff
@@ -13,12 +13,14 @@ import {
 export class CoursesStepComponent implements OnChanges {
 
   @Input() stepInfo: any = {
+    id: '',
     stepTitle: '',
-    description: ''
+    description: '',
```
lmmrssa: extra comma

src/app/courses/add-courses/courses-add.component.ts
```diff
@@ -126,13 +126,23 @@ export class CoursesAddComponent implements OnInit {
...
-  removeStep(pos) {
-    this.steps.splice(pos, 1);
+  uniqueIdOfStep() {
+     // Math.random should be unique because of its seeding algorithm.
+     // Convert it to base 36 (numbers + letters), and grab the first 9 characters
+    // after the decimal.
+    return '_' + Math.random().toString(36).substr(2, 9);
+  }
+
+  removeStep(step) {
```
lmmrssa: why is this removestep required need to change?

src/app/courses/add-courses/courses-add.component.html
```diff
@@ -60,10 +60,11 @@
     <div>
       <planet-courses-step
         *ngFor="let step of steps; index as i; trackBy: stepTrackByFn"
-        (stepRemove)="removeStep(i)"
+        (stepRemove)="removeStep(step)"
         (stepOrder)="orderStep(i, $event)"
         [(stepInfo)]="steps[i]"
         [stepCount]="steps.length"
+        [stepId]= "step.id"
```
lmmrssa: why this input field is required. It is not used anywhere

src/app/courses/add-courses/courses-add.component.ts
lmmrssa: it seems like you already added filter which is breaking it instead of removing item from array

lmmrssa:
This PR is not solving the issue. It is just adding ID but not fixing the expansion being open.

Issue #697, opened 2018-04-20
warning on style compile


lmmrssa:
![screen shot 2018-04-20 at 14 38 06](https://user-images.githubusercontent.com/2909490/39041495-968fbba6-44a8-11e8-8387-935f5a6c62a7.png)

PR #719, opened 2018-04-23
[#697] scss compiling warning

lmmrssa:


paulbert:
I'd prefer to find a solution which doesn't add duplication across our CSS and removes the warning.  Maybe  using a mixin like the last answer to [this SO post](https://stackoverflow.com/questions/48588493/angular-5-and-css-grid-cannot-find-grid-areas-warning) is the closest we can come to that.

Issue #730, opened 2018-04-24
Table sorting is case sensitive


paulbert:
Tables sorting should be case insensitive.

PR #737, opened 2018-04-24
[#730] Table sort8ing is case sensitive (connects #730)

MacNew:


nirojdyola:
sortingDataAccessor is repeatedly using in all compoent, so we can reduce it. We"ll create new issue for this once it is merge.

Issue #726, opened 2018-04-23
Code duplication: MatTables


paulbert:
Right now we have a lot of duplication across our components with MatTables.

We should refactor to reduce these components to reduce this repetition as much as possible (follow [T-DRY](https://angular.io/guide/styleguide#t-dry-try-to-be-dry)

Right now the components with tables are 4 of the top 5 worst files according to CodeClimate:

![image](https://user-images.githubusercontent.com/9203229/39157002-b50ae788-4726-11e8-962b-1dd15e845c0d.png)

paulbert:
Feedback table also has some duplication now (see #1569)

lmmrssa:
@paulbert shall we go with inheritance of classes then? So we have all common functions on parent class and child will overwrite based on their requirement.

PR #741, opened 2018-04-24
[#726] duplicate code removed (connects #726)

BasnetSamriddhi:


src/app/courses/courses.component.html
```diff
@@ -1,5 +1,5 @@
 <mat-toolbar>
-  <button mat-icon-button (click)="goBack()"><mat-icon>arrow_back</mat-icon></button>
+  <button mat-icon-button (click)="planetMatTableService.goBack(parent)"><mat-icon>arrow_back</mat-icon></button>
```
lmmrssa: goback is not a part of mattable. It is used across all application to route back to relevant upper hierarchy page.

lmmrssa:
This code is completely failing because of all the code that is not properly introduced.

paulbert:
This shifts the duplication from the components to the templates with `planetMessageService` all over.

Also, we should not be directly accessing services in the templates anyway.

There's a lot to redo, I think it would be better to start fresh.

PR #762, opened 2018-04-26
[#761] Remove Meetups from Dashboard (connects #761)

nirojdyola:
design needs to be changed. Raju is working on design . And once this has been approved will make a new issue for removing the shelf from dashboard.


paulbert:
This creates another problem by forcing the user to click on the text rather than anywhere on the tile to go to that resource/meetup/etc.  Also only creates the X for the first tile.

Issue #765, opened 2018-04-26
Add links in empty dashboard areas


paulbert:
When, for example, myLibrary is empty fill the dashboard tile with a link to resources.

PR #778, opened 2018-04-27
[#765] Empty card message with link (connects #765)

lmmrssa:


lmmrssa:
screen shot 2018-04-27 at 07 27 23

Issue #777, opened 2018-04-26
what is the local url there for


leomaxi:
Do we need to change it? If not, why do we show it at all? 
<img width="712" alt="screen shot 2018-04-26 at 5 43 10 pm" src="https://user-images.githubusercontent.com/2836302/39322471-8266b8a2-4979-11e8-953d-a77b3528fffd.png">

paulbert:
I'm going to close this because I think it's covered by the change of #780.  We still may take out localUrl eventually, but for now it's in a good place

PR #780, opened 2018-04-27
[#777] Remove local url input field (connects #777)

Rupesh87:


paulbert:
I looked into hiding this field within an [expansion panel](https://material.angular.io/components/expansion/overview), but the panel is not showing for some reason.  We'd like to keep the option to change this, but hide it by default.

Issue #758, opened 2018-04-26
Shelf update function in userService


paulbert:
Now that we are storing the shelf in the userService, we can simplify our code further by adding a function for updating the shelf which takes an array of ids, the name of the shelf property to be updated, and a string with 'add' or 'remove' depending on which we are doing.

Our user shelfChange Subject should send the next value of `shelf` so any components listening just use that value rather than accessing the getUserShelf method.

paulbert:
Fixed #1368

PR #781, opened 2018-04-27
[#758] Shelf update function in UserService (connects #758) 

MacNew:


src/app/resources/resources.component.html
```diff
@@ -23,8 +23,8 @@
 <div class="space-container primary-link-hover">
   <mat-toolbar class="primary-color font-size-1">
     <span class="toolbar-fill"></span>
-    <span *ngIf="!parent">
-      <button mat-button [disabled]="!selection.selected.length" (click)="addToLibrary(selection.selected)">
+    <span *ngIf="!parent"><!-- (click)="addToLibrary(selection.selected)" -->
```
razu9861: if this click function is not needed than you can remove it

Issue #775, opened 2018-04-26
Selected resources indication (number of selected, which ones added)


Tille88:
**Tested on 0.1.6**

I would like to have an indication of number of selected/checked resources by the "Add to library"-button

<img width="165" alt="screen shot 2018-04-26 at 22 06 13" src="https://user-images.githubusercontent.com/16757941/39314680-0527eaca-499f-11e8-8e01-168cc7d395d7.png">

Alternatively, or in combination: have the added snackbar say something like:
“{{ ResourceName }} added to your library" Instead of “Resource added to your library”

<img width="370" alt="screen shot 2018-04-26 at 22 13 29" src="https://user-images.githubusercontent.com/16757941/39314703-118f4146-499f-11e8-8597-4a7fb69a9861.png">

lmmrssa:
There is #776 specifically for already added indication. Number of selected item is already done.

PR #783, opened 2018-04-27
[#775] Selected resources indication (connects #775)

Rupesh87:


Tille88:
@Rupesh87 
Okay, it works for me. Having the number of resources added was a nice touch. Brings me back to the original issue a bit though. If multiselect over several pages is possible, it would be nice to have something like this (but nicer)
screen shot 2018-04-27 at 19 43 20

In other words, have next to the button having a counter updating when the selection changes.

screen shot 2018-04-27 at 19 48 02

I can also click the checkbox and add the same resource any number of times as I said in the other issue. The dropdown updates to remove resource, but the checkbox add still is active:

Before (added already as seen by snackbar):
screen shot 2018-04-27 at 19 52 17

After:
screen shot 2018-04-27 at 19 52 23

Issue #766, opened 2018-04-26
Added items ends up as last item in list views


Tille88:
**Tested on 0.1.6**

Resources/library section. If I add resources, they end up at the bottom. Probably should be top. If there are several pages of content, you would have to go to the last page to see your added resource.

Also needed for some other list views

PR #800, opened 2018-04-30
[#766] Fix Added items ends up as last item in list views (connects #766)

lucy811:


paulbert:
For courses and meetups we should also add a `createdDate` field.

Issue #812, opened 2018-05-01
Error changing password message when change is successful for admin


paulbert:
Password has been changed to new one, but the error message pops up and there are the below errors in the console.  Only happened for admin user.
![image](https://user-images.githubusercontent.com/9203229/39460581-7e0aeeca-4cd2-11e8-8f46-968aa8236053.png)

PR #821, opened 2018-05-02
[#812] Error changing password message when change is successful (connects #812 )

Rupesh87:


src/app/users/users-profile/users-profile.component.ts
```diff
@@ -85,13 +85,13 @@ export class UsersProfileComponent implements OnInit {
       this.couchService.put(this.dbName + '/' + userData._id, userData)
     ];
     if (isUserAdmin) {
-      observables.push(this.couchService.get('_users/' + userData._id , { domain: this.userService.getConfig().parentDomain })
+     /*  observables.push(this.couchService.get('_users/' + userData._id , { domain: this.userService.getConfig().parentDomain })
```
lmmrssa: this is for updating admin password on parent planet. we will have to check if parent planet exist before executing this one to prevent failure

Issue #809, opened 2018-05-01
Courses add template media queries


paulbert:
Add a few media queries for screen size to reduce the first column of the main grid template.

Set breaks at 1000px and 700px, reducing the first column to 365px and 180px, respectively.

PR #823, opened 2018-05-02
[#809] Courses add template media queries (connects #809)

MacNew:


lmmrssa:
Removed duplication of code as we just need to change one line of style per screen size. Also adjusted the width as it was looking too small for 700px.
We still have issue of step overflow as seen in picture below but it will probably be solved by #822 
screen shot 2018-05-02 at 15 14 41

src/app/courses/add-courses/courses-add.scss
```diff
@@ -13,3 +13,39 @@
     align-content: flex-start;
   }
 }
+
+@media only screen and (max-width: 1000px) and (min-width: 700px) {
```
lmmrssa: all style is duplicated we just need to adjust grid column width

paulbert:
Fixed up the widths so the breakpoint to 1 column is just as the step begins to overflow.  Also changed the width for the 1 column because that 1 extra pixel made the buttons display side by side rather than one on top of the other.

Also added scss variables for the breakpoints so we can be consistent with our other views if needed.  Note that both min-width and max-width are not required (which makes it much easier to use variables!).

Issue #832, opened 2018-05-03
Fix course mockup


lmmrssa:
Currently course mockup does not have memberLimit and also steps. We need to update course mockup to represent our basic course structure.
![screen shot 2018-05-03 at 09 19 26](https://user-images.githubusercontent.com/2909490/39559367-65a7104a-4eb5-11e8-86d5-ca7c68b2c86d.png)

PR #833, opened 2018-05-03
[#832] Update course mock to basic setup (connects #832)

lmmrssa:


paulbert:
I think we should keep course-validators even though we are not using it since it will be useful for meetups

Issue #840, opened 2018-05-03
config special feature should not trigger when on but nothing changed


dogi:
![a-20180503-101114](https://user-images.githubusercontent.com/13684059/39581947-bbc9bd10-4eba-11e8-8cb0-0d38751e744d.png)
special feature on but nothing change

![a-20180503-101204](https://user-images.githubusercontent.com/13684059/39581953-bec38c8a-4eba-11e8-8a40-824594fbec46.png)
there should be no special question

PR #848, opened 2018-05-03
[#840] Fixing single variable condition (connects #840)

lmmrssa:
This PR uses single variable to fix the error by updating condition to check true instead of false.


paulbert:
I'm going to close this in favor of #849

PR #849, opened 2018-05-03
[#840] Introduce new variable for configuration confirm (connects #840)

lmmrssa:
This PR introduce new variable to check confirmation checked.


paulbert:
I like having separate variables for the state of the input, toggle, & checkbox.  Implemented like this (with the confirmation showing no matter based on the input value) I think we can actually get rid of `showConfirmAdvancedOptions`.

Issue #841, opened 2018-05-03
Admin configuration should not be available once done


lmmrssa:
Even when we have created admin and configuration. We can go to the configuration by moving to link `/login/configuration` This should not happen.

PR #854, opened 2018-05-04
[#841] Admin configuration should not be available once done (connects #841)

Rupesh87:


paulbert:
The previous implementation causes breaking errors with the configuration component.  The way to stop a component from loading is with a [route guard](https://angular.io/guide/router#milestone-5-route-guards).  When `ngOnInit` runs the component is ready to render and expecting template variables to be available on the component.

Issue #837, opened 2018-05-03
Deleting a community with one click


paulbert:
Right now we have to

- Delete from "Community List"
- Delete from "Requests List"
- Manually delete associated user (see #836)

One click on the Community List should do all of these tasks.

PR #855, opened 2018-05-04
[#837] Delete request, parent and user (connects #837)

lmmrssa:


paulbert:
Ended up with nearly identical code for `NationComponent` and `CommunityComponent`, but we should decide if we need both before creating a service to reduce the duplication.

Issue #691, opened 2018-04-19
Add google analytics (or other) tracking code


sente:

Rupesh87:
@paul @sente  to use google analytics our planet app must run in production environment. So once it is hosted we can implement it. am i right

sente:
This is not true.  Google analytics and other trackers don't care if you're in production or not.  Perhaps I a not understanding what you mean exactly?

If for some reason the tracker endpoints are unreachable then they're designed to fail gracefully, etc.

paulbert:
@Rupesh87 We can use Google Analytics from any environment.  The issue is that we would not want to send data into our production anaylitics from a development environment, but there are ways to fix that.

@sente I didn't consider this earlier because we need to collect data while offline, but I'm seeing a simple solution for that here: https://developers.google.com/web/tools/workbox/guides/enable-offline-analytics.  Any thoughts on Workbox or using GA with a service worker?

Also something to consider: We'll need an easy way for setting up tracking for different communities and nations.  Ideally we could add a new GA account when installing the community.

Generally I like the idea since it's much easier to add GA than to implement our own analytics tracking.

PR #860, opened 2018-05-05
[#691] Google Analytics added for page view tracking

paulbert:


lmmrssa:
@paulbert as we will have different deployments with same url, can we also save deployment detail

paulbert:
@lmmrssa There are ways to do that, but this is more a first step to get minimal tracking to develop off of.

Issue #808, opened 2018-05-01
Add markdown support


paulbert:

lmmrssa:
List of some MDE
* https://github.com/dimpu/ngx-md
* https://github.com/sparksuite/simplemde-markdown-editor

paulbert:
One issue with ngx-md: They have an accepted PR which maybe creates an XSS vulnerability.  I commented on their PR, so we should wait to see what their response is.

XSS is a good thing to be aware of as we look at these editors as we should make sure we do not allow malicious scripts to run.

PR #862, opened 2018-05-06
[#808] Add markdown support (connect #808)

MacNew:
screenshot-2018-5-6 planet learning

### DIFF
--- a/.angular-cli.json
+++ b/.angular-cli.json
@@ -17,10 +17,14 @@
       "prefix": "app",
       "styles": [
         "../node_modules/material-design-icons/iconfont/material-icons.css",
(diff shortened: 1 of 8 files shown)


MacNew:
screenshot-2018-5-7 planet learning
screenshot-2018-5-7 planet learning 1

lmmrssa:
@MacNew why this PR has multiple markdown code included? Can we have separate implementation of each markdown editor to see each one separately

paulbert:
@MacNew Ok, I'm going to close this, then.  We should look for a solution that uses one library.  If we can't find one, then we can reopen this.

Issue #834, opened 2018-05-03
Deleting items table errors


paulbert:
After deleting items from a table once all subsequent deletion attempts fail.  Looking at the requests, the app is trying to delete already deleted docs.

Found this on the courses table.  Did not test with others.

Rupesh87:
@paul i haven't got any error while checking in courses table

PR #864, opened 2018-05-07
[#834] Delecting items from course fix (connects #834)

Rupesh87:


lmmrssa:
@paulbert Is this issue about error during deletion of course as it is sending already deleted course id. Problem I see here is selection is not cleared which is happening even on admission. So have implemented selection clear after all dialog close irrespective of success or failure.

Adding switchmap does not seem to make difference @paulbert please confirm does switchmap makes any difference in this code?

Issue #878, opened 2018-05-08
_users design doc: Do not permit deleting admin user


paulbert:

lmmrssa:
this issue is not about hiding button

PR #890, opened 2018-05-08
[#878] Do not permit deleting admin user (connects #878)

Rupesh87:


Rupesh87:
@lmmrssa i can't update _users design/auth so i created new design/prevent_admin in my couchdb. Is there any other way to fix this issue

```
function(newDoc, oldDoc, userCtx, secObj){ 
    if(oldDoc._deleted === true && oldDoc.isUserAdmin === true)
      throw({'forbidden': 'admin cannot be deleted'});
    }
} 
```

Issue #867, opened 2018-05-07
Delete user's shelf with user


paulbert:
We also need to delete the shelf of the user with the user document

PR #897, opened 2018-05-09
[#867] Delete user's shelf with user (connect #867)

MacNew:


lmmrssa:
@paulbert have made some changes to code. Still it not be optimized so please review once.

src/app/users/users.component.ts
```diff
@@ -157,27 +157,17 @@ export class UsersComponent implements OnInit, AfterViewInit {
     const myUserId = '_users/org.couchdb.user:' + this.userService.get().name;
     this.couchService.allDocs('shelf').subscribe(shelfUser => {
       shelfUser.map(shelf => {
-        const myTeamIds = [ ...shelf.myTeamIds ] || [];
-        if(myTeamIds.indexOf(idShouldBeRemove)) {
+        if ('myTeamIds' in shelf) {
+          const myTeamIds = [ ...shelf.myTeamIds ] || [];
```
lmmrssa: this code is supposed to set empty array if there is no myTeamIds so if condition should not be needed

src/app/users/users.component.ts
```diff
@@ -157,27 +157,17 @@ export class UsersComponent implements OnInit, AfterViewInit {
     const myUserId = '_users/org.couchdb.user:' + this.userService.get().name;
     this.couchService.allDocs('shelf').subscribe(shelfUser => {
       shelfUser.map(shelf => {
-        const myTeamIds = [ ...shelf.myTeamIds ] || [];
-        if(myTeamIds.indexOf(idShouldBeRemove)) {
```
lmmrssa: You need to check if Id exists before you update else it will update shelf of all users increasing request to couchdb

PR #899, opened 2018-05-09
Replication (connects #901)

lmmrssa:


lmmrssa:
Probably we can close this if we are going with #1333

Issue #895, opened 2018-05-08
Feedback scrolls to oldest message when replying


paulbert:
Newest messages are on the bottom of the list, but the feedback scrolls to the top.

PR #900, opened 2018-05-09
[#895] Feedback scrolls to oldest message when replying (connect #895)

MacNew:


paulbert:
Looking at this more, I'm only seeing the issue in Chrome.  Adding an interval Observable to repeat `scrollToBottom` still results in the list scrolling to the top for a split second before scrolling to the bottom.

I'm not sure what is causing the scrolling to be off in Chrome, but since this at least adds a nice transition for scrolling we can merge it.  I'll make a new issue with what I learned about this.

Issue #920, opened 2018-05-10
Do not allow special characters


lmmrssa:
For username and planet code just allow `.` `-` `_` as special character.

PR #922, opened 2018-05-10
[#920] Do not allow special characters (connect #920)

lucy811:


src/app/configuration/configuration.component.ts
```diff
@@ -45,7 +45,7 @@ export class ConfigurationComponent implements OnInit {
 
   ngOnInit() {
     this.loginForm = this.formBuilder.group({
-      name: [ '', Validators.required ],
+      name: [ '', [ Validators.required, Validators.pattern(/^[a-z0-9]+$/i) ] ],
```
lmmrssa: we also want to allow `.` `-` and `_`

src/app/login/login-form.component.ts
```diff
@@ -22,7 +22,7 @@ const repeatPassword = {
 };
 
 const loginForm = {
-  name: [ '', [ Validators.required, Validators.pattern(/^[a-z0-9]+$/i) ] ],
+  name: [ '', [ Validators.required, Validators.pattern(/^[a-z0-9_.-]+$/i) ] ],
```
lmmrssa: We will allow even `@` on login on parent. And we will just need validation for registration. We can skip validator on login.

Issue #519, opened 2018-03-16
User name unique across nation/community/center


paulbert:
Leading idea is to create user name as {nation name}:{community name}?:{user name}

lmmrssa:
@paulbert _auth design doc of _users has 
`var badUserNameChars = [':']`
which will make ':' invalid on username

lmmrssa:
As per last discussion @paulbert @dogi we will add `username@planet_code` to be saved on parent planet as username so that username can only be validated within planet not with parent.

PR #927, opened 2018-05-11
[#519] Planet code on parent with username (connects #519)

lmmrssa:


lmmrssa:
This will also require #926 which will remove special characters from planet code.

Issue #846, opened 2018-05-03
Take exam from course step view


paulbert:
Requires new view for taking an exam and routing from the course step view

paulbert:
Fixed in #928

PR #928, opened 2018-05-11
[#846] Taking an exam from course step view

paulbert:
Related to #846


dogi:
@paulbert I did not test it but don't wanted you to get stock

Issue #923, opened 2018-05-11
meetup view selection not working 


Rupesh87:
version: 0.1.12

![meetup](https://user-images.githubusercontent.com/18145146/39920400-aa1dcf2a-5536-11e8-9418-43030f341270.png)

lmmrssa:
Issue is not clearly described. In master I could see select all being unselected when doing any action but selection of all items remain. If issue is to remove selection then there is issue #869 which contradict with it.

paulbert:
Fixed in #924

PR #932, opened 2018-05-14
[#923] meetup view selection not working (connects #923) 

MacNew:


lmmrssa:
Please refer to #939 adding just a line of code on update listener will do it.

paulbert:
Fixed in #924

Issue #931, opened 2018-05-13
Courses options not in Course details view


Tille88:
**Tested on 0.1.12**

This whole section is probably semi-finished at best (taking a course I assume will look differently), but from the details view on a course I would at least be able to have the enroll/admission option that is in the list view. Having to click back to enroll if interested in a course is a bit useless :)

![screen shot 2018-05-13 at 8 37 24 pm](https://user-images.githubusercontent.com/16757941/39967876-99603a20-56ed-11e8-9192-791d1fc09c9f.png)

These is how the details view looks:

![screen shot 2018-05-13 at 8 40 39 pm](https://user-images.githubusercontent.com/16757941/39967900-f542a68e-56ed-11e8-8cde-0f2e3d6c0c3e.png)

PR #933, opened 2018-05-14
[#931] Courses options not in Course details view (#Connect 931)

MacNew:


src/app/courses/courses.service.ts
```diff
@@ -1,19 +1,29 @@
...
 import { Subject } from 'rxjs/Subject';
+import { PlanetMessageService } from '../shared/planet-message.service';
+import { UserService } from '../shared/user.service';
 
 // Service for updating and storing active course for single course views.
 @Injectable()
 export class CoursesService {
 
   course: any = { _id: '' };
+  private shelfUpdate = new Subject<any[]>();
```
lmmrssa: there is already observable for shelfupdate on user service we can just listen to that service. So it is not required to create new listener

Issue #869, opened 2018-05-07
Update shelf in table without clearing selection


paulbert:
It's not a great user experience to reset the selected items in a table after every action.

PR #939, opened 2018-05-14
[#923] Meetup clear selection on update (connects #923)

lmmrssa:


paulbert:
This would conflict with #869, and the issue has been fixed with #924

Issue #935, opened 2018-05-14
Repeating meetup days


lmmrssa:
If meetup is repeating then shouldn't days of repeat be mentioned. Such as repeat on certain days
![screen shot 2018-05-14 at 18 43 46](https://user-images.githubusercontent.com/2909490/39998896-39934334-57a7-11e8-8ab5-d839fed18085.png)

PR #943, opened 2018-05-15
[#935] Repeating meetup days (connects #935)

Rupesh87:


src/app/meetups/add-meetups/meetups-add.component.ts
```diff
@@ -120,4 +136,25 @@ export class MeetupsAddComponent implements OnInit {
...
+      }
+    } else {
+      // remove from day array if unchecked
+      const index = this.meetupFrequency.findIndex(x => x.value === day);
+      this.meetupFrequency.splice(index);
+    }
+  }
+
+  toggleDaily(val) {
+    val ? this.showDaysCheckBox = val : this.showDaysCheckBox = val;
```
paulbert: @Rupesh87 A couple of things I want to point out here:

1. Please do not use the ternary operator to run different code blocks.  It's easier to read our code if we only use it to assign a variable different values based on a condition.
2. In addition, you also have the same exact code on either side of the ternary operator.  There is no need for a conditional statement here at all.

Rupesh87: thanks @paulbert

Issue #936, opened 2018-05-14
List of languages


lmmrssa:
On profile language, we should add list of languages to select. Also purpose of language is not clear on profile.

- Preferred language of system
- Native or spoken of user
- Language for resources or course to give preference

![screen shot 2018-05-14 at 18 51 15](https://user-images.githubusercontent.com/2909490/39999215-3f34bc72-57a8-11e8-866f-64835b923f37.png)

lmmrssa:
@dogi @paulbert #971 is assumes the language in profile is preferable language of system. Are we going to use it for that purpose or do we use it as native language of user.

PR #971, opened 2018-05-16
[#936] List of Languages (connects #936)

MacNew:


lmmrssa:
Solution presented here is assumes that language will be preferred language of system. If we intend to use this field as user spoken language then we need to generate separate language list.

paulbert:
Since this is a language selected for the individual user, I'm presenting this as the user spoken language (which would be their preferred language of system, I suppose).  In any case, we don't need to complicate this since we're not sure what it will ultimately be used for.

Issue #972, opened 2018-05-16
update meetup created date field shows in timestamp format


Rupesh87:
![timstamp](https://user-images.githubusercontent.com/18145146/40105550-fc2e4972-5912-11e8-97a1-25ec300c9a38.png)

paulbert:
Closing this because we hid that field in #945

PR #974, opened 2018-05-16
[#972] update meetup created date field shows in timestamp format (connect #972)

MacNew:
@lmmrssa @paulbert @Rupesh87 Time stamp appears on created date only if a user creates a new meet up. So I use a moment library to convert time stamp to planet date format


paulbert:
Closing this because we hid that field

Issue #959, opened 2018-05-16
Do not allow special characters to start username or planet name code


lmmrssa:
This issue can be linked along with #958
![screen shot 2018-05-16 at 09 40 43](https://user-images.githubusercontent.com/2909490/40095742-533312aa-58ed-11e8-800c-eac88977c5c5.png)
![screen shot 2018-05-16 at 09 41 04](https://user-images.githubusercontent.com/2909490/40095743-536efe46-58ed-11e8-9699-e3aaca34c9b1.png)

PR #979, opened 2018-05-16
[#959] Do not allow special characters to start username or planet name code (connects #959)

lucy811:


src/app/configuration/configuration.component.ts
```diff
@@ -66,7 +66,8 @@ export class ConfigurationComponent implements OnInit {
       localDomain: this.defaultLocal,
       name: [
         '',
-        Validators.required,
+        [ Validators.required,
```
lmmrssa: planet name does not have any character restriction right now except it should not start with special characters

Issue #977, opened 2018-05-16
Rating comment gets old comment


lmmrssa:
When you click rating star, comment from last form remains on form.
![screen shot 2018-05-16 at 19 04 34](https://user-images.githubusercontent.com/2909490/40119662-f34289ca-593c-11e8-94f2-2c45d0f99121.png)

paulbert:
I'm seeing this only happen on ResourcesView component, and when I click on the rating the entire view changes to another resource.

Didn't we realize that this was an issue with our mock data?

PR #983, opened 2018-05-17
[#977] Rating comment gets old comment (connects #977)

MacNew:


src/app/resources/rating-resources/resources-rating.component.ts
```diff
@@ -42,7 +42,7 @@ export class ResourcesRatingComponent implements OnChanges {
     return { rate: this.rating.userRating.rate || 0 };
   }
   get commentField() {
-    return { comment: this.rating.userRating.comment || '' };
+    return { comment: '' };
```
lmmrssa: this code is even removing old comment of user for that resource. Issue is not to have comment filled for new rating

paulbert:
This would only create another problem, so I'm going to close this.

Issue #948, opened 2018-05-15
FeedbackView: Chrome issue


paulbert:
When adding a new reply in Chrome the list jumps to the top before scrolling to the bottom.

paulbert:
https://www.useloom.com/share/dd26ca5565ae467ba85405dc0d8eb170

PR #984, opened 2018-05-17
[#948] FeedbackView: Chrome issue (connect #948)

MacNew:


lmmrssa:
The changes is creating issue. Also the changes made is reverting it back to old code
screen shot 2018-05-17 at 18 45 25

Issue #998, opened 2018-05-17
signaling of status and "parent" availability


dogi:

lmmrssa:
We need to display different status of parent and also color code
* Not configured
* Pending Acceptance
* Accepted
* Accepted but not reachable

PR #1008, opened 2018-05-18
[#998] Parent online status (connects #998)

lmmrssa:


Rupesh87:
@lmmrssa online status is showing only on login page. But in BellApps online status is shown at dashboard

Issue #989, opened 2018-05-17
Add buttons on course view.


lmmrssa:
We can probably add take test or view resource on course view itself. Also admission button needs to be added on course.
![screen shot 2018-05-17 at 19 19 29](https://user-images.githubusercontent.com/2909490/40180755-4f2e5c08-5a07-11e8-9ce4-e785bf2981ef.png)

Rupesh87:
@lmmrssa for admission/resign button there is already issue https://github.com/open-learning-exchange/planet/issues/931

PR #1009, opened 2018-05-18
[#989] Added take test and view resource buttons on course step view (connects #989)

Rupesh87:


src/app/exams/exams-view.component.ts
```diff
@@ -51,7 +53,7 @@ export class ExamsViewComponent implements OnInit, OnDestroy {
   }
 
   goBack() {
-    this.router.navigate([ '../../' ], { relativeTo: this.route });
+    this.location.back();
```
lmmrssa: this code can be reverted so exam will take back to step

Rupesh87: Ok i will revert the mention changes but if i click take test from course view then it will send back to step view

PR #1010, opened 2018-05-18
[#997] Show parent version on manager dashboard (connects #997)

lmmrssa:


lmmrssa:
@paulbert do you want to have new db just for version? or update version on configuration table on each db-init?
We will have to sync configuration between parent child planets anyway.

paulbert:
I think I've got it: we can have a document on the `configuration` database with the id `version` that holds this.  That way we just have a simple upsert from the script and we don't need to get the full configuration document, update the version in it, and then update the document on the database.

We'll need to update the configuration guard to check for > 1 docs on that DB rather than > 0.

paulbert:
I'm going to merge this so we have it added, but knowing that we may revert some of the changes if we make changes to directly access version number on package.json for the parent.

Issue #1006, opened 2018-05-17
All child planets in communityregistrationrequests


paulbert:
We do not need to duplicate data across two databases, so keeping all child planets in communityregistrationrequests will work.

Other changes needed:
- Only show accepted in `NationComponent`
- For configuration will need to pull accepted nation from communityregistrationrequests

Maybe will need more changes

PR #1012, opened 2018-05-18
[#1006] Do not use nations database (connects #1006)

lmmrssa:


lmmrssa:
We have another PR to change DB name #658 We can either update that PR to remove nations DB as well or can add that code over there along with change of DB name.
@paulbert Please give your input

Issue #1013, opened 2018-05-18
Change add course text


Rupesh87:
In add exam toolbar section it is showing Add Course instead of Add Exam

![exam](https://user-images.githubusercontent.com/18145146/40230738-6baf9692-5ab8-11e8-98f1-84e8566fc810.png)

PR #1017, opened 2018-05-18
[#1013] Change add course text (connect #1013)

MacNew:


src/app/exams/exams-add.component.html
```diff
@@ -4,7 +4,7 @@
 
 <div class="space-container">
   <mat-toolbar class="primary-color font-size-1">
-    <span i18n>Add Course</span>
+    <span i18n>Add Exam</span>
```
lmmrssa: It need to be Add or Update of exam

Issue #991, opened 2018-05-17
Phone number format


lmmrssa:
Phone format can be just numbers and optional + at first for area code

PR #1019, opened 2018-05-18
[#991] Phone number format fix (connects #991)

Rupesh87:


Rupesh87:
@lmmrssa as u mention in comment i have updated validator which allow + optional for area code and numbers only. Please review once.

lmmrssa:
If we want other characters to be allowed as well then we might have to update this PR. Phonic symbols include `+ - / ()  #`

lmmrssa:
closing this pr as we are not much concern about all format of phone number right now

Issue #1023, opened 2018-05-18
Take exam button text: continue if in progress


paulbert:
In CoursesStepView.  If examStart > 1, then change "Take Exam" to "Continue Exam".

PR #1024, opened 2018-05-19
Take exam to Continue Exam if not first step (connects #1023)

Tille88:
Seemed like a small enough change to keep the logic in the template

### DIFF
--- a/src/app/courses/step-view-courses/courses-step-view.component.html
+++ b/src/app/courses/step-view-courses/courses-step-view.component.html
@@ -14,7 +14,7 @@
       *ngIf="stepDetail.exam"
       (click)="goToExam()"
       i18n>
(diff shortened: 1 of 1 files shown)


src/app/courses/step-view-courses/courses-step-view.component.html
```diff
@@ -14,7 +14,7 @@
       *ngIf="stepDetail.exam"
       (click)="goToExam()"
       i18n>
-      Take exam
+      {{stepNum === 1 ? "Take Exam" : "Continue Exam"}}
```
paulbert: Should be `examStart` rather than `stepNum`

Tille88:
@paulbert that wasn't super clear to me that each step would be considered a separate exam. 

I think there are a few more issues to raise here now when I understand a bit more how it's supposed to work.

PR #1025, opened 2018-05-19
[WIP] Integrate PouchDB

LKhadka:
What works?

- Authentication (Signing up normal users and logging in, used in login-form and auth-guard)
- Fetching and viewing courses from the Courses Database through the local PouchDB instance

If you want to test it please do the following:

- Create a database called "test" in CouchDB which is needed for authentication. It is a temporary placeholder for the roles (learner, intern) databases.
- Test it while logged in as the administration. Normal users will not work because they don't have any roles set and CouchDB has rules that allows only specific users to query those databases.

Notes:
1. I am storing the localdb and the authdb in separate dedicated variables but the other databases in an object. It will make it easier to replicate changes.
2. I have created an index for the field "kind" but some of the documents do not have that field so I think we should update them to include that. Indexing makes the lookup faster. Right now, all the courses are in the local PouchDB but you can't see them all in the app because some of them do not have the field "kind".
3. Similarly, some courses documents do not have the field "createdAt" which is used for sorting. This needs to be fixed and sorting should be fixed.


LKhadka:
Sorry about that @paulbert. VSCode automatically made those changes. I will keep that in mind.

LKhadka:
Closing for new PR

Issue #1003, opened 2018-05-17
Course update limitation


lmmrssa:
We should make course manageable by either course leader or by admin only. Or may be course creator.
Should see manage buttons on course view as well to make it more prominent.

paulbert:
Admins & course creator.  Requires creating a `creator` property on the course.

PR #1026, opened 2018-05-21
[#1003] Allow course manage by either course creator or by admin only (connects #1003)

Rupesh87:


src/app/courses/add-courses/courses-add.component.ts
```diff
@@ -69,7 +71,8 @@ export class CoursesAddComponent implements OnInit {
       method: '',
       gradeLevel: this.gradeLevels[0],
       subjectLevel: this.subjectLevels[0],
-      createdDate: Date.now()
+      createdDate: Date.now(),
+      creator: this.userService.get().name
```
lmmrssa: once we have course shared between planets it will not be valid as same username can be taken by another user on different planet

src/app/courses/courses.component.html
```diff
@@ -68,7 +68,7 @@ <h3 class="header"><a [routerLink]="['view', element._id]">{{element.courseTitle
               <mat-icon>clear</mat-icon>
               <span i18n>Resign</span>
             </a>
-            <a mat-menu-item (click)="updateCourse(element)">
+            <a mat-menu-item *ngIf="userService.get().isUserAdmin || element.creator === userService.get().name" (click)="updateCourse(element)">
```
lmmrssa: even delete can use this condition

lmmrssa:
Added condition to delete and for bulk action added filter to not delete unauthorized courses.
Made creator to use planet code to make it unique across all planet

Issue #990, opened 2018-05-17
Show test only if in mycourse


lmmrssa:
Take test should be available only when member had added to myCourse.
Otherwise we might show add to myCourse or some message which gives information they need to add before taking test.
Or when user clicks take test we can ask for confirmation and add it to mycourse.

PR #1027, opened 2018-05-21
[#990] Show test only if in mycourse (connect #990)

MacNew:
Once  this PR https://github.com/open-learning-exchange/planet/pull/942  is merged it works fine


src/app/courses/step-view-courses/courses-step-view.component.ts
```diff
@@ -67,7 +71,14 @@ export class CoursesStepViewComponent implements OnInit, OnDestroy {
   }
 
   goToExam() {
```
paulbert: It would be much simpler to hide the button if the user has not added the exam

Issue #1043, opened 2018-05-22
show message while viewing parent resources, meetups, courses


Rupesh87:
If parent planet is not connected then while viewing parent resources in manager view we should show parent not available
![parentmessage](https://user-images.githubusercontent.com/18145146/40341728-7f1bdb5a-5da6-11e8-8c31-ac0522860915.png)

PR #1059, opened 2018-05-22
[#1043] Show message while viewing parent resources, meetups, courses (connects #1043)

MacNew:


src/app/manager-dashboard/manager-dashboard.component.ts
```diff
@@ -84,4 +91,18 @@ export class ManagerDashboardComponent implements OnInit {
       }, error => this.planetMessageService.showAlert('An error occurred please try again.'));
   }
 
+  checkRequestAcceptedOrNot() {
+    this.couchService.allDocs('communityregistrationrequests', { domain: this.userService.getConfig().parentDomain }) .subscribe(data => {
```
lmmrssa: why you are looking after alldocs when you just need configuration with the code of current configuration
Also later we will have configuration synced back to local config so it will not require to request with remote parent

src/app/manager-dashboard/manager-dashboard.component.ts
```diff
@@ -20,10 +20,15 @@ import { PlanetMessageService } from '../shared/planet-message.service';
       <a routerLink="/feedback" i18n mat-raised-button>Feedback</a>
     </div>
     <div class="view-container" *ngIf="displayDashboard && planetType !== 'center'">
-      <h3 i18n>{{ planetType === 'community' ? 'Nation' : 'Center' }} List</h3><br />
+      <h3 i18n *ngIf="showParentList">{{ planetType === 'community' ? 'Nation' : 'Center' }} List</h3><br />
+      <div *ngIf="showParentList">
```
lmmrssa: fix indentation

src/app/manager-dashboard/manager-dashboard.component.ts
```diff
@@ -84,4 +91,10 @@ export class ManagerDashboardComponent implements OnInit {
       }, error => this.planetMessageService.showAlert('An error occurred please try again.'));
   }
 
+  checkRequestAcceptedOrNot() {
+    if (this.userService.getConfig().registrationRequest === 'accepted') {
```
lmmrssa: As we do not sync configuration back to community right now we should look for it on parent. My previous comment mentioned **Later** we will not require to look parent. But right now this feature will not work.

Issue #1060, opened 2018-05-22
Take test submitting answer shows error


Rupesh87:
![course take test](https://user-images.githubusercontent.com/18145146/40355949-7cb070ca-5dd7-11e8-9a17-753f9a911127.png)

lmmrssa:
closing this issue as per comment on PR

PR #1063, opened 2018-05-22
[#1060] take test submit (connects #1060)

lmmrssa:


src/app/courses/courses.service.ts
```diff
@@ -60,7 +60,7 @@ export class CoursesService {
   }
 
   updateSubmission(answer, index: number, close: boolean) {
-    const submission = { ...this.submission, answers: [ ...this.submission.answers ] };
+    const submission = { ...this.submission, answers: this.submission.answers ? [ ...this.submission.answers ] : [] };
```
lmmrssa: `answers: [ ...this.submission.answers ]` is not required at all as it will assign answers of submission to answers itself without any modification

paulbert:
@lmmrssa I'm using the spread operator to copy `this.submission.answers` to make sure the object is not modified until after the Couch request is successful.

I haven't been able to recreate this issue.  `this.submission` is initialized with an `answers` property, and I'm not sure what is making it lose that property.  @Rupesh87 any more detail on what you did before that error occurred would be helpful.

Rupesh87:
@paulbert actually the problem was submissions database was missing in my couchdb so the error occured and after creation of submissions database once again i checked and there was not any error

lmmrssa:
ok then closing this PR

Issue #1000, opened 2018-05-17
Course & exam: Add steps & Add question button placement


paulbert:
It's very easy now to miss adding steps to a course or questions to an exam.  Need to rework.

PR #1070, opened 2018-05-23
[#1000] Add button placement (connects #1000)

lmmrssa:


lmmrssa:
Placed **Add button** next to **submit** so make it more visible.
We might want to add some styling to make it position better later on

PR #1081, opened 2018-05-24
Set the max-height of the attachment resource to same as parent less …

### LINKED ISSUE #857: Course view video shrink to screen size

i5o:
…margin (fixes #857)

this also changes the size of images, do we only need to do it for videos?

### DIFF
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -308,6 +308,10 @@ body {
     height: $view-container-height;
   }
 
+  planet-resources-viewer div * {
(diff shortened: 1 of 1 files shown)


src/styles.scss
```diff
@@ -308,7 +308,7 @@ body {
     height: $view-container-height;
   }
 
-  planet-resources-viewer div * {
+  planet-resources-viewer div > video {
```
paulbert: Could we add this as part of the PlanetResourcesViewer styles instead of the main style.scss file?

Issue #901, opened 2018-05-09
Sync parent and child records


lmmrssa:
Admin should be able to push records (resources, courses, etc) to parent planet and also should be able to fetch records from parent planet.

PR #1089, opened 2018-05-24
[#901] Pull selected resource from parent (connects #901)

lmmrssa:
@paulbert @dogi This is the working example of resource to pull selected resource from parent. If this approach seems good then will work on creating more generic class for any database.


paulbert:
@lmmrssa There is an issue if I fetch a resource, then delete it, I can no longer fetch it.  I'm going to merge this, but it is something we will have to consider in the future.

Issue #1091, opened 2018-05-24
Resend Registration Request button showing incorrectly


paulbert:
As a note, configuration.name !== _user.name.  We had many places in our code treating these properties as equal.  I removed some of them, but this one example and maybe others remain.

paulbert:
Fixed in #1059

PR #1107, opened 2018-05-25
[#1091] Checking child config in parent planet with congfiguration name (connects #1091)

Rupesh87:


src/app/home/home.component.ts
```diff
@@ -125,7 +125,8 @@ export class HomeComponent implements OnInit, AfterViewInit, OnDestroy {
   logoutClick() {
     this.userService.endSessionLog().pipe(switchMap(() => {
       const obsArr = [ this.couchService.delete('_session', { withCredentials: true }) ];
-      if (this.userService.getConfig().name === this.userService.get().name) {
```
lmmrssa: this weird condition was added on https://github.com/open-learning-exchange/planet/pull/578/commits/43e83065d668b3e1a15d67e98c94e7cd359acae3
Yes we have been using same admin and config name in most cases but that will not always be the case. So we need to check config admin and user name

src/app/manager-dashboard/manager-dashboard.component.ts
```diff
@@ -50,7 +50,7 @@ export class ManagerDashboardComponent implements OnInit {
       this.message = 'Access restricted to admins';
     }
     // Check parent if configuration exists
-    this.couchService.post(`communityregistrationrequests/_find`, findOneDocument('name', this.userService.get().name),
+    this.couchService.post(`communityregistrationrequests/_find`, findOneDocument('name', this.userService.getConfig().name),
```
lmmrssa: we should check code instead of name as that is what we have been using to identify planet

paulbert:
Ended up unrelated to #1091, but still a good change

Issue #1090, opened 2018-05-24
Only ASCII characters allowed in username?


Tille88:
If only ASCII or numbers are allowed, the message should indicate that somehow - letters like ü or æ ä å, etc are still considered letters in many non-English alphabets. 

![screen shot 2018-05-24 at 8 39 30 pm](https://user-images.githubusercontent.com/16757941/40489179-bd0d8c36-5f92-11e8-8e3c-59d60d5d98c3.png)

i5o:
Isn't it better if we change the message to something like
"No special characters are allowed"

Tille88:
@i5o Still a bit of technical language for a regular user perhaps, but an improvement in my book. Can't exactly have mapping from any unicode char to closest allowed one if any... so... I don't have a good fix really

paulbert:
Maybe we can update our regex with something like this: https://stackoverflow.com/questions/150033/regular-expression-to-match-non-english-characters

PR #1113, opened 2018-05-25
[#1090] Allow non-english characters in username (connects #1090)

lucy811:
username

### DIFF
--- a/src/app/configuration/configuration.component.ts
+++ b/src/app/configuration/configuration.component.ts
@@ -47,8 +47,8 @@ export class ConfigurationComponent implements OnInit {
     this.loginForm = this.formBuilder.group({
       name: [ '', [
         Validators.required,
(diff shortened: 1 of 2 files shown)


paulbert:
Hmm...I think this only covers a small set of non-english characters.  There's a really nice, simple syntax proposed for ES2018 that TypeScript can compile (maybe a newer version of TypeScript than in our package.json...): https://github.com/tc39/proposal-regexp-unicode-property-escapes (it's mentioned in that SO question as well).

Maybe we can try implementing that?

lucy811:
I spend some time in trying to implement unicode property escapes , but I don’t know how to load 'XRegExp' or 'Regenerate' in angular correctly.
https://stackoverflow.com/questions/36071164/how-to-load-xregexp-in-angular2

Issue #921, opened 2018-05-10
Allow connecting with parent optional on setup


lmmrssa:
During setup of planet, make connecting with nation optional. We can have that done later on the configuration update by admin.

PR #1116, opened 2018-05-25
[#921] Moving nation linking to new function (connects #921)

lmmrssa:
Right now its just removing nation link functionality.
Next step will be transition nation connect.


paulbert:
I think this is outdated, so closing

Issue #752, opened 2018-04-25
Navbar slider delay effect


lmmrssa:
When we slide in/out navbar, narrow white line is visible during transition.

Tille88:
Advice: For debugging, Chrome (as well as FF has great tools for slowing down/replaying/debugging animations, found here:
<img width="675" alt="screen shot 2018-04-25 at 16 04 23" src="https://user-images.githubusercontent.com/16757941/39236249-7c80f56a-48a2-11e8-88b4-a0bdb0c6b2e2.png">

Tille88:
Looking at this, it seem like it is probably a visual illusion. Please look into it from your side as well @lmmrssa 
On open, the start of letters are a white line, which our eyes extrapolate to a narrow white line (at least that is what I think is happening).

To help this, either (from the top of my head):
1. Just make the text a bit softer during the transition (border blur doesn't seem very standard, so just opacity change should be performant enough)
2. Add a kind of blur to make the lines experience bit softer. Adding a slight gradient as: *linear-gradient([…], rgba(x, x, x, 1), rgba(x, x, x, 0))* to the edge.

Would be happy to give it a try. So it depends on how much more complexity it would add, and… if the devices it will be used on are performant enough for it to make much difference for the development time to make sense. 

Let me know :)

Tille88:
Having looked into this a bit more in detail:
- On my laptop with almost nothing else running it fluctuates between 20-40 fps or so.
- The current transition triggers pains all over the place due to the width alterations:
<img width="1214" alt="screen shot 2018-04-28 at 10 08 34" src="https://user-images.githubusercontent.com/16757941/39392132-72039dce-4ada-11e8-8eae-649c7f3b153b.png">

- More costly than the paints are the constant layout updates this transition triggers 
<img width="1040" alt="screen shot 2018-04-28 at 10 22 03" src="https://user-images.githubusercontent.com/16757941/39392135-82733eee-4ada-11e8-94a9-b553b22f0f36.png">

- It would be possible to make improvements, but I think that would be a premature optimization effort when the design may update.
- Would be nice to have a 'on hold list' for such things (adding one now, can be removed if deemed not useful)
- Possible changes that can be attempted:
  - remove the "<- Library" part to a div that have an absolute position (could perhaps get rid of white border line issue)
  - Can make animations running in sequence, avoiding everything to happen at once.
  - have a hidden div off screen with text that is hidden and then using a transform transition would be brought into view
[...]

PR #1123, opened 2018-05-26
navbar slider glitch background color change (connects #752)

Tille88:
Not an ideal solution. The simple change in home.scss overrides default styling, making the glitch less obvious, does not solve the long term issue of that this animation creates new layout calculations every 15ms. 

Changes in home.component.ts are fixes for issue described in https://github.com/open-learning-exchange/planet/issues/1122
These could be ignored if permanent fix to the debug logger are implemented instead.


Tille88:
With improved logger, the rewrite ought to not be needed, simply having the cosmetic stylechange/hack left in this PR

paulbert:
Closing.  May come back to this in the future, but it's not a priority now.

Issue #1093, opened 2018-05-24
Single send configuration function from service


paulbert:
The initial configuration registration and the "Resend Registration Request" buttons should reference a single function on a service which handles sending configuration information to the parent.

Currently the "Resend Registration Request" is sending the incorrect user name to the parent (without the @) so do not use that part of the functionality.

paulbert:
Closing, outdated.

PR #1127, opened 2018-05-28
[#1093] Single send configuration function form service (connect #1093)

MacNew:


paulbert:
Closing this.  It just runs the requests on the manager dashboard in the configuration as well, not taking into account the configuration code is running some of the same requests already.

Issue #1129, opened 2018-05-28
Check parent if the configuration exists should check throw a code (ManagerDashboardComponent.ts)


MacNew:
Because a code is unique. name (username) might not be unique in our application. The same username might be from different Nation.

paulbert:
Same as #1091

PR #1130, opened 2018-05-28
[#1129] Check parent if the configuration exists should check throw a code (connect #1129)

MacNew:
… code


paulbert:
Already fixed as part of #1059

Issue #1080, opened 2018-05-23
Filter users from child planet/community


paulbert:
Users with an @ symbol are admins from a child planet/community.  We should have an option to filter these out so the admin can manage users on the nation.

Rupesh87:
@paulbert where to add the filter? Is that in request table?

lmmrssa:
This issue is about filtering user list not request table. We need to be able to look after users of child planets connected to nation or center. So that admin can control and manage those users easily.

PR #1131, opened 2018-05-28
[#1080] Filter users from child planet (connects #1080)

Rupesh87:


lmmrssa:
It is adding filter selection for each user. For nations or center list of user could be long. Also we don't need all of actions for child users. Only few actions needs to be there.

paulbert:
Closing in favor of #1133

PR #1133, opened 2018-05-28
[#1080] user child filter (connects #1080)

lmmrssa:


src/app/users/users.component.ts
```diff
@@ -68,6 +76,11 @@ export class UsersComponent implements OnInit, AfterViewInit {
     }
   }
 
+  changeFilter(type) {
+    const route = (type === 'associated') ? [ '/users', type ] : [ '/users' ];
+    this.router.navigate(route);
```
paulbert: I don't think we should change routes for the filter, just filter the list of users.  We can add a filter function in `table-helpers.ts` to help.

Issue #1139, opened 2018-05-29
Resource selection after deleting doc is off


paulbert:
![image](https://user-images.githubusercontent.com/9203229/40637075-cc8e80f2-62d0-11e8-8231-a470197af908.png)

PR #1150, opened 2018-05-29
[#1139] Resource selection after delection doc fix (connects #1139)

Rupesh87:


lmmrssa:
If this is about saving resource under selected even after its been deleted then same thing needs to be fixed even on courses and other places.

Rupesh87:
@lmmrssa as mention in issue https://github.com/open-learning-exchange/planet/issues/869 it seems like while updating shelf unclear selection is not implemented in course section so we new to make new issue for it..

paulbert:
@Rupesh87 There are no semicolons which are required by our lint rules.  Please make sure you use the pre push lint hook on any machine you work from (by running `npm run install-hooks`)

Issue #937, opened 2018-05-14
Meaningful level


lmmrssa:
Currently level is 1-5, It is not clear which level is highest or lowest. Also level on resource seems to have different level.
![screen shot 2018-05-14 at 18 55 29](https://user-images.githubusercontent.com/2909490/39999358-a2187d56-57a8-11e8-86ab-6fb637ce4a9b.png)

Level of user can be started from beginner and as per their progress can be advanced further.
If level will be used to suggest or search for resources then we can have multiple level on profile. Or we can have multiple tags used to filter instead of each separate fields for each.

PR #1152, opened 2018-05-29
[#937] Meaningful Level (#937)

ambitdev01:
I have renamed the levels as 'Beginner', 'Intermediate', 'Advanced', 'Expert' , same as we are using for Courses. Please advise if we need to do it differently.
Thank you.

### DIFF
--- a/src/app/users/users-update/users-update.component.ts
+++ b/src/app/users/users-update/users-update.component.ts
@@ -31,7 +31,7 @@ import { languages } from '../../shared/languages';
 })
 export class UsersUpdateComponent implements OnInit {
   user: any = {};
-  educationLevel = [ '1', '2', '3', '4', '5', '6' , '7', '8', '9', '11', '12', 'Higher' ];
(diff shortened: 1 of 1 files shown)


paulbert:
@dogi @lmmrssa I thought this was more related to the user's level of education, in which case we could use the most of the levels from resource (minus "Professional"):

```
export const levelList = [
  'Early Education',
  'Lower Primary',
  'Upper Primary',
  'Lower Secondary',
  'Upper Secondary',
  'Undergraduate',
  'Graduate'
];
```

lmmrssa:
Sure @paulbert I had mentioned about unclarity of level purpose of user. Current list has been used based on Level used on Subject level of Course.
Also this application is not supposed to be for only educational institutes so not sure which levels shall we use :) @dogi  could you have more clear idea on it. Course uses two levels and others use educational level only.

paulbert:
@lmmrssa Ok, I think I'll merge this to clear the issue & PR, but we can discuss & change it later if need be.

Issue #1143, opened 2018-05-29
Meetups: Values for recurring should be capitalized


paulbert:
Since we are showing the values directly on the meetups list, the `recurring` property should be capitalized:

![image](https://user-images.githubusercontent.com/9203229/40637492-450ae366-62d3-11e8-915d-bd386ce61ec9.png)

PR #1157, opened 2018-05-29
[#1143] Meetups: Values for recurring capitalized (#1143)

ambitdev01:


Tille88:
Would it be possible to use a pipe (TitleCasePipe?) or otherwise refactor the logic away from the markup? I would prefer something more declarative, instead of having to figure out what the code does, so that I can just glance over it more easily :)

PR #1158, opened 2018-05-29
xi18n

lmmrssa:


mappuji:
Is it pass the `ng build` with translation?

lmmrssa:
it just generate new message file which then will create other files from crowdin

PR #1167, opened 2018-05-30
Integrate PouchDB

lawrencek0:
To test this PR, please make sure to create a database called "test" and add "pouchIndex": "courses" to the documents in the courses database.

What Works?

- Authentication (Sign up, login)
- Courses module (only the courses services use pouch)


src/app/courses/courses.service.ts
```diff
@@ -26,7 +27,7 @@ export class CoursesService {
   // If the id already matches what is stored on the service, return that.
   // Or will get new version if forceLatest set to true
   requestCourse({ courseId, forceLatest = false }, opts: any = {}) {
-    if (!forceLatest && this.course && courseId === this.course._id) {
+    if (!forceLatest && courseId === this.course._id) {
```
lawrencek0: So, I added a check for this.course because it's null sometimes and gives an error when accessing the id

lawrencek0: For example, if you directly go to the step view page then BehaviorSubject is null and throws an error

paulbert:
Ok here are some thoughts on this:

- I think we should not start with courses on this.  This is actively in development and changing weekly, making the conflicts pile up quickly.  I think feedback would be best.
- Before doing that, maybe let's just add in the PouchDB service into the codebase rather than try to integrate it fully into a section of the app.  I know that has its own challenges, but there's so much changing here that it'll be easier to progress with a small change adding PouchDB
- The folder structure is adding an unnecessary layer (`shared/services`) when I think we should just have the couch and pouch services in a `shared/database` folder or even just a `database` folder.
- It's also easier to follow the changes if each commit isn't so specific (a few that I looked at were one line changes).  We don't want to get too broad, but something in the middle would improve this.
- Thinking about it more, I think we need have separate PRs for adding interfaces.  This will also help to reduce the size of this PR and get it moving faster.

lawrencek0:
Thanks for the feedback @paulbert. Closing in favor of #1355

Issue #1170, opened 2018-05-30
Multiple deleting courses and meetups not working 


Rupesh87:
after deleting multiple courses or meetups it is still showing in table but after refreshing it is deleted...

PR #1177, opened 2018-05-30
[#1170] Multiple deleting courses and meetups fix (connects #1170)

Rupesh87:


src/app/meetups/meetups.component.html
```diff
@@ -32,7 +32,7 @@
         </mat-checkbox>
       </mat-header-cell>
       <mat-cell *matCellDef="let row">
-        <mat-checkbox (change)="$event ? selection.toggle(row._id) : null" [checked]="selection.isSelected(row._id)">
+        <mat-checkbox (change)="$event ? selection.toggle(row) : null" [checked]="selection.isSelected(row)">
```
paulbert: @Rupesh87 This change has broken selection.  The selection section of the Material docs is a guide, so to maintain the connection as the data is updated we changed meetups to control selection based on document id rather than the object itself.

Issue #1176, opened 2018-05-30
 Filter of course does not work for drop down after Reset


MacNew:
same as https://github.com/open-learning-exchange/planet/issues/1174

paulbert:
Please close duplicate issues.

PR #1178, opened 2018-05-30
[#1176] Filter of course does not work for drop down after reset (connect #1176)

MacNew:


paulbert:
Closing because this duplicates #1175

Issue #1171, opened 2018-05-30
login activity _rev is not updated 


MacNew:
It gives us a conflict error because _rev is pointing previous _rev id of login activity. so if a user makes login couple of times and tries to log out then it shows a conflict error. It is happening because on line number 152 of user.service.ts we are updating _rev only if our sessionId is undefined. but in actual, our _rev will change each logout activity of a user.

lmmrssa:
Can you give more detail on when _rev no is not updating. Only case we found so far is when we are on resource or some page it does not reload application on logout so trying to logout later will have conflict. If that is only case then there is PR created to fix issue of resource page.

paulbert:
@MacNew The _rev gets updated when we create a new login_activity object with `newSessionLog` lines 137-146 after every log in, so no there is not an issue when a user logs out then logs back in.

The function you are looking at is for finding the most recent login_activity in the case that the user refreshes the browser (in which case sessionId is unset).

We could add an else block at line 160 which finds the specific document based on the sessionId to ensure the _rev is up to date with the database.

paulbert:
Actually, we can pipe in a map function which sets the _rev to the new value after the first logout attempt if there is an error.

PR #1179, opened 2018-05-30
[#1171] Logout reset session (connects #1171)

lmmrssa:


paulbert:
I'm going to close this in favor of #1182.  I'd rather we maintain the sessionId in the service so we are applying the logout time to the same session (if the user logs in from 2 different devices it could get mixed up).

PR #1192, opened 2018-05-31
[#775] Selecting meetups and courses indication (connects #775)

Rupesh87:


lmmrssa:
made some changes to show count along with button instead of showing it as separate section
screen shot 2018-05-31 at 16 32 53

Issue #1197, opened 2018-05-31
Registration request list: Created Date breaks sort


paulbert:
Likely due to attempting String.toLowerCase on a number.  Throws the follow error:

![image](https://user-images.githubusercontent.com/9203229/40788030-0d6830f8-64bd-11e8-9746-218da3c411c9.png)

lmmrssa:
We should just filter based on certain fields and remove date, _id, _rev, etc from filter

PR #1206, opened 2018-05-31
[#1197] Created Date Sort on Request List (connects #1197)

ambitdev01:


src/app/community/community.component.ts
```diff
@@ -36,13 +36,13 @@ export class CommunityComponent implements OnInit, AfterViewInit {
...
     this.getCommunityList();
-    this.communities.sortingDataAccessor = (item, property) => item[property].toLowerCase();
+    this.communities.sort = this.sort;
     this.communities.filterPredicate = filterSpecificFields([ 'code', 'name' ]);
   }
 
   ngAfterViewInit() {
     this.communities.paginator = this.paginator;
-    this.communities.sort = this.sort;
+    this.communities.sortingDataAccessor = (item, property) => item[property];
```
paulbert: This change should be reverted.  `this.sort` is referencing a `viewChild`, so it must be set in the `ngAfterViewInit` lifecycle hook.

src/app/community/community.component.ts
```diff
@@ -36,13 +36,13 @@ export class CommunityComponent implements OnInit, AfterViewInit {
 
   ngOnInit() {
     this.getCommunityList();
-    this.communities.sortingDataAccessor = (item, property) => item[property].toLowerCase();
+    this.communities.sort = this.sort;
```
paulbert: Rather than reset the sort function to be case sensitive, you should do a type check on the `item[property]` and only to the `toLowerCase()` transformation if it is a string.

paulbert: No, we need to use toLowerCase() when it is a string.  Otherwise it will sort uppercase letters first, then lowercase next.

lmmrssa:
same issue is with nationcomponent as well. but that will be fixed once both of component are merged.

paulbert:
@ambitdev01 This is not good.  We need to check if the value is a string, not convert numbers to strings.  Sending in the number as a string reverts to the standard JavaScript sort function, which does not sort numbers well:

image

I will fix this shortly.

PR #941, opened 2018-05-14
804 update config

ArashDai:

Issue #804, opened 2018-05-01
Update configuration in app


paulbert:
After creating the planet, there isn't an option to update the configuration.  The admin should have this option from the manager dashboard.

PR #1220, opened 2018-06-04
[#804] Configuration update (connects #804)

lmmrssa:
Created new branch from code of #941 as previous PR was giving breaking files when we run linting.


lmmrssa:
@paulbert If we are doing connect to nation later, then we might want to change other configurations as well. Regarding contact info, it is the system admin configuration which we might want to be saved on configuration table and updated to parent as well.

Issue #1226, opened 2018-06-05
Error on request list


lmmrssa:
Navigating to `Connected` or `Rejected` is giving error
![sketch 45](https://user-images.githubusercontent.com/2909490/40950528-dcebd824-6892-11e8-88ad-27d21f9ca81c.png)

lmmrssa:
There is another issue though, which is related to mangoQueries we use. By default its limit is 25, so filter does not seem like working on earth. We need a way to fetch more records. #1241

PR #1240, opened 2018-06-05
[#1226] Do not check parent of center (connects #1226)

lmmrssa:


lmmrssa:
Right now we are checking the planetType and we have 3 level of it. If we do not want to limit to 3 level of hierarchy then we can just have two planetType **Connectable** and **Not Connectable**

Issue #1223, opened 2018-06-05
Cannot delete community from dev


paulbert:
![image](https://user-images.githubusercontent.com/9203229/40949592-c391f63e-683c-11e8-8e54-e8734a264c87.png)

It only runs the `_find` requests.  No errors in the console.

![image](https://user-images.githubusercontent.com/9203229/40949608-d315e340-683c-11e8-8372-b06a98124ab0.png)

PR #1242, opened 2018-06-05
[#1223] Community delete (connects #1223)

lmmrssa:


lmmrssa:
If #1243 is merged then this PR does not require to merge

lmmrssa:
@paulbert Are we still keeping nation as we are merging all request on one page?

PR #1247, opened 2018-06-06
[#881] Angular 6 (fixes #881)

### LINKED ISSUE #881: Update to Angular 6

paulbert:
Note: Vagrantfile does not successfully update Node to 8.x.  Need to run `sudo apt upgrade` from inside the VM to complete the upgrade.

Includes a lot of automatic changes from Angular and Rxjs upgrade scripts.

#881


paulbert:
@sahilph @dogi @mappuji I had to remove the `--arch` flag from the `ng build` command in the crosscompile script. Angular was throwing an error & not running the build command otherwise.

Tested the build and it is still working on a Raspberry Pi, but let me know if we need that flag or something similar for the build command.

lmmrssa:
Delete button on titlebar does not work. It gives undefined as meetup id. Delete on individual row works.
screen shot 2018-06-07 at 16 57 30

Layout of User list is broken. Also Add Team button does not work on each row. All other action including remove team works. Add Team on titlebar also works.
Add Team and Remove Team of individual row does not update.
screen shot 2018-06-07 at 16 58 05

Open profile of another user and edit it. Then profile icon on top header bar takes to profile of edited user. Homepage shows correct profile details.
screen shot 2018-06-07 at 17 04 11
screen shot 2018-06-07 at 17 04 19

lmmrssa:
All these issues seems to be on latest release itself so not related to angular except broken layout of users list view.

Issue #1001, opened 2018-05-17
Input text mask


paulbert:
Helpful package to only allow certain characters in an input

https://github.com/text-mask/text-mask/tree/master/angular2#readme

See it in action here:
https://stackblitz.com/edit/angular-mat-date-mask?file=app%2Fdatepicker-overview-example.ts

lmmrssa:
Looks like masking can be used only for fields like date where we have fixed characters. Or we need to find better masking packages

paulbert:
Closing.  This repo is pretty outdated, and the other text mask libraries are not very good.

PR #1248, opened 2018-06-06
[#1001] Input text mask (connects #1001)

lucy811:
email text mask 
20d39eb5-922a-4ecd-8154-6bdda0fb50b4

1) I think the phone number in different countries may have different phone number formats. so I didn’t change phone to the US phone number format.
2) I tried to implement date (e.g birthdate format in user profile), but I have no success yet.
3) If we plan to change some input in configuration.component, we must import { TextMaskModule } from 'angular2-text-mask’; in configuration.module.ts.  But I don’t know how to create a new configuration.module.ts.


src/app/app.module.ts
```diff
@@ -30,6 +32,8 @@ import { CoursesService } from './courses/courses.service';
     HttpClientModule,
     BrowserAnimationsModule,
     MaterialModule,
+    FormsModule,
```
paulbert: We don't need these here, just in the specific modules that use them (in this case UsersModule which you've already added them to).

lucy811:
email text mask
email

Issue #1250, opened 2018-06-06
CourseView: view resource button list all resources


paulbert:
If there are multiple resources on a step, let's have the "View Resource" button open a menu with a list of links to all of the step's resources.

https://material.angular.io/components/menu/overview

PR #1266, opened 2018-06-07
[#1250] Mulitple Resources View in Course Step (connects #1250)

ambitdev01:


lmmrssa:
Looks good as per issue it was asked to add menu. If we want to show dialog of resource list then we might want to change some code. Else it looks good as per issue.

Issue #995, opened 2018-05-17
Add material badge to notification icon


paulbert:
https://material.angular.io/components/badge/overview

Tille88:
Looked into this, and it will have to wait until Angular 6 is used, since the badges are only available in Angular Material 6

5.2.4:
![screen shot 2018-05-19 at 12 50 40 pm](https://user-images.githubusercontent.com/16757941/40265495-79c45438-5b63-11e8-98ef-39b2bec7bfcc.png)

6.0.1:
![screen shot 2018-05-19 at 12 50 57 pm](https://user-images.githubusercontent.com/16757941/40265498-84c1cb54-5b63-11e8-82d6-1d1b0ec17325.png)

I kept getting this error if trying to use that before Angular 6 update is complete:
https://stackoverflow.com/questions/49899641/type-elementref-is-not-generic

PR #1270, opened 2018-06-08
Material badge notification icon (Connects #995)

Tille88:
Basically only two changes, as seen by diff, even though some ver. 6 commits listed, but figured since will be squashed anyway it wouldn't matter:

1) Material icon badge, color could be either red or yellow, red looked better, same location as facebook badge location chosen

2) Reordering of import list to be in alphabetic order, since looking for things there was not... super easy.

Creating new issue regarding how notifications are handled on click.


Tille88:
Travis is not playing nicely today, one of the steps keeps failing, but these minor changes seems unlikely to have broken anything after latest master tests passed...

lmmrssa:
@Tille88 sorry you will have to create new PR of it. While trying to rebasing it have issue on this branch.

PR #1278, opened 2018-06-08
[#1229] Buttons in dialog moved to the right, and correct order of them (fixes #1229)

### LINKED ISSUE #1229: Group buttons on the right side on modals?
Right now the buttons are on the left side, why not on the right side like everyone else?
image

i5o:
before:
image

after:
image


lmmrssa:
changes looks good. @paulbert Shall we also align action buttons on other forms.

PR #1281, opened 2018-06-09
[#1237] Update only the resource rated (fixes #1237)

### LINKED ISSUE #1237: Rating a resource takes me back to the first one
the gif shows what I mean
peek 2018-06-05 00-09

i5o:


lmmrssa:
Once rating is done on resource list it is remove all other resources from list.
screen shot 2018-06-11 at 13 20 29

paulbert:
The `updateResources` function will trigger an update to the full resource list, returning an array with only the resources in the filter.  The ResourcesRating component is used both when the view is one resource & all resources.

We need to either make `ResourcesService` a little more flexible, or continue to request all resources from `updateResources` and find the doc with the correct _id in `ResourcesView` rather than just taking the first element.

i5o:
I just pushed a solution @paulbert, just wondering if this makes sense

PR #1283, opened 2018-06-09
[#1230] Show alert when the user is not authorized to list resources/meetups/etc (fixes #1230)

### LINKED ISSUE #1230: Add feedback that my user needs to be approved to create resources
I created a new user, and tried to create a resource, there wasn't feedback that I need to be approved by someone.

I just wasn't able to create it

i5o:


i5o:
@lmmrssa @paulbert is this better now? I think it makes more sense

paulbert:
@lmmrssa I went ahead and changed the users component so it also reacts to security from the DB rather than from our app code.  We can set security in another PR if needed.

PR #1286, opened 2018-06-09
Correct answer for multiple choice questions (connects #668)

paulbert:


lmmrssa:
@paulbert There is one issue. When removing choices, it removes always first choice.
screen shot 2018-06-11 at 12 38 33

paulbert:
Ah yes, I removed the index from the *ngFor thinking it didn't need it anymore.  @lmmrssa Could you check to confirm it is working now?

lmmrssa:
Removing choice worked but editing it again is not getting the choice options
screen shot 2018-06-11 at 22 31 36

PR #1291, opened 2018-06-11
[#1260] Updated Exam Page Back Button (connects #1260)

ambitdev01:


src/app/exams/exams-add.component.ts
```diff
@@ -125,4 +125,8 @@ export class ExamsAddComponent implements OnInit {
     this.router.navigate([ this.coursesService.returnUrl ]);
   }
 
+  backToCourse() {
```
lmmrssa: please add `this.coursesService.reset();` if not reset it next course edit will trigger to same course

PR #1293, opened 2018-06-11
Disable submit button upon submission (fixes #1200)

### LINKED ISSUE #1200: Disable action buttons once event is triggered and show indicator
We should disable buttons once someone trigger event to prevent immediate concurrent event request.
Also as some request will take a while to return response we should have some indicator showing action being carried on.

gcantieni:
### Current state
Added `isSubmitted` variable and bound the `disabled` attribute of the `Submit` button to it. Now it should be impossible to add a resource multiple times.

### Next step
As suggested by @lmmrssa in #1200
> as some request will take a while to return response we should have some indicator showing action being carried on.

So next step is to add some kind of loading sign. What type of loading signal do we use in other parts of the app?

### DIFF
--- a/src/app/resources/resources-add.component.html
+++ b/src/app/resources/resources-add.component.html
@@ -106,7 +106,7 @@
       <div *ngIf="pageType==='Update'">
         <mat-checkbox (change)="deleteAttachmentToggle($event)" i18n>Delete File</mat-checkbox>
       </div>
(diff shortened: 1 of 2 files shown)


gcantieni:
Also as @lmmrssa commented:
> This error is also happening with actionbar so once we figure out it on resources-add, we need to implement it for all buttons.

So I was wondering all the places that need to be implemented. Is there a solution that would work for all the buttons without manually changing every submit button?

PR #1315, opened 2018-06-12
[#1304] buttons in user component are now next to each other (fixes #1304)

### LINKED ISSUE #1304: Users table button alignment/cut off
Since the upgrade to 6, some of the buttons are running into one another and cut off:

image

i5o:


lmmrssa:
@i5o It fixes layout broken but still it has buttons next to each other so can you add something for it.

i5o:
Okay, so I added something to styles.scss I think this is what we need
image

image
image

src/styles.scss
```diff
@@ -334,4 +334,8 @@ body {
   mat-dialog-actions {
     justify-content: flex-end;
   }
+
+  .mat-raised-button {
```
paulbert: This is overriding margins added via other classes to some buttons.  Let's just have this as part of the component style.

i5o: What do you mean by that @paulbert? 
Only adding it to the users component?

i5o:
Hey @paulbert I moved the css to the component

lmmrssa:
@paulbert @i5o can we have this style implemented all places or as global as all buttons are touching each other

PR #1321, opened 2018-06-12
Push course

lmmrssa:
This is just for the testing purpose right now.
It works for any new courses pushed to community.
If you push same course again it will have conflict.
It is non-replicator approach.


lmmrssa:
Requires update once #1333 is merged

lmmrssa:
Will have to update based on #1465

PR #1333, opened 2018-06-13
Manage replicator task

lmmrssa:


lmmrssa:
On registration create replicator to get status of request
On manager sync will push all db to parent
Use single sync service with course resource pull as well
Change replicator from continuous to one time

src/app/shared/sync.service.ts
```diff
@@ -64,24 +64,9 @@ export class SyncService {
   private syncParams(opt, credentials, type) {
     const dbSource = opt.dbSource || opt.db;
     const dbTarget = opt.dbTarget || opt.db;
-    const parentDomain = opt.parentDomain || this.userService.getConfig().parentDomain;
-    const code = opt.code || this.userService.getConfig().code;
```
lmmrssa: we will not have userService when we are doing configuration so we do not have code as well. That's why had kept code on opt.

paulbert: See line 40.  It's included there.

lmmrssa: sorry @paulbert I'm pointing to code you use on line 95. you need opt for code as well.

src/app/shared/sync.service.ts
```diff
@@ -23,76 +22,85 @@ const passwordFormFields = [
...
-    private planetMessageService: PlanetMessageService
+    private dialogsFormService: DialogsFormService
   ) {}
 
-  private defaultConfig = {
-    'create_target':  false,
-    'continuous': false,
-    'owner': this.userService.get().name
-  };
+  confirmPasswordAndRunReplicators(replicators) {
```
lmmrssa: there will be cases when we do not need to ask for password confirmation. Such as when we are doing setup of application. So we will require replicator to be able to run without password confirmation

Issue #1302, opened 2018-06-12
UserUpdate: Do not allow birth date in the future


paulbert:
![image](https://user-images.githubusercontent.com/9203229/41265407-1d9f2902-6dc0-11e8-919f-2cad25b5de61.png)

PR #1334, opened 2018-06-13
[#1302] UserUpdate: Do not allow birth date in the future (#1302)

ambitdev01:
For now, I have displayed dates for the last 30 years,
i.e. it will show dates starting 1988, Is that fine @paulbert  or do we need to change that ?
Please advise.

CC: @lmmrssa , @dogi 

Thank you

### DIFF
--- a/src/app/users/users-update/users-update.component.ts
+++ b/src/app/users/users-update/users-update.component.ts
@@ -44,7 +44,6 @@ export class UsersUpdateComponent implements OnInit {
   file: any;
   roles: string[] = [];
   languages = languages;
-  minDate = new Date(2000, 0, 1);
   maxDate = new Date();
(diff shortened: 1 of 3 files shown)


src/app/users/users-update/users-update.component.ts
```diff
@@ -44,6 +44,8 @@ export class UsersUpdateComponent implements OnInit {
   file: any;
   roles: string[] = [];
   languages = languages;
+  minDate = new Date(2000, 0, 1);
```
lmmrssa: you have min date set 2000 and max date is -30 which is less that 2000 so this does not work at all.
We do not need min date also max date should need to be today.
We do not want to have more control on DOB.

Issue #1332, opened 2018-06-13
Drop-down can be added on profile option


NirmalKumarJoshi:
As of now when we click on profile icon we are directed to users page, rather we can add a drop-down which has following options:

- Full Name
- Details (which redirects to current **Users** page)
- Edit Profile (redirects to  **Edit Profile** page)
- Change Password or Settings (redirects to **Change Password** page )

NirmalKumarJoshi:
Working on this issue.

PR #1335, opened 2018-06-13
[#1332] Drop Down List on Profile Option (connects #1332)

ambitdev01:


src/app/home/home.component.ts
```diff
@@ -167,4 +176,128 @@ export class HomeComponent implements OnInit, AfterViewInit, OnDestroy {
     },  (err) => console.log(err));
   }
 
+  onSubmit(credentialData, userDetail) {
```
ambitdev01: These has been copied from users profile which is duplicated. So I think we need to make it as a service.

paulbert: I think a directive would be better, but I agree we should consolidate this code into one file

src/app/home/home.component.html
```diff
@@ -14,10 +14,15 @@ <h1 i18n>Planet Learning</h1>
     <img *ngIf="currentFlag" src="assets/flags/{{currentFlag}}.png" alt="{{currentLang}}"
       title="{{currentLang}}" />
   </button>
-  <button mat-icon-button routerLink="/users/profile/{{user.name}}">
+  <button mat-icon-button [matMenuTriggerFor]="userProfile">
     <img *ngIf="user._attachments; else accountIcon" class="profile-image-large" [src]="userImgSrc">
     <ng-template #accountIcon><mat-icon>account_circle</mat-icon></ng-template>
   </button>
+  <mat-menu #userProfile="matMenu" [overlapTrigger]="false">
+    <button mat-menu-item routerLink="/users/profile/{{user.name}}">{{user.name}}</button>
```
paulbert: Let's just have the text be "View Profile"

Issue #1261, opened 2018-06-07
Remove NationComponent


paulbert:
With most of the functionality going to CommunityComponent, we can remove the NationComponent.

One thing that hasn't made it into CommunityComponent: listing the connected communities of child components.

PR #1337, opened 2018-06-13
[#1261] Remove nation component (connects #1261)

lmmrssa:
Moved child communities as dialog list.
Updated dialog list to have non selectable row and opts on request.


src/app/manager-dashboard/manager-dashboard.component.ts
```diff
@@ -13,7 +13,7 @@ import { debug } from '../debug-operator';
 @Component({
   template: `
     <div *ngIf="displayDashboard">
-      <span *ngIf="planetType !== 'community'">
+      <span *ngIf="planetType === 'community'">
```
paulbert: I don't think we want this change (we should only see the request button from nations & centers)

src/app/manager-dashboard/manager-dashboard.component.ts
```diff
@@ -13,7 +13,7 @@ import { debug } from '../debug-operator';
...
         <a routerLink="/requests" i18n mat-raised-button>Requests</a>
         <a routerLink="/associated/{{ planetType === 'center' ? 'nation' : 'community' }}"
```
paulbert: Need to remove this link to the now removed NationComponent

PR #1339, opened 2018-06-13
Fix translations

paulbert:
Fixing situations where `i18n` tags were used for elements with only interpolation (which should not be translated & causes errors).

### DIFF
--- a/src/app/users/users.component.html
+++ b/src/app/users/users.component.html
@@ -103,10 +103,7 @@
           <a
             *ngIf="element.doc.roles.length > 0 && !filterAssociated"
             (click)="setRoles(element.doc, element.doc.isUserAdmin ? element.doc.oldRoles : ['manager'])"
-            mat-raised-button color="primary" i18n>
-            <!--Convert boolean to number for i18n-->
-            {+element.doc.isUserAdmin, select, 1 {Demote} 0 {Promote}}
-          </a>
+            mat-raised-button color="primary" i18n>{+element.doc.isUserAdmin, select, 1 {Demote} 0 {Promote}}</a>
         </span>
         <ng-container *ngIf="!filterAssociated">
           <button mat-raised-button color="primary" *ngIf="!element.myTeamInfo else addTeam" (click)="addTeams([element.doc._id])">Add Team</button>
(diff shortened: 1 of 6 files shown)


src/i18n/messages.ar.xlf
```diff
@@ -52,7 +52,7 @@
         </context-group>
         <target xml:lang="ar" state="needs-translation">Are you sure you want to <x id="ICU" equiv-text="{data.changeType, select, accept {...} reject {...} unlink {...} delete {...}}"/>
       the <x id="ICU_1" equiv-text="{data.amount, select, single {...} many {...}}"/>
-  
+
```
lmmrssa: not sure if we want to update xlf file manually or let it be handled by crowdin or xi18n

src/app/users/users.component.html
```diff
@@ -103,10 +103,7 @@ <h1 i18n>All Users</h1>
           <a
             *ngIf="element.doc.roles.length > 0 && !filterAssociated"
             (click)="setRoles(element.doc, element.doc.isUserAdmin ? element.doc.oldRoles : ['manager'])"
-            mat-raised-button color="primary" i18n>
-            <!--Convert boolean to number for i18n-->
-            {+element.doc.isUserAdmin, select, 1 {Demote} 0 {Promote}}
-          </a>
+            mat-raised-button color="primary" i18n>{+element.doc.isUserAdmin, select, 1 {Demote} 0 {Promote}}</a>
```
lmmrssa: We might want to remove this one as well from i18n. Also we need to update style guide README to not use i18n if we do not have any string inside element. Will create new issue for that.

src/i18n/messages.ar.xlf
paulbert: Long term I agree we do not want to edit this manually, but in the short term I modified this to help @mappuji get started

src/app/users/users.component.html
paulbert: This is the `i18n` select syntax for having a few strings translated based on a variable value, so it is valid.

src/i18n/messages.ar.xlf
lmmrssa: ok then @mappuji will have to update other xlf file to match it as he is building multiple language

PR #1341, opened 2018-06-14
Fix center.media.mit.edu admin

paulbert:
Since center tries to log into the "parent" and fails, it loses the session immediately pushing you back to the login screen on every route change.


lmmrssa:
I was thinking to remove parentDomain url from the configuration so we do not need to check planetType and our planetType can have any depth. But that will be different talk

paulbert:
@lmmrssa I agree that would be ideal, but I think our time is better spent elsewhere since we won't be setting up centers too often.  And this has the added bonus that it should fix the pesky popup on Earth when we login.

PR #1344, opened 2018-06-14
[#668] Exam errors leader view (connects #668)

paulbert:
Currently any user can access this, so we will have to restrict that at a later time.

This adds the view for a course leader to see what questions learner's are having difficulty with.


paulbert:
Incomplete exam breaks it, too:
image

paulbert:
@lmmrssa Note when testing: I had to change the `Submissions` doc format, so you'll have to clear that DB and retake exams to see this working.  Any exam created since our last commit (v0.2ish) should work, though.

PR #1347, opened 2018-06-14
Auto send resource course on accept

lmmrssa:


lmmrssa:
will require to update based on #1465

PR #1355, opened 2018-06-14
Add PouchDB

lawrencek0:
This PR will only add the base service so that we can start to integrate PouchDB with Angular

- I did not move the couch service to the database folder because by doing that there will have to be a lot of one line changes to many files. I think it's better to do that when most of the services have been replaced by PouchDB.
- Angular 6 removed some polyfills because of which PouchDB was not working so I had to add some polyfills as suggested on pouchdb/pouchdb#7263


src/app/shared/database/pouch.service.ts
```diff
@@ -0,0 +1,65 @@
...
+PouchDB.plugin(PouchDBFind);
+
+@Injectable()
+export class PouchService {
+  private baseUrl = environment.couchAddress;
+  private localDB;
+  private remoteDBs = [];
+
+  constructor() {
+    this.localDB = new PouchDB('local-pouchdb');
```
paulbert: I wonder if this approach of creating one database for PouchDB and storing all dbs from Couch here will cause similar problems that we saw with BeLL-apps of id conflicts.  @dogi wondering what you think?

lawrencek0: We could create one PouchDB instance per remote database if that's an issue.

paulbert:
@lawrencek0 I was about to merge this, and then realized you were returning from Array.forEach a few times which is always undefined.  I went ahead and removed the returns in https://github.com/open-learning-exchange/planet/pull/1355/commits/825407459f6921e349e21c04736733a5e2ec11bd, but if you intended to return a value from those functions you can change that in an upcoming PR.

PR #1357, opened 2018-06-15
[#1356] Course rating (connects #1356)

lmmrssa:


lmmrssa:
sure @paulbert that was just a starting of course rating to see if we would want this feature or not.

Issue #1372, opened 2018-06-18
Meetup Form Update RevID


ambitdev01:
On Meetup form,
 while we are on update form, we can save the data without changing any value. It updated the rev value in couch.So, I think we should not allow to do that until the value has been changed.

PR #1373, opened 2018-06-18
[#1372] Meetup Form Update RevID (connects #1372)

ambitdev01:


lmmrssa:
@paulbert this is something we might want to have on all of the components may be with https://github.com/open-learning-exchange/planet/issues/1200

Issue #1388, opened 2018-06-19
Introduce source and update time


lmmrssa:
For sharing resource and course between planets we need to introduce these fields to keep track
**sourcePlanet** code of planet where it is created
**createdDate** time of creation
**updatedDate** last updated time
**resideOn** code of planet where copy is made

Only for document from another planet
**copyOf** id of original document (only for local copy)

PR #1389, opened 2018-06-19
[#1388] Fields added to track source (connects #1388)

lmmrssa:


lmmrssa:
@paulbert There is already **articleDate** on **resources** database. Shall we update to **createdDate** so we do not duplicate.

PR #1394, opened 2018-06-19
Check Latest commit (Fixes #1354)

### LINKED ISSUE #1354: docker: latest tag has to be always also latest commit
right now every build on master triggered commit or tag will overwrite latest tag ... also in the case of rebuild of an older commit on travis ...

... in this case latest just points to the latest successful build but not to the latest commit,
which is wrong

ToDo:
make sure when latest tags get written that they are also lastest commits

sahilph:
This Fixes #1354 .
Instead of checking only the branch name, we now check the latest Commit hash.

### DIFF
--- a/.travis/travis_utils.sh
+++ b/.travis/travis_utils.sh
@@ -26,14 +26,14 @@ prepare_ci(){
 
 push_a_docker(){
   build_message pushing $1
-	docker push $1
-	build_message done pushing $1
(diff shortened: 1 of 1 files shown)


sahilph:
@dogi Fixed that

mappuji:
Congrats @sahilph 🎉

PR #1402, opened 2018-06-20
[#1381] Add margin to mat-raised-buttons (fixes #1381)

### LINKED ISSUE #1381: NationComponent buttons cut off
Similar to #1315.  Making a class that can be applied to tables would be good to reduce some of the duplication.  We don't want to broadly change all buttons.

image

i5o:


i5o:
@paulbert wondering if this is good now

Issue #877, opened 2018-05-08
Confusing cancel button in rating popup


paulbert:
The user clicks the ratings, then there is a popup to give them an option to add a comment.

Right now there are a few things that could possibly be confusing:

- The star rating is displayed & the user can change it, but their initial rating will be added to the database.
- There are OK/Cancel buttons but the initial rating is already on the database so the user cannot cancel that.

There could be more cues to tell the user that the popup is for an optional comment to add to their star rating.

singharpita:
Rating dialog box has been changed as attached below.
Please suggest if the text has to be changed. @paulbert @lmmrssa 

![screenshot from 2018-06-21 10-26-55](https://user-images.githubusercontent.com/27276132/41698366-b5dba888-753d-11e8-97b9-784e1d085013.png)

PR #1410, opened 2018-06-21
[#877] resource-rating (connects #877)

singharpita:


lmmrssa:
Looking after the screenshot on the issue. It looks more clearer than before. @paulbert do you feel need to change any text

Issue #1425, opened 2018-06-22
Remove "Add Question" from ToolBar


nirojdyola:
There are two same buttons "Add Question" that does the same thing. We can remove from the toolbar.
![screen shot 2018-06-22 at 7 57 48 am](https://user-images.githubusercontent.com/7495421/41754534-71630178-75f2-11e8-8eb9-011579a839a7.png)

paulbert:
Closing since we can keep this redundancy

PR #1426, opened 2018-06-22
[#1425] Removed Add Question buttons from Toolbar (connects #1425)

nirojdyola:


paulbert:
It was decided to keep this redundancy

Issue #1422, opened 2018-06-21
Language switch on header should switch domain


lmmrssa:
Now we have multi-language on our production release, we can make language switch to navigate between these different domain instances.
Sample links
https://dev.media.mit.edu/eng
https://dev.media.mit.edu/nep
https://dev.media.mit.edu/ara

PR #1427, opened 2018-06-22
[#1422] language switch (connects #1422)

singharpita:


paulbert:
@singharpita This looks ok, but we ended up working in parallel (I should have assigned this issue to myself).  Closing in favor of #1433

Issue #1421, opened 2018-06-21
Meetups: no delete if not admin or owner


paulbert:
User's should only be allowed to delete a meetup if they are the owner of the meetup or they are an admin.

![image](https://user-images.githubusercontent.com/9203229/41725758-91b1c5cc-753e-11e8-9807-e4b48d620516.png)

PR #1430, opened 2018-06-22
[#1421] meetup check owner on delete (connects #1421)

singharpita:
Shall we control **Delete** button on title bar on selection of meetups by disabling or filter delete on component function.
Or do we need to have this button at all? @paulbert @lmmrssa


paulbert:
Needed one small change to avoid accessing the service directly in the template

paulbert:
@singharpita This works for fixing the issue, but you raise a good point so I created a new issue #1453

PR #1433, opened 2018-06-22
New language menu

paulbert:
Switches between the different languages as set up by Mappuji.

- Language names are now listed in their own language (or at least what Google says it is)
- Removed the flags and added the language in text to the top bar.  Flags !== language for many cases, so I don't think we should use them.

### DIFF
--- a/src/app/home/home.component.ts
+++ b/src/app/home/home.component.ts
@@ -26,9 +26,8 @@ import { debug } from '../debug-operator';
   ]
 })
 export class HomeComponent implements OnInit, AfterViewInit, OnDestroy {
-  languages = [];
-  currentFlag = 'en';
-  currentLang = 'English';
+  languages = languages;
+  currentLanguage: any = { name: 'English', shortCode: 'eng' };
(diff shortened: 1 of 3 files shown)


src/app/home/home.component.ts
```diff
@@ -62,15 +61,9 @@ export class HomeComponent implements OnInit, AfterViewInit, OnDestroy {
...
-    this.languages = (<any>languages).map(language => {
-      if (language.servedUrl === document.baseURI) {
-        this.currentFlag = language.shortCode;
-        this.currentLang = language.name;
-      }
-      return language;
-    }).filter(lang  => {
-      return lang['active'] !== 'N';
-    });
+    this.currentLanguage = this.languages.find(language => {
```
lmmrssa: can we also filter out the current language from the language list. As it does not look nice to let user switch to same language.

lmmrssa:
Tested latest build of it and it seems to be working just few questions I have mentioned on the codes. could you please go through it.

Issue #1241, opened 2018-06-05
Limit of _find query


lmmrssa:
Default limit of _find query is **25** so we need some way to be able to fetch more records. May be by multiple request or get count of total records.

PR #1434, opened 2018-06-22
[#1241] Limit in  _find Query (connects #1241)

nirojdyola:
Implemented in users. Should I change in entire component ? if it is good to use, then I"ll change it.
screen shot 2018-06-22 at 10 18 01 pm

### DIFF
--- a/src/app/users/users.component.ts
+++ b/src/app/users/users.component.ts
@@ -112,7 +112,12 @@ export class UsersComponent implements OnInit, AfterViewInit {
   }
 
   getUsers() {
-    return this.couchService.post(this.dbName + '/_find', { 'selector': {} });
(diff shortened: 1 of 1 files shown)


src/app/users/users.component.ts
```diff
@@ -112,7 +112,12 @@ export class UsersComponent implements OnInit, AfterViewInit {
   }
 
   getUsers() {
-    return this.couchService.post(this.dbName + '/_find', { 'selector': {} });
+    return this.couchService.allDocs(this.dbName)
```
lmmrssa: non admin will not be able to get alldocs that's why we have introduced _find so this will not work

paulbert:
Closing in favor of #1485

PR #1445, opened 2018-06-23
[#1431] Created the PlanetLanguage component (Fixes #1431)

### LINKED ISSUE #1431: Language switch option on login screen
We should give option of language switch even on login/configuration pages.

i5o:
planet-language is missing in configuration


src/app/shared/planet-language.module.ts
```diff
@@ -0,0 +1,17 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { PlanetLanguageComponent } from './planet-language.component';
+import { MaterialModule } from './material.module';
```
i5o: wondering if common/material is needed, didn't test without them

i5o: it doesn't work wihthout common/material

PR #1449, opened 2018-06-25
[#1448] Clean configuration replicator add replicator on admin login (connects #1448)

lmmrssa:
We were using feedback sync on configuration which is not required anymore so removed it from there. And to update configuration of community when nation approves, Used the same id between nation and community for configuration. Then on admin login it will create replicator until community configuration status is not updated from pending to something else.


paulbert:
Missed the reason why you were setting the _id of configuration to match the communityregistrationrequest in https://github.com/open-learning-exchange/planet/pull/1449/commits/d5b6e1a9be8da4f7a67c7a8b4d9eb76369f9b146.  Making the adjustment now, but I need to go to sleep so won't be able to test & merge this tonight.

lmmrssa:
@paulbert Updated this issue so that when deleting/rejecting community it does not delete user from parent. This helps when we resend request as we are not sending user credentials at that time.
Also checking parent configuration during request resend to make sure it does not conflict on insert or update of configuration on parent.

paulbert:
@lmmrssa We should delete the user as well.

If the community has an issue and fully reinstalls or signs up with a different name that would mean the admin user is left on the planet unused.  Worst case scenario is that this would silently block future community installs (which is likely if communities use a generic name like "admin" for the administrator user), but it still is not a great idea to leave inactive users on our nations.

PR #1458, opened 2018-06-26
[#1451] Set max-width of img to 100px (fixes #1451)

### LINKED ISSUE #1451: UsersComponent "Profile Image" column set width
The "Profile Image" column is much wider than it needs to be.  Please set it to a max-width of 100px.

i5o:


paulbert:
I should have added an image to clarify in the issue description, I meant setting the profile image column in this table to 100px:

image

i5o:
image

src/app/users/users.component.ts
```diff
@@ -21,6 +21,12 @@ import { debug } from '../debug-operator';
     .mat-column-select {
```
paulbert: We'd like to set the column max-width similar to this `.mat-column-select` class but for `.mat-column-profile`.  We should revert the img tag to use the `.profile-image` class.

Issue #847, opened 2018-05-03
Resource attachments in separate table


paulbert:

paulbert:
We are going to take a different approach to storing resources, so closing.

PR #1480, opened 2018-06-26
[#847] separate attachments resources (connects #847)

ambitdev01:


paulbert:
Closing since this would be impractical at this stage

PR #1481, opened 2018-06-26
[#1472] Invalid password warning on replicator (connects #1472)

lmmrssa:


lmmrssa:
as @gcantieni has already taken #1527 will wait for it to complete and make changes on this PR to pass error

Issue #1450, opened 2018-06-25
Do not change "createdBy" on update


paulbert:
Found in MeetupsAddComponent.

If we are updating the document after creation, the `createdBy` field should not change

lmmrssa:
@paulbert @nirojdyola @ArashDai This issue does not look valid anymore as we are only allowing to update meetup by its owner.

paulbert:
Also retested and it seems to be working as it should now (and with the link not showing unless someone types update into the url only the creator will update meetups anyway)

PR #1483, opened 2018-06-26
[#1450] No change on CreatedBy field for Meetup while update (connects #1450)

nirojdyola:


paulbert:
Retested the current code and it is working.  Not sure why the error happened before, and we changed to only show the manage link to the course creator anyway so it's unlikely other users will edit meetups.

Closing this so we don't clutter our component.

PR #1486, opened 2018-06-26
Default user for android

paulbert:
On community configuration creates a `satellite` user with the learner role and a 4 digit pin for its password that is also stored on the database configuration.  That pin is displayed on the manager dashboard.

In addition as part of cleanup when deleting the community the satellite user & the configuration pin value are deleted.


lmmrssa:
We might want to have feature to reset/regenerate the pin of satellite user but that can be different issue.

Issue #1419, opened 2018-06-21
Error for invalid date


paulbert:
On birth date field for UserUpdate an invalid value produces a "This field is required" rather than "Invalid date" error:

![image](https://user-images.githubusercontent.com/9203229/41724608-dfa434ac-753b-11e8-8e63-ce0f05035947.png)

PR #1489, opened 2018-06-26
[#1419] Error for invalid date (connects#1419)

lucy811:
If we use datepicker, when we input alphabet (q, w..) or some special characters( @, #..), input will pass null or undefined as birthdate value. 

[Date picker input and change events](https://stackblitz.com/angular/aaledypnmljg?file=app%2Fdatepicker-events-example.ts) can show this situation.
input

So I revise form-error-message, error message can be suitable to input nothing or to input something not date.
screen shot


lmmrssa:
Once merged we will have to make this changes reflected to https://github.com/open-learning-exchange/planet/pull/1493 as well

Issue #1484, opened 2018-06-26
UsersComponent: Remove unnecessary if statement


paulbert:
In `initializeData` the if block on line 121 is not necessary.  `Array.filter` returns an array of items that return true for the value passed into it.

paulbert:
Currently fixed as part of #1486 so closing this issue.

PR #1491, opened 2018-06-26
[#1484] UsersComponent: Remove unnecessary if statement (connects #1484)

lucy811:


paulbert:
There was a bit of wires crossing and this is actually fixed in #1486 since I needed to change that statement anyway.  I'll close this since PR it also has a problem.

This change removes the filter functionality entirely.  The conditional statement is still required, but rather than using it as part of an if block we should return the value of the conditional in the function passed to `Array.filter`.  So when that value is true, it keeps that item of the array, and when it is false it discards it.

There are some good examples of `Array.filter` uses in the MDN web docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

Issue #1455, opened 2018-06-26
Changing language should load same page


paulbert:
When changing the language in the top navbar it then loads the dashboard.  We should instead read the current route and load that in the new language.

i5o:
working on this one, just fyi

PR #1497, opened 2018-06-27
[#1455] Load the same page on changing Language (connects #1455)

nirojdyola:
Loading the same page while changing the language. But how to test it, I can not test in my end. 
I am using vagrant.
@lmmrssa @paulbert 
screen shot 2018-06-27 at 7 47 52 pm

### DIFF
--- a/src/app/shared/planet-language.component.html
+++ b/src/app/shared/planet-language.component.html
@@ -4,7 +4,7 @@
 
   <!--Languages dropdown menu-->
 <mat-menu #languageMenu="matMenu" [overlapTrigger]="false">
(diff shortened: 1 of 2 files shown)


i5o:
This won't work if you are on "/eng", it will redirect you to "/eng/esp". I'm currently working on this too (you should have checked the comment in the issue), but feel free to finish it.

How I tested was easy, build a docker image with all the languages and my current code, then use docker to load the planet.yml

i5o:
replaced by: https://github.com/open-learning-exchange/planet/pull/1501

Issue #1465, opened 2018-06-26
Replicate resources with course


paulbert:
A course with attached resources should replicate the resources as well

PR #1498, opened 2018-06-27
[#1465] Sync resource along with course (connects #1465)

lmmrssa:


src/app/courses/courses.component.ts
```diff
@@ -254,8 +254,16 @@ export class CoursesComponent implements OnInit, AfterViewInit {
   }
 
   fetchCourse(courses) {
-    this.syncService.confirmPasswordAndRunReplicators([ { db: this.dbName, items: courses, type: 'pull', date: true } ])
-    .subscribe((response: any) => {
+    let resources = [];
+    courses.map(course => {
+      course.steps.map(step => {
```
paulbert: @lmmrssa Just a note: You can only omit the return with an arrow function if you omit the curly brackets.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

paulbert: Ah, my mistake you aren't returning anything just applying the side effect for the resources array.  We should use `forEach` instead as that doesn't return anything & more quickly communicates to future developers that nothing is being returned

Issue #1463, opened 2018-06-26
Exam question Input with MC choices


paulbert:
After switching from multiple choice to input the choices are still present.  They should be cleared after the switch:

![image](https://user-images.githubusercontent.com/9203229/41885592-a252bd60-78c6-11e8-9f81-1b1c4236753e.png)

paulbert:
In addition, the `correctChoice` property should be removed for an "Input" type question because we do not do automatic grading for these questions:

![image](https://user-images.githubusercontent.com/9203229/41885672-031e04a6-78c7-11e8-93b6-ba2871796e62.png)

PR #1499, opened 2018-06-27
[#1463] Exam question Input with MC choices (connects #1463)

nirojdyola:


src/app/exams/exams-question.component.ts
```diff
@@ -64,4 +64,12 @@ export class ExamsQuestionComponent implements OnInit {
     return item.id;
   }
 
+  onChange(type) {
+    if (this.choices.length > 0) {
+      for (let i = 1 ; i <= this.choices.length ; i ++) {
+        this.removeChoice(0);
```
paulbert: Please test your PRs.  This only removes one choice.

src/app/exams/exams-question.component.ts
```diff
@@ -64,4 +64,12 @@ export class ExamsQuestionComponent implements OnInit {
     return item.id;
   }
 
+  onChange(type) {
+    if (this.choices.length > 0) {
```
paulbert: This if statement is unnecessary.  The for loop will never be entered if length is 0.

Issue #1201, opened 2018-05-31
Course progress on database


paulbert:

paulbert:
Fixed in #1209

PR #1500, opened 2018-06-27
Add course progress visuals for learner (connects #1201)

paulbert:


lmmrssa:
Whenever member does admission to course. Progressbar shows blue for the first step and grey for remaining even when there is not exam created for that step.
Looks like green is for completed, blue for incomplete and grey for unattempt. But color code is not working as expected.
sketch 56

paulbert:
Correct, grey is for steps not started, blue is in progress, and green is completed.

I made an update to make all steps grey until the user actually views one, and the final step will be green after either a view of the last step if there is no exam or an exam completion if there is.

PR #1501, opened 2018-06-28
[#1455] Load the same page on changing Language (fixes #1455)

### LINKED ISSUE #1455: Changing language should load same page

i5o:


src/app/shared/planet-language.component.ts
```diff
@@ -21,4 +22,12 @@ export class PlanetLanguageComponent implements OnInit {
       language => language.shortCode !== this.currentLanguage.shortCode);
   }
 
+  getRouterUrl(language) {
+    const newRoute = '/' + language.shortCode + this.router.url;
+    if (this.router.url.indexOf('/' + this.currentLanguage.shortCode + '/') === 0) {
```
lmmrssa: language code is not part of router.url as we are changing basehref. So removing this condition

i5o: perfect. I wasn't aware about the basehref changes

PR #1506, opened 2018-06-28
Try to fix the ng e2e test #1436

### LINKED ISSUE #1436: "ng e2e --environment test" is test nothing after upgrading the Angular version
See this https://travis-ci.org/open-learning-exchange/planet/jobs/395550934#L891

After upgrading the Angular version seems karma test doesn't work.

screen shot 2018-06-23 at 12 09 30 am

mappuji:
Try to fix #1436


e2e/login.e2e-spec.ts
```diff
@@ -22,7 +22,7 @@ describe('Login', () => {
     passInput.sendKeys('e2e');
     page.clickSignin();
     browser.getCurrentUrl().then((url) => {
-      expect(url).toEqual('http://0.0.0.0:49152/');
+      expect(url).toEqual('http://localhost:4200/');
```
mappuji: can url stored in config?

paulbert: When we set up more tests, yes this should be set in config.

paulbert:
Found a possible problem:  For the latest commits the first two `planet-test` builds are not being tagged with the branch name and commit id:

image

mappuji:
Thanks @paulbert will take a look at it.

PR #1514, opened 2018-06-28
1487 configuration service

### LINKED ISSUE #1487: ConfigurationComponent has too many functions
I think we can split this file into a ConfigurationService with the functions that handle data manipulation moved there to make this file more manageable.

ArashDai:
fixes #1487


src/app/configuration/configuration.service.ts
```diff
@@ -0,0 +1,200 @@
...
+        },
+        'url': 'https://' + configuration.parentDomain + '/feedback'
+      },
+      'create_target': false,
+      'continuous': true,
+      'owner': credentials.name
+    };
+  }
+
+  getFeedbackSyncDown(feedbackSyncUp, configuration) {
```
lmmrssa: we will have to do some rework as we do not want feedback sync in configuration now. But that changes has not been merged to master yet.

Issue #1267, opened 2018-06-07
Add spacing between `create account` fields


gcantieni:
### Problem
When two input fields are next to each other, if the input is long enough they bleed together. Though in many of these fields the inputs won't be long enough for this to happen, this isn't guaranteed. For example, I noticed the problem because my email and phone number bled together.

![text_bleeds_together](https://user-images.githubusercontent.com/34512397/41115605-02e21b14-6a56-11e8-8714-75aab40e623a.png)

### Proposed solution
It would be more aesthetically pleasing if there were a slight space between the text of each field.

paulbert:
This happened with the upgrade to Angular 6 (and version 6 of the @angular/material module).  I'm not sure how it was implemented before, but something like a `grid-column-gap` implementation would be ideal (only adds space between elements, not around the outside).

paulbert:
Fixed in #1653

PR #1526, opened 2018-07-02
[#1267] added spacing to configuration form (connects #1267)

singharpita:


paulbert:
Desired result:
image

PR #1529, opened 2018-07-02
[#1200] Add loading spinner after submission (fixes #1200)

### LINKED ISSUE #1200: Disable action buttons once event is triggered and show indicator

gcantieni:
Added:
* DialogsLoadingComponent which opens a dialog with a mat-spinner 
* SubmitDirective which opens a DialogsLoadingComponent upon a valid submit

SubmitDirective currently is declared in the language module, which in future commits will be transformed into the shared module.


gcantieni:
@lmmrssa can you point me on some resource on how to clean up my branch/how to make a clean PR? As is obvious I've had trouble doing so and I'm not quite clear where I went wrong--I thought I accumulated these commits after a `git pull`, but your comment "instead of rebasing or merging" implies that the pull wouldn't be the problem cuz it's just fetch+merge.

paulbert:
@gcantieni I was revisiting this and realized there was an easy fix for the `closeOnNavigation` limitation: close the dialog in `ngOnDestroy`.  When the route changes that instance of the directive is destroyed, so the popup is removed.

PR #1540, opened 2018-07-03
[#1536] Make question description required (fixes #1536)

### LINKED ISSUE #1536: Exam question: make description required
We shouldn't allow exam questions with no question at all.  We can require the description for each question and leave the header optional.

i5o:


lmmrssa:
@i5o The error message will not be visible as we have these questions under expandable. So may be we need to highlight the expandable itself with some color or give general message next to submit button that some input is has error so please fix it.

Issue #1546, opened 2018-07-03
Remove request course


lmmrssa:
As we have feature using which planet can pull any course available. We do not need this feature anymore.

lmmrssa:
We still need to clean RequestComponent

lmmrssa:
Close in reference to #1567

PR #1553, opened 2018-07-03
[#1546 remove request] (connects #1546)

singharpita:


lmmrssa:
@singharpita also remove requestcomponent and all its call

Issue #1544, opened 2018-07-03
Add feedback for course


lmmrssa:
We have feedback button on resource. We should make similar for course as well, to send feedback specific to course

PR #1555, opened 2018-07-03
[#1544] course feedback (connects #1544)

singharpita:


lmmrssa:
@singharpita it looks nice can we change the icon on feedback view as well based on the type of feedback. Right now its only showing library icon. If its course it should show course icon and for resource library icon

Issue #1539, opened 2018-07-03
course view layout issue


singharpita:
![course-view-layout](https://user-images.githubusercontent.com/27276132/42195009-7fb2066e-7e96-11e8-83ee-dc19455cc292.png)
![course-view-layout-1](https://user-images.githubusercontent.com/27276132/42195012-8526082a-7e96-11e8-8508-425aa565f74f.png)

paulbert:
The panels on the right should not expand beyond the height of the containing element.  We should make the `overflow-y: auto` (so a scrollbar appears) for that part of the view

PR #1557, opened 2018-07-03
[#1539] course view layout issue (connects #1539)

lucy811:
course_view


lmmrssa:
@lucy811 can we also use auto collapse of expandable when you open second expandable item of step.

lucy811:
I think the second expandable item of step is within the height of the containing element.
step-view

lmmrssa:
@paulbert please check the changes

Issue #1550, opened 2018-07-03
Removing meetup from dashboard


lmmrssa:
When we remove meetup from dashboard it does not update the list.
It is working for resource and course.
![sketch 68](https://user-images.githubusercontent.com/2909490/42196672-f30979f0-7e9e-11e8-9c56-2247d36e0ea4.png)

lmmrssa:
Issue also exist for myTeam

PR #1561, opened 2018-07-03
[#1550] Removing meetup from dashboard (connects #1550)

lucy811:
The error happens if we remove the last one item from the dashboard, it can’t  update the list. Because the last one item may appear in the myLibrary, myCourses, myMeetups or myTeams, so this error may happen in in the myLibrary, myCourses, myMeetups and myTeams.
dashboard


lmmrssa:
@paulbert please check changes

PR #1565, opened 2018-07-04
[#1342] Add il8n section to style guide (fixes #1342)

### LINKED ISSUE #1342: Update Style-Guide.md
Update style guide to not use `i18n` attribute on tags which do not have any String (i.e empty tag or only contain Interpolation String)

gcantieni:


Style-Guide.md
```diff
@@ -43,6 +43,9 @@ Please don't use the following name for variable since it is used by our `entryp
 | planet-db-host 	| used in the docker script 	|
 | planet-db-port 	| used in the docker script 	|
 
+### il8n
```
lmmrssa: @gcantieni can we add few more examples as people might not be familiar with interpolation or some other terms. Also i18n can be used for attributes such as `i18n-title` `i18n-placeholder`

Style-Guide.md
```diff
@@ -43,6 +43,9 @@ Please don't use the following name for variable since it is used by our `entryp
...
+Planet uses the `il8n` translation module to translate our content and make it accessible. This module can be included in an html tag with the keyword il8n. For example, `<p il8n> Text to be translated </p>`. However, if there are no strings enclosed by the tag, or if only interpolation strings are enclosed, then the `il8n` attribute should not be used as it will raise errors during compilation.
```
paulbert: Let's remove the starting & ending spaces in the HTML tag example.  Those are mistakes that have unfortunately snuck into many of our HTML files.

gcantieni: Ok sounds good!

paulbert:
Looks good, we can always add more later if we feel the need

PR #1599, opened 2018-07-10
 [#1524] upgrade planet using nginx and the ui (fixes #1524)

### LINKED ISSUE #1524: Software Update or hack portainer
https://portainer.io/install.html

i5o:


docker/planet/upgrade.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+echo "HTTP/1.0 200 OK"
+echo "Content-type: text/plain"
+echo ""
+curl --unix-socket /var/run/docker.sock -X POST "http://localhost/images/create?fromImage=treehouses/planet:db-init"
```
mappuji: Can we dry it perhaps? using function?

i5o: Sure. I will push this on wednesday

mappuji: cool!

docker/planet/upgrade.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
i5o: Bash is not installed on the docker image, thats why we are using sh

paulbert: Ok, will start a quick PR to install it.  `sh` doesn't support arrays, so throws an error.

PR #1605, opened 2018-07-10
[#1597] Subscribe to paramMap and check search [REQUIRES TESTING ON NATION] (fixes #1597)

### LINKED ISSUE #1597: Request list: Subscribe to param list for changes
In the community request list if you click a notification while looking at the list, the search filter will not update.

Rather than use `this.route.snapshot` on line 47 we should subscribe to the paramMap so when the parameter updates the view will change based on those changes.

See line 71 of `courses-step-view.component.ts` for an example implementation:
https://github.com/open-learning-exchange/planet/blob/master/src/app/courses/step-view-courses/courses-step-view.component.ts#L71

i5o:


paulbert:
Just going to do one final check before merging.

PR #1616, opened 2018-07-11
[#1267] spacing to configuration form (connects #1267)

singharpita:


paulbert:
Ok let's try just adding some css to `styles.scss` to add a little space between `mat-form-fields` that are children of `form` elements in the entire app.

We may need to add this as a class to be added to the `form` elements if this creates unwanted results in some of the forms.  That's more maintainable than adding a class to every `mat-form-field`.

lmmrssa:
#1653 fixes it so this PR can be closed

PR #1654, opened 2018-07-13
Survey functionality

paulbert:
The goal is to have adding, sending, taking, and receiving surveys in this PR.

Currently working on the adding part.


lmmrssa:
@paulbert still few things need improvement. But merging it so that we can do full test with large number of people. Initial test seems good to play with.

PR #1658, opened 2018-07-13
Add members to teams (fixes #1629)

### LINKED ISSUE #1629: In team, there should be invite member option as in meetups
invitemember

paulbert:


src/app/shared/dialogs/dialogs-list.service.ts
```diff
@@ -13,11 +14,26 @@ const listColumns = {
 @Injectable()
 export class DialogsListService {
 
+  defaultSelectors = {
```
lmmrssa: I'm not sure if we want defaultselector of user here. As this dialoglist is used for resources and courses as well.

PR #1663, opened 2018-07-13
1487 previous course progress

### LINKED ISSUE #1584: Once resign not able to check previous progress
When someone resign from the course previous progress made by user is not visible on **My Progress** section. It does show on course leader progress section though.
But as data is not lost once readmission show all old progress.

It will be nice to be able to see progress of all courses taken previously even if you are not currently enrolled.

ArashDai:
fixes #1584


lmmrssa:
Only problem I see right now is when user is new and there is not course_progress or assigned courses, then it is showing all courses on progress which will not be good as there will be lot of courses in the system.
sketch 80

PR #1664, opened 2018-07-13
[#1647] Added retry option when the docker timeouts (fixes #1647)

### LINKED ISSUE #1647: software upgrade - fix message when timeout
screenshot_20180712_120345

i5o:
screenshot from 2018-07-13 20-28-56
screenshot from 2018-07-13 20-29-18

we can change the colors

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ang-bell-app",
-  "version": "0.2.10",
+  "version": "0.2.11",
   "license": "AGPL-3.0",
   "scripts": {
     "ng": "ng",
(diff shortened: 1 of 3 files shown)


dogi:
awesome @i5o

lmmrssa:
@i5o can we have some number of limit to retries and if it does not pass then show Failed message. Otherwise it will keep retrying continuously.

i5o:
Sure @lmmrssa. I guess we can add 5, after that we ask the user to submit feedback.
When the page is reloaded that "5" will be lost, is that ok?

lmmrssa:
@i5o yes it is ok As after reload user will have to again click button to trigger it. We just need to be able to notify user

i5o:
image

Issue #1671, opened 2018-07-14
software upgrade - checking and comparing parent for version


dogi:
-> upgrade button
-> notification?

PR #1676, opened 2018-07-16
Show version of parent & planet (connects #1671)

paulbert:
Uses the new `/version` endpoint rather than the database.

### DIFF
--- a/couchdb-setup.sh
+++ b/couchdb-setup.sh
@@ -117,8 +117,6 @@ curl -X PUT $COUCHURL/teams
 
 # Create design documents
 node ./design/create-design-docs.js
-# Update version number in configurations database
-upsert_doc configurations version $(jq -c '{version: .version}' package.json)
 # Add or update design docs
 upsert_doc nations _design/nation-validators @./design/nations/nation-validators.json
 # Insert indexes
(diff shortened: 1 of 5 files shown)


paulbert:
I'd like to merge this so we at least don't have to manually add a configuration document to be able to login, but this doesn't fully fix the issue.

It's not working to show the version in the dashboard, and the GET request is failing to the nation and returning this error:
```
Failed to load https://paulbert.media.mit.edu/version: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://localhost:3000' is therefore not allowed access. The response had HTTP status code 405.
```
Because the development environment won't have a local `/version` location, I can't debug why the local version isn't showing up in the docker container, either.  In any case I just wanted to fix up the changes we made to the configuration database and then pass this off.

paulbert:
Figured out what was going wrong: the headers were incorrect for the request.  This appears to work now.

paulbert:
After testing with the `dev` and `paulbert` nations, it looks like the changes I made were not enough and we will need to add the Access-Control-Allow-Origin header to the server.

@mappuji @dogi @lmmrssa Tagging so we can discuss at systems meeting.

paulbert:
@mappuji We're not using the database anymore for the version number.  This PR removes changes that would involve `db-init` or CouchDB for storing & accessing the version.  Basically the DB would be easier to accidentally change, so with this we just expose a file to read externally with the version as text (see #1666 for that change).  Problem is we need to read that file with cross origins.

dogi:
@paulbert how would I test this?

paulbert:
@dogi Yes look at the manager view to test.  If you create a community on the "Paul" nation you will see both nation & community versions, if not you'll just see community.

paulbert:
@dogi I'm going to merge this & we can test tomorrow.

Issue #1640, opened 2018-07-12
Dashboard: X is overlapping with text


paulbert:
This is especially apparent on smaller screens.  Ideally we should have the text remain centered in the box, but it avoids the space with the X icon.  See examples:

![image](https://user-images.githubusercontent.com/9203229/42638605-1634ba42-85bc-11e8-84d1-0a6c40eed8da.png)

and

![image](https://user-images.githubusercontent.com/9203229/42638546-eccff540-85bb-11e8-86f9-d8c8f7a2c26b.png)

PR #1705, opened 2018-07-19
[#1640] dashboard overlapping text (connects #1640)

singharpita:


lmmrssa:
Looks good with the margin but found that whole tile gets affected by height of one item which is happening on master branch as well. So could you also fix it along with this PR.
sketch 79

paulbert:
Added some spacing for readability and changed how the height is set to preserve the vertical centering of the text.

After looking at this, I think there isn't a pure CSS solution.  The text needs to be vertically centered when it doesn't overflow, but aligned to top when it does.

lmmrssa:
Text is still overlapped by `remove` icon
screen shot 2018-07-31 at 14 45 55

lmmrssa:
@dogi @paulbert changes made will add margin to top and center text based on remaining height. But it will still only show center part of the title. Couldn't find conditional styling that can be implemented based on content height.

lmmrssa:
@paulbert reverted back the changes

paulbert:
Closing.  This doesn't really improve upon the issue.

Issue #1622, opened 2018-07-12
Send notification to all team members on request


paulbert:
When another user requests to join a team, we should send a notification to all team members that a user is requesting membership to their team.

Issue #1687, opened 2018-07-17
Team : Notification to member


singharpita:
if in team new member is added, that user should get notification.

PR #1707, opened 2018-07-19
[#1622] send notification to all team (connects #1622)

singharpita:


lmmrssa:
@singharpita this PR is currently only fixing #1687 could you also add fixes for #1622

Issue #1711, opened 2018-07-19
Team view: font color highlight of Request Pending


singharpita:
![request_pending](https://user-images.githubusercontent.com/27276132/42948580-8186b844-8b8f-11e8-89e5-af3a4dd8801c.png)

paulbert:
We could make that button yellow with `color="accent"` and `mat-raised-button` instead of `mat-button`

PR #1718, opened 2018-07-20
[#1711] team view request pending highlight (connects #1711)

singharpita:
also removed the "disabled" attribute as it affects the color and style of button.


lmmrssa:
@paulbert Instead of button then we can just have a text so it uses white color instead of grey.

PR #1720, opened 2018-07-21
Resource tags

paulbert:
Implementing a system for tagging/labeling resources


lmmrssa:
@paulbert Is this still in progress. I couldn't test the feature. Or did i not get the clear picture of it. It is not saving tags in resources.

i5o:
Looks good, but I would probably limit the tags to 5
image

or maybe add some "..." button after 5+ tags

i5o:
I would also use "," as a separator for tags

i5o:
also clicking a tag could search for it

paulbert:
Updated with the changes that we talked about (well, we talked about making issues for...)

@lmmrssa @i5o @dogi If this still looks good we should merge so we can do more testing Thursday.

Issue #1716, opened 2018-07-19
Database access on subdomain rather than different port for production


paulbert:
We'd like to change the production version of the application to connect to the database on a different subdomain (like `db.earth.ole.org`) rather than a different port (or in addition to).  Note this will also require changes to srv.

Hardcoding the port to 2200, for example, causes problems with SSH tunneling using a different port.  The app still tries to access on port 2200, but from the client the database is only available on another port.

We should discuss tomorrow.

i5o:
```
        client_max_body_size 1g;
```
needs to be increased, the default is 1m

paulbert:
Closing, this is working well

PR #1724, opened 2018-07-24
[#1716] Pass database to  http://host/di/ 

i5o:


i5o:
Thanks for testing @lmmrssa, we still need to make "_utils" works.
It basically involves playing with nginx rules, which I guess we are doing on friday

lmmrssa:
@i5o @paulbert  @dogi  we are able to access fauxton with different port tunnel to rpi
Please check pirate.sh:22682 for fauxton and pirate.sh:22680 for planet.
So I think we can merge this PR.

i5o:
Are we going to merge it @lmmrssa ?

Perhaps changing /di to /db before?

docker/planet/default.conf
```diff
@@ -57,7 +57,7 @@ server {
     add_header Content-Type text/plain;
   }
 
-  location /di/ {
+  location /db/ {
```
i5o: :cry: :disappointed:

paulbert:
Almost merged this before realizing that the changes to `environment.ts` would break our development environment.  Only needed to change the value for production where we have docker/nginx.

Issue #1731, opened 2018-07-24
Notify user that they are not approved


lmmrssa:
Make always visible banner that show user that their status is unapproved and administrator need to approve to use full feature.

lmmrssa:
As per last talk with @paulbert we are not showing warning always instead we will make non collapsing warning message when we navigate to disabled pages.

PR #1738, opened 2018-07-25
[#1731] notify not approved user (connects #1731)

singharpita:


src/app/home/home.component.html
```diff
@@ -3,7 +3,7 @@
   <button mat-icon-button (click)="toggleNav()" i18n-title title="Menu"><mat-icon>menu</mat-icon></button>
   <img src="assets/cropped-ole-ico-logo-32x32.png" class="ole-logo"/>
   <h1 i18n>Planet Learning</h1>
-  <span class="toolbar-fill"></span>
+  <span class="toolbar-fill warn-text-color warn-text-align" *ngIf="userservice.roles.length === 0 && !userservice.isUserAdmin">You are not authorized. Please contact administrator.</span>
```
paulbert: The `<span class="toolbar-fill"></span>` is a special element which is very important for the layout of our toolbars making the elements before it aligned to the left, and the elements after aligned to the right.

Modifying it/removing it (the *ngIf will remove it when the user has access) will break the layout:
image

Please add the message as part of a different element.

Issue #1527, opened 2018-07-02
Refactor DialogsFormComponent & DialogsFormService


paulbert:
DialogsFormComponent should not close from the component itself because it will not & should not have knowledge of the state of the form to know if it is valid.

gcantieni:
### Plan:
- [ ] replace `confirm` function in DialogsFormService with `openDialog` and `closeDialog` functions
- [ ] add callback function in DialogFormService to be injected it into the DialogsFormComponent
- [ ] the callback function will make a request that we can subscribe to and error check, and only upon a successful request will we call `closeDialog`.

paulbert:
To put them in one place, here's a list of all of the related issues that have come up since creating this one:

#3033 #2844 #2795 #2541 #2540 #2053

PR #1739, opened 2018-07-25
Refactor dialogs forms (connects #1527)

paulbert:
Adds a function which takes a callback to DialogsFormService and implements on FeedbackDirective


gcantieni:
Ugh, so ironically I was trying to clean this branch through rebasing and it added all of the commits from master from the past few days. Also, last time I tried to rebase this happened as--I'd appreciate if someone could tell me how to fix this so I don't make the mistake in the future.

gcantieni:
Ok nvm I cleaned it up

gcantieni:
Upon refactoring the ChangePasswordDirective, it has become apparent that when requesting to change the password it doesn't matter if you input the correct old password (the password required to update the user's password). In the console we get:
`ERROR >>>  Http post 14 request HttpErrorResponse {headers: HttpHeaders, status: 401, statusText: "Unauthorized", url: "http://localhost:2200/_session", ok: false, …}`
and yet the password is still updated successfully. I think this can be recreated in master... should I open it as a new issue or incorporate it as part of this refactoring--it might be as simple as a single check perhaps?

paulbert:
Closing due to inactivity

Issue #1747, opened 2018-07-26
Users list routing


paulbert:
Now that we have created the teams view and moved links to the users list to be on the teams & manager dashboard pages we should update the routing.  Two changes are needed.

1. The routes for the users page should be a level above the page they are linked from (i.e. `manager/users` and `teams/users`.  The notification sent to the admin when a user is created should link to `manager/users`.
2. The back button should change to go up one level of the routing structure to the page it was linked from using [relativeTo](https://angular.io/api/router/NavigationExtras#relativeTo) for the `route.navigate`

lmmrssa:
@paulbert shall we completely replace users route with /teams/users in case of opening users profile from top navbar profile image as well

PR #1748, opened 2018-07-27
[#1747] user list routing (connects #1747)

singharpita:
users's profile has direct url so `/users` route has not removed from homerouter.

### DIFF
--- a/src/app/users/users.component.ts
+++ b/src/app/users/users.component.ts
@@ -255,7 +255,7 @@ export class UsersComponent implements OnInit, OnDestroy, AfterViewInit {
   }
 
   back() {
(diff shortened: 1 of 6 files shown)


src/app/users/users.component.ts
```diff
@@ -255,7 +255,7 @@ export class UsersComponent implements OnInit, OnDestroy, AfterViewInit {
   }
 
   back() {
-    this.router.navigate([ '/' ]);
+    this.router.navigate([ '../../' ], { relativeTo: this.route });
```
paulbert: Should only go up one level.

singharpita: One level does not change url for `users` page that why needed to add two level

Issue #1725, opened 2018-07-24
Manage Sync: only delete completed replicators


paulbert:
When we push Run Sync on the below route all replicators are deleted before creating the sync replicators.

We should only delete the completed replicators because if one is retrying due to a connectivity issue or in the process of running we do not want to delete it before it completes.

![image](https://user-images.githubusercontent.com/9203229/43112804-b6bf59ec-8ec5-11e8-94a6-e38b28e35be2.png)

PR #1761, opened 2018-07-30
[#1725] Clean only completed replicator (connects #1725)

lmmrssa:


src/app/manager-dashboard/manager-sync.component.ts
```diff
@@ -44,21 +69,8 @@ export class ManagerSyncComponent implements OnInit {
...
-      { db: 'resource_activities' }
-    ];
-    // List of replicators to pull from parent
-    const pullList = [
-      { db: 'feedback', selector: { source: this.userService.getConfig().code } },
-      { db: 'notifications', selector: { target: this.userService.getConfig().code } }
-    ];
     const addType = (type) => (val) => ({ ...val, type });
-    return pushList.map(addType('push')).concat(pullList.map(addType('pull')));
+    return this.pushList.map(addType('push')).concat(this.pullList.map(addType('pull')));
```
paulbert: You should refactor to use `getDefaultList` rather than duplicate this line.  `getDefaultList` should use the addType function on line 72.  You can still keep these in the `deleteArray` list by using `defaultList.findIndex(...)` rather than `indexOf`

Issue #1765, opened 2018-07-31
Broken back button


ArashDai:
In the library, when adding a new resource, if you press the back button it breaks needing a hard refresh.
ERROR Error: Uncaught (in promise): TypeError: Cannot read property 'close' of undefined

![12341234](https://user-images.githubusercontent.com/8280071/43434406-55a5a704-9430-11e8-9ab6-7c08b3ccdd42.png)

PR #1778, opened 2018-08-01
Checking the status of SpinnerDialog before closing (#1743 , #1765)

### LINKED ISSUE #1743: SpinnerDirective: Need to check if open before running close()
The spinnerDialog in SubmitDirective is closed on destroy, but if it was not open this will throw an error and crash the app.

We need to only close the dialog if it is open.

BhargaviNadendla:
fixes #1743 #1765


lmmrssa:
@BhargaviNadendla there was codeclimate error, if possible try to fix those as well.

Issue #1813, opened 2018-08-07
create environment for parentprotocol


lmmrssa:
Add environment variables to setup docker to have http or https parentProtocol similar to what we had for centerProtocol

PR #1817, opened 2018-08-08
[#1813] Protocol from yml file (connects #1813)

lmmrssa:
This will enable to setup community to connect to nation with `http` or `https`
**PARENT_PROTOCOL** can be used on yml file. If not declared will use `https` by default

### DIFF
--- a/docker/planet/scripts/docker-entrypoint.sh
+++ b/docker/planet/scripts/docker-entrypoint.sh
@@ -1,25 +1,23 @@
 #!/usr/bin/env sh
 
-if [ -z "$HOST_PROTOCOL" ]
-then
-  PROTOCOL="http"
-elif [ "$HOST_PROTOCOL" = "https" ]
-then
-  PROTOCOL="https"
-else
-  PROTOCOL="http"
-fi
+# Function for protocol
+check_protocol() {
+  PARAM=$1
+  DEFAULT=$2
+  if [ -z "$PARAM" ]
+  then
+    PROTO="$DEFAULT"
+  elif [ "$PARAM" = "https" ]
+  then
+    PROTO="https"
+  else
+    PROTO="http"
+  fi
+  echo "$PROTO"
+}
 
-if [ -z "$PARENT_PROTOCOL" ]
-then
-  PPROTOCOL="https"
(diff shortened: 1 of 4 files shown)


src/environments/environment.test.ts
```diff
@@ -4,6 +4,6 @@ export const environment = {
   couchAddress: 'http://127.0.0.1:5984/',
   centerAddress: 'earth.ole.org:2200',
   centerProtocol: 'https',
-  parentProtocol: 'https',
+  parentProtocol: 'planet-parent-protocol',
```
paulbert: @lmmrssa This is not overwritten by the docker script.  That only runs for the production version.  Please do not change as this will break tests.

src/environments/environment.ts
```diff
@@ -9,6 +9,6 @@ export const environment = {
   couchAddress: window.location.protocol + '//' + window.location.hostname + ':2200/',
   centerAddress: 'earth.ole.org:2200',
   centerProtocol: 'https',
-  parentProtocol: 'https',
+  parentProtocol: 'planet-parent-protocol',
```
paulbert: @lmmrssa This is not overwritten by the docker script.  That only runs for the production version.  Please do not change as this will break our development environment.

docker/planet/scripts/docker-entrypoint.sh
```diff
@@ -10,6 +10,17 @@ else
...
+then
+  PPROTOCOL="https"
+elif [ "$PARENT_PROTOCOL" = "http" ]
+then
+  PPROTOCOL="http"
+else
+  PPROTOCOL="https"
+fi
+
+
```
paulbert: Extra line break

docker/planet/scripts/docker-entrypoint.sh
```diff
@@ -10,6 +10,17 @@ else
   PROTOCOL="http"
 fi
 
+if [ -z "$PARENT_PROTOCOL" ]
```
paulbert: And maybe we should write this as a function to combine with the above block and be a little more DRY

src/environments/environment.test.ts
lmmrssa: thought so :) changes reverted

PR #1819, opened 2018-08-09
Change course add steps to custom panel like list

paulbert:
Using the mat-expansion-panel for the course steps makes it difficult if you needed to scroll to see the add resource/exam buttons.  For courses with ~50 steps it would be very cumbersome trying to scroll to find the one step you have open to edit.

Added a new component which we can reuse for similar lists which opens a form when an item is clicked on to edit that item's attributes.

### DIFF
--- a/src/app/courses/add-courses/courses-add.component.html
+++ b/src/app/courses/add-courses/courses-add.component.html
@@ -39,14 +39,7 @@
       </form>
     </div>
     <div>
-      <planet-courses-step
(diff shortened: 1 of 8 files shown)


lmmrssa:
@paulbert navigation seems to work but resource add does not show up once you return back. Which could be confusing.

lmmrssa:
@paulbert looks good. Resource is adding and also exam is working.
Still there is minor issue of added resource not showing on runtime. Will create new issue for that

PR #1857, opened 2018-08-15
Deselecting the resources in course-step (fixes #1847)

### LINKED ISSUE #1847: CourseStep: remove all resources
For a course step we should be able to deselect all resources and hit OK.  There is a property on DialogsList that can enable this.

BhargaviNadendla:


src/app/shared/dialogs/dialogs-list.component.ts
```diff
@@ -45,7 +45,7 @@ export class DialogsListComponent implements AfterViewInit {
     this.tableData.data = this.data.tableData;
     this.tableColumns = this.data.columns;
     this.disableRowClick = this.data.disableSelection || false;
-    this.emptySubmit = this.data.selectionOptional || false;
+    this.emptySubmit = this.data.selectionOptional || true;
```
lmmrssa: @BhargaviNadendla you should not change it here. It is default value if you do not pass it during call from other component. You should pass `selectionOptional` form course-step during call for dialoglist

Issue #1858, opened 2018-08-15
Request entity too large uploading resource 1mb+


paulbert:
When the attachment is >1mb on a local docker instance (RPi or in vagrant) I'm seeing the below error.
![image](https://user-images.githubusercontent.com/9203229/44127324-d162c706-a00a-11e8-9a5a-16bc6c158f6e.png)

PR #1859, opened 2018-08-15
Resource attachment fix (connects #1858)

paulbert:
Added `client_max_body_size` to `location /db/`.  Not sure if this will fix the problem, but it was quick to add to get a build started.


i5o:
When we were testing the first time we had to do the same

https://github.com/open-learning-exchange/planet/issues/1716#issuecomment-406766856

We used 1gb not 1024mb, not sure which one is better (both are the same, but meh)

PR #1891, opened 2018-08-17
Fix spinner not closing on login error (fixes #1804)

### LINKED ISSUE #1804: Spinner on login does not close with incorrect password
The spinner is set to close when the app navigates to a new route.  If an error occurs with the form being sent (incorrect password, HTTP request fails) the app will not navigate to a new route in many cases meaning the spinner dialog will never close.

Let's instead use a service to control when the dialog closes.

paulbert:


gcantieni:
Great, so it looks like we don't need a service after all!

PR #1895, opened 2018-08-18
[1180] planet code for every doc (connects #1180)

yuviii:


lmmrssa:
@yuviii could you fix your global config on machine.

PR #1925, opened 2018-08-21
fix multiple choice width for survey (fixes #1869)

### LINKED ISSUE #1869: Fix multiple choice width for survey
Width of choices for every other survey choice is too small.  It's using the same layout as exams, so these are the space for checkboxes (which are not needed for surveys).

image

leahphuong:
fixes #1869 

screen shot 2018-08-21 at 2 45 40 pm

### DIFF
--- a/src/app/exams/exams-question.component.html
+++ b/src/app/exams/exams-question.component.html
@@ -33,7 +33,8 @@
       </mat-form-field>
       <button type="button" mat-button (click)="addChoice()" class="margin-lr-5" *ngIf="questionForm.controls.type.value === 'select'">Add Choice</button>
     </div>
-    <div *ngIf="examType !== 'surveys'" class="exam-question-choices">
+    <div class="question-choices"
+      [ngClass]="{'survey-question': examType === 'surveys'}">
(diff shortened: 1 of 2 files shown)


src/app/exams/exams-question.component.html
```diff
@@ -33,19 +33,26 @@
       </mat-form-field>
       <button type="button" mat-button (click)="addChoice()" class="margin-lr-5" *ngIf="questionForm.controls.type.value === 'select'">Add Choice</button>
     </div>
-    <div class="question-choices">
+    <div *ngIf="examType !== 'surveys'" class="exam-question-choices">
```
lmmrssa: @leahphuong you can just use conditional class name instead of having *ngIf and repeated code block

src/app/exams/exams-question.component.ts
```diff
@@ -10,7 +10,13 @@ import { uniqueId } from '../shared/utils';
   selector: 'planet-exam-question',
   templateUrl: 'exams-question.component.html',
   styles: [ `
-    .question-choices {
+    .survey-question-choices {
```
lmmrssa: you can leave the class as it is and just add another class which will just overwrite grid-template-columns

Issue #1916, opened 2018-08-21
Route parameters persist from users notification


paulbert:
The route parameter to set the search filter after clicking on a user notification persists so if you remove the filter, look at a user's profile, then hit the back button the filter returns.  This isn't a great user experience.

https://www.useloom.com/share/b4f2d805b562453e844379a873f72a1f

A broader thought would be to set a parameter to match the filter as it is when you leave the page.  That way while the user is navigating they will still see the list as it was previously filtered when they hit the back button.

PR #1927, opened 2018-08-22
[1916] route parameters from user notification(connects #1916)

yuviii:
Route parameters cannot be changed hence queryparams is used.


paulbert:
RouteParams can be updated with `this.router.navigate([ '..', {params})`

yuviii:
@paulbert Route param pr fixed. Please review the changes.

PR #1929, opened 2018-08-23
Save admin credentials for upgrade script

paulbert:
First pass at a fix for this.  A few comments

1. I left it so the dialog to confirm the administrator's password comes up every time.  This is the simplest way to make sure that if communities change administrators or the administrator changes his/her password the upgrade doesn't silently fail to modify the CouchDB in the background.
2. I put the new yml file in /etc/planet, but if there's a more appropriate place for this then we should move it there.
3. The new compose command after the reboot would be:
```
docker-compose -f planet.yml -f /etc/planet/admin.yml -p planet up -d
```
4. As I'm writing this I realize that I think it will break if the password has a comma in it.

@dogi @i5o Feel free to take over from here and add commits directly to this branch.


paulbert:
To test:
1. Run `./docker/generate_yml.sh` from this branch
  - Alternatively `wget https://raw.githubusercontent.com/open-learning-exchange/planet/update-credentials/docker/planet.yml` & below install.yml
2. `docker-compose -f planet.yml -f install.yml -p planet up -d`
3. Log into the app & run the manager upgrade (submitting your password now)
4. `docker-compose -f planet.yml -f /etc/planet/admin.yml -p planet up -d`
  - On a RasPi you could change the autorun script and simply reboot

Install.yml
```
services:
  db-init:
    image: treehouses/planet-tags:db-init-0.3.6-update-credentials-6ac0c746
  planet:
    image: treehouses/planet-tags:0.3.6-update-credentials-6ac0c746
version: "2"
```

paulbert:
One possible issue: If the autorun compose file has `/etc/planet/admin.yml` and for whatever reason the file does not exist the compose will fail and planet will not start.  I think we should check if the file exists first.

Issue #1872, opened 2018-08-16
New user error: Password does not match


paulbert:
Spinner and the user/name password does not match pop up and are stuck on the screen.

On reload the user is logged in, so the user is created.

I overwrote my screenshot with another error, so we'll have to try to replicate this error.

paulbert:
I'm going to close this because we haven't been able to figure out the cause and it has not come up again

PR #1932, opened 2018-08-23
[#1872] new user error update (connects #1872)

singharpita:


paulbert:
I'm going to close this because we haven't been able to figure out the cause of the issue #1872 and it has not come up again

Issue #1920, opened 2018-08-21
Mime type error


lmmrssa:
Looks like service worker is keeping cache of file which is not valid anymore which is causing issue on application loading.
![sketch 94](https://user-images.githubusercontent.com/2909490/44379053-e0dfc580-a523-11e8-865c-63a137b4e5a2.png)

PR #1942, opened 2018-08-23
Fixing MIME type errors (connects #1920)

paulbert:


lmmrssa:
it looks good on one of machine but i have problem on another machine which is not gone even on this branch. So will create separate issue to debug that problem.

Issue #1936, opened 2018-08-23
Still some places showing exam for survey


lmmrssa:
![sketch 95](https://user-images.githubusercontent.com/2909490/44532783-adfc2400-a713-11e8-89be-eaa76f6ea725.png)

paulbert:
Closing, fixed in #2012

PR #1943, opened 2018-08-23
Survey title change according to examType(#1936)

jiajunc:
Placeholder on survey-add page can be changed according to examType variable.


paulbert:
Whoops.  Closing as this was fixed in #2012

PR #2014, opened 2018-09-02
[1980] file-size-limit-512MB-in-resource-add(connect #1980)

yuviii:
@lmmrssa @paulbert   File upload is limit to 512 MB in resource addition.

### DIFF
--- a/src/app/resources/resources-add.component.html
+++ b/src/app/resources/resources-add.component.html
@@ -86,7 +86,8 @@
       </mat-form-field>
       <div class="inner-gaps by-column full-width">
         <label i18n>File Upload:</label>
-        <input type="file" (change)="bindFile($event)" />
+        <input type="file" (change)="checkFileSize($event)" />
+        <label i18 class="warn-text-color" *ngIf="fileSizeExceeds">File size cannot exceed more than 512 MB</label>
       </div>
       <p *ngIf="existingResource._attachments"><b i18n>Existing Resource/file:</b> {{resourceFilename}}</p>
       <div class="inner-gaps full-width">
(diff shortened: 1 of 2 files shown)


src/app/resources/resources-add.component.html
```diff
@@ -86,7 +86,8 @@
       </mat-form-field>
       <div class="inner-gaps by-column full-width">
         <label i18n>File Upload:</label>
-        <input type="file" (change)="bindFile($event)" />
+        <input type="file" (change)="checkFileSize($event)" />
+        <label i18 class="warn-text-color" *ngIf="fileSizeExceeds">File size cannot exceed more than 512 MB</label>
```
paulbert: Should be `i18n` instead of `i18`

src/app/resources/resources-add.component.ts
```diff
@@ -233,8 +234,17 @@ export class ResourcesAddComponent implements OnInit {
...
-  bindFile(event) {
+  checkFileSize(event) {
     this.file = event.target.files[0];
+    this.fileSizeExceeds = this.file.size / 1024 / 1024 > 512;
+    if (!this.fileSizeExceeds) {
+      this.bindFile();
+    }
+  }
+
+  bindFile() {
```
paulbert: We shouldn't have a one line function that is only used once.

src/app/resources/resources-add.component.ts
```diff
@@ -117,7 +118,7 @@ export class ResourcesAddComponent implements OnInit {
   }
 
   onSubmit() {
-    if (this.resourceForm.valid) {
+    if (this.resourceForm.valid && !this.fileSizeExceeds) {
```
paulbert: We should set form invalid when we check file size rather than adding this other variable on the class.

yuviii:
@paulbert how can we add the validator in the file?

Issue #2001, opened 2018-08-30
Remove or disable credit of course


lmmrssa:
As we do not have credit action working on course right now. We should disable it.

lmmrssa:
![screen shot 2018-08-30 at 20 51 52](https://user-images.githubusercontent.com/2909490/44860695-92a29300-ac96-11e8-8d2a-a34beb1d63b7.png)

PR #2015, opened 2018-09-03
[2001] credit-disabled-in-courses(connects #2001)

yuviii:
@lmmrssa Please review the fix.


paulbert:
@lmmrssa I think we should remove it entirely rather than disable it since it may be a while before we implement any functionality for that.

yuviii:
@paul please review the changes.

PR #2019, opened 2018-09-03
Only select items on the view (fixes #1976)

### LINKED ISSUE #1976: Select all for tables: only select items viewed
In our tables clicking the select all button selects every resource/course/etc in the entire database, not just the ones viewed.  This leads to easy confusion & could result in deleting every resource from a community.

Right now we have a partial solution to warn when selected all before deleting.  We should better safeguard resources by only selecting the ones in view.

BhargaviNadendla:


singharpita:
meetup showing some weird behavior  on selectAll
screenshot from 2018-09-05 15-59-26
screenshot from 2018-09-05 15-59-41

BhargaviNadendla:
@singharpita It is woring fine at my end, I am unable to figure out since the same functionality is being added to meetups as courses and resources and there is no change with the code.

paulbert:
@BhargaviNadendla The problem is with the `isAllSelected` function.  You are setting this to true if the selected number equals the page size OR the remainder of total items / page size.  There are two issues depending on the size of `leftOverRows`:

- **> 0**: On every page except the last one `isAllSelected` will be true when the number of selected items equals `leftOverRows`, turning the master select into a full checkmark and deselecting all if clicked.
- **=== 0**: When there are no items selected, the master toggle will change from checked to unchecked when clicked, but no items will be selected (`isAllSelected` is true, so it actually clears the selection).

paulbert:
@singharpita I applied a fix.  Please review and @lmmrssa merge if approved.

paulbert:
@yuviii Please do not add merge commits if there are no conflicts with the master or updates on the master that the branch requires.

PR #2033, opened 2018-09-04
switch user to member

### LINKED ISSUE #1956: Switch "Users" to "Members"
A couple of places where we should switch "User(s)" to "Member(s)"

/login/newuser
1. Change route to `/login/newmember`
2. Change button text to "Become a member"
image

/manager
Change "Users" to "Members" on button
image

leahphuong:
fixes #1956


singharpita:
can't the user page Heading should also be updated?
screenshot from 2018-09-05 15-53-50

PR #2035, opened 2018-09-04
Parent resources: Speed up and fix layout

paulbert:


src/app/shared/couchdb.service.ts
```diff
@@ -79,7 +79,7 @@ export class CouchService {
 
   localComparison(db: string, parentDocs: any[]) {
     const ids = parentDocs.map((parentDoc: any) => parentDoc._id);
```
lmmrssa: this code is not used anymore

PR #2038, opened 2018-09-05
'Select all' only selects the users on view(fixes #2026)

### LINKED ISSUE #2026: Selecting only the users on view
In the users, clicking the select all button selects all of the users to add roles to them. It would be better if we select only the users who are on the view by 'select all'. This resembles #1976

BhargaviNadendla:


paulbert:
Closing this and applying changes to #2019

Issue #2044, opened 2018-09-06
Submission- submitted by shows "undefined undefined"


singharpita:
![submission-submttedby](https://user-images.githubusercontent.com/27276132/45160068-34caf980-b208-11e8-924e-41f5df98c8ab.png)

PR #2056, opened 2018-09-07
[#2044] submission submittedBy undefined update (connects #2044)

singharpita:


src/app/submissions/submissions.component.ts
```diff
@@ -50,7 +50,8 @@ export class SubmissionsComponent implements OnInit, AfterViewInit, OnDestroy {
     this.submissionsService.submissionsUpdated$.pipe(takeUntil(this.onDestroy$))
     .subscribe((submissions) => {
       this.submissions.data = submissions.map(submission => ({
-        ...submission, submittedBy: submission.user.name || (submission.user.firstName + ' ' + submission.user.lastName).trim()
+        ...submission, submittedBy: submission.user.name ?
+        submission.user.name || (submission.user.firstName + ' ' + submission.user.lastName).trim() : ''
```
paulbert: @singharpita This will never result in `(submission.user.firstName + ' ' + submission.user.lastName).trim()`.  If `submission.user.name` is truthy/has a value, it will always end up as `submission.user.name`.  If it is not it will be an empty string.

We need a solution which uses the `name` property of the user if it exists, the `firstName` + `lastName` if it does not, and "Unknown" if it neither exist.

lmmrssa:
@singharpita could you add the case of undefined undefined. Current implementation of checking `user.name` will not return `user.firstName and user.lastName` as user.name is not always available.
May be checking `user` might work but need to check the case of error happening.

src/app/submissions/submissions.component.ts
lmmrssa: oops @paulbert both of us commented :D

paulbert:
I just made the change.  Checks each of `firstName` and `lastName` for truthy values, so it won't return `undefined undefined`

Issue #2049, opened 2018-09-06
Step shouldnot be created empty


BhargaviNadendla:
The step should not be added if the description is left empty

![image](https://user-images.githubusercontent.com/25775298/45164474-21903d80-b1b8-11e8-8bdb-51b8e31ccb72.png)

paulbert:
To give course creators the flexibility to just have a video or PDF as the content for a step, we should allow blank description for course steps

PR #2057, opened 2018-09-08
Step should not be created empty(connects #2049)

lucy811:
Course step description is empty
disable

Course step description is not empty
enable


lmmrssa:
@lucy811 you also change permission of _auth.json please make sure you use `git add <filename>` instead of `git add .`

lucy811:
@lmmrssa git revert the second commit, then git add courses-step.component.ts again.

paulbert:
To give course creators the flexibility to just have a video or PDF as the content for a step, we should allow blank description for course steps

Issue #1766, opened 2018-07-31
action link on header should be made button


singharpita:
can be applied to entire application.
![action button](https://user-images.githubusercontent.com/27276132/43435647-a19d3c1a-94a0-11e8-8e21-1757e199688c.png)

lmmrssa:
Yes we can make is something like raised button will have to check if we want to keep it for now or update.

paulbert:
Let's see how `mat-raised-button` looks while keeping it the primary color.  We should only use the accent color sparingly.

lmmrssa:
@jiajunc we want to see how it will look

jiajunc:
@leomaxi Hey, does it mean that I should make the "Edit Profile" link into a button, and the form itself should be prefilled with user's original info?

paulbert:
Closing since we mainly (if not completely) use raised or stroked buttons.  If we find another which is not we can open a new issue.

PR #2058, opened 2018-09-08
Two links in header changed to raised button(connects #1766)

jiajunc:
Both "Edit Profile" and "Change Password" links are raised button with primary color now.


singharpita:
user-profile
right now its look bit confusing since header and button color are same.
Can we use light color for the button if we can't use accent or define light shade of blue for button ?

lmmrssa:
@jiajunc looks like you create another PR #2062 so closing this one

Issue #2006, opened 2018-08-31
In Exam submission multichoice, multianswer doesn't submit


singharpita:
Even the answer is correct.
![screenshot from 2018-08-31 12-12-09](https://user-images.githubusercontent.com/27276132/44896251-464b6780-ad17-11e8-8245-df5dbae9996e.png)

paulbert:
This was only implemented for surveys, so the initial issue #1966 was left open and should be closed when this is fixed.

PR #2060, opened 2018-09-09
[#2006] multianswer correct submission (connects #2006)

singharpita:


paulbert:
I tested this out on a question with three choices.  Through clicking around the choices I was able to submit an incorrect answer.  Looking on the database the value for that answer was an empty array, so even though the checkbox was checked in the UI the selection was not part of the answer value array.

Issue #2069, opened 2018-09-11
record survey : in case of multiple choice multi select question padding should be added at bottom


singharpita:
in between submit and checkbox
![screenshot from 2018-09-11 08-04-24](https://user-images.githubusercontent.com/27276132/45334211-88876b00-b599-11e8-926f-8743ad36bd8d.png)
it should look similar to this
![screenshot from 2018-09-11 08-04-37](https://user-images.githubusercontent.com/27276132/45334234-a05eef00-b599-11e8-9f25-1df28952e870.png)

PR #2074, opened 2018-09-11
Record survey : in case of multiple choice multi select question padding should be added at bottom(connects #2069)

lucy811:
padding


lucy811:
@lmmrssa 
multiple answer

Issue #2066, opened 2018-09-11
paginator in notifications


singharpita:
![screenshot from 2018-09-11 07-26-53](https://user-images.githubusercontent.com/27276132/45333175-7ce57580-b594-11e8-8ac3-02dbc88242f8.png)

paulbert:
Note that currently this is implemented as a `mat-list`, not a `mat-table`.  To add a paginator I believe this needs to be a `mat-table`.

PR #2075, opened 2018-09-12
[#2066] paginator added (connects #2066)

singharpita:


singharpita:
@paulbert please review the changes.

PR #2094, opened 2018-09-14
surveys number added(connects#2031)

yuviii:


src/app/surveys/surveys.component.ts
```diff
@@ -34,8 +35,12 @@ export class SurveysComponent implements OnInit, AfterViewInit {
 
   ngOnInit() {
     this.surveys.filterPredicate = filterSpecificFields([ 'name' ]);
-    this.getSurveys().subscribe((surveys) => {
-      this.surveys.data = surveys;
+    this.getSurveys()
+    .subscribe((surveys) => {
+      this.getSubmission().subscribe((submissions) => {
```
paulbert: We should not nest observables.  Please use switchMap or forkJoin to flatten this code.

src/app/surveys/surveys.component.html
```diff
@@ -22,6 +22,10 @@
       <mat-header-cell *matHeaderCellDef mat-sort-header i18n>Survey</mat-header-cell>
       <mat-cell *matCellDef="let element">{{element.name}}</mat-cell>
     </ng-container>
+    <ng-container matColumnDef="taken">
+      <mat-header-cell *matHeaderCellDef mat-sort-header i18n>No of taken</mat-header-cell>
+      <mat-cell *matCellDef="let element">{{ getTaken(element._id) }}</mat-cell>
```
paulbert: The `getTaken` function will run any time any part of the component changes*.  If there are a lot of submissions, this could severely affect performance.

It would be better to run the submission filter once on init and then add the number of times a survey is taken as a property for the objects in `this.surveys.data`.

*For more detail see https://blog.appverse.io/why-it-is-a-bad-idea-to-use-methods-in-the-html-templates-with-angular-2-30d49f0d3b16

yuviii:
@lmmrssa @paulbert please review the fixes

Issue #2081, opened 2018-09-13
In submission, exam and survey list should show recent at top 


singharpita:
i.e updated on by descending order.
![desc](https://user-images.githubusercontent.com/27276132/45494150-2fcdf300-b790-11e8-9350-c5475a3b751b.png)

lmmrssa:
We would also like to have sorting of submissions by other fields. So please use sortable table.

PR #2097, opened 2018-09-14
In submission, exam and survey list should show recent at top(connects #2081)

lucy811:


lmmrssa:
@lucy811  We would also like to have sorting of submissions by other fields. So please use sortable table.

Issue #2068, opened 2018-09-11
Configuration not getting updated


lmmrssa:
ConfigurationService is only getting data on application load so new configured application is not getting updated configuration until you reload the application. We should fetch new updated configuration on every configuration update.

PR #2103, opened 2018-09-17
[#2068] update configurations (connects #2068)

singharpita:


src/app/configuration/configuration.service.ts
```diff
@@ -40,6 +40,7 @@ export class ConfigurationService {
         'status': 'unread',
         'time': Date.now()
       };
+      this.configuration = configuration;
```
paulbert: This function should only create the notification on the parent planet.  We should not introduce the side effect of updating the configuration on the service as well.

In addition, this will only update the configuration when the configuration didn't exist on the parent and only for the open application.  Any other users who are logged in when the change will made will still see these errors.

paulbert:
Closing due to inactivity

Issue #2117, opened 2018-09-18
showing deleted meet up in dashborad


yuviii:
<img width="1307" alt="screenshot 2018-09-18 08 41 08" src="https://user-images.githubusercontent.com/6964191/45661795-c7c83580-bb1e-11e8-8c68-6dab6c96220e.png">

cant delete the meet up either

PR #2118, opened 2018-09-18
Fix deleted items on dashboard

paulbert:
Quick fix to solve the issue of deleted items showing up on the dashboard.

We may also need to clean up the documents on the shelf database after an item is deleted.  Not sure what the best way to do that is because it may make deleting documents from other databases very slow.


lmmrssa:
It is removing wrong meetup item. Always removing first on the list. Seems to be ok for course and resources

paulbert:
#2117

PR #2120, opened 2018-09-18
remove toggle filter since this should only show surveys (fixes #2067)

### LINKED ISSUE #2067: mySurvey shows Exam and Survey filter
i think Exam filter should be removed so there should be no toggle filter since it has only survey.
screenshot from 2018-09-11 07-37-56

leahphuong:
fixes #2067


lmmrssa:
@leahphuong You just removed the toggle button. We also need to make sure that `http://localhost:3000/mySurveys` should default to Survey. Right now its still default to Exam(Submission)

PR #2134, opened 2018-09-19
Check _changes to always store up to date planet configuration (connects #2068)

paulbert:
This will need to be tested with multiple devices connected to a raspberry pi to see if the _changes feeds impact performance.


paulbert:
This is a little larger of a change than before because every instance of `this.userService.getConfig()` had to be converted to use configurationService.

Slight snag currently due to a circular dependency, so labeling this WIP for now.

paulbert:
Needed to incorporate the `StateService` which is also introduced in #2172.

@dogi @singharpita @yuviii 
Both this and #2172 need reviews as they are very important performance improvements for Planet.

Issue #2052, opened 2018-09-06
feedback list status color separation


yuviii:
<img width="1337" alt="screenshot 2018-09-06 20 45 06" src="https://user-images.githubusercontent.com/6964191/45166030-cbea7e00-b215-11e8-9e5d-f98368ed24ad.png">

> Would be better if we use different color to differentiate status **close**  and **open**  for instance here.

PR #2137, opened 2018-09-19
2052-feedback status and priority list color

leahphuong:
(connects #2052) 
also add color to emphasize urgent item in priority list.

screen shot 2018-09-19 at 10 08 10 am

### DIFF
--- a/src/app/feedback/feedback.component.html
+++ b/src/app/feedback/feedback.component.html
@@ -44,7 +44,7 @@
       <!-- Priorty Column -->
       <ng-container matColumnDef="priority">
         <mat-header-cell *matHeaderCellDef mat-sort-header i18n>Priority</mat-header-cell>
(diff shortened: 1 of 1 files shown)


paulbert:
Let's use mat-chips for this `primary` color for open status & `accent` color for priority yes

paulbert:
https://material.angular.io/components/chips/overview

src/app/feedback/feedback.component.html
```diff
@@ -42,19 +42,25 @@
         </mat-cell>
       </ng-container>
       <!-- Priorty Column -->
-      <ng-container matColumnDef="priority">
+      <ng-container matColumnDef="priority" [ngSwitch]="true">
```
lmmrssa: ngswitch use should be based on variable and you do not need to do == case check on ngswitchcase

Please add ng-container inside mat-cell 
ngswitch=element.priority on ng-container
ngswitchcase=Yes on mat-chip-list

PR #2172, opened 2018-09-21
Speed up resources list in course step (fixes #2131)

### LINKED ISSUE #2131: Selecting resource for course step only shows first 25
Currently the list of resources to add to a course step will only show the first 25 because it is using a `_find` query with the default limit of 25.  This needs to show all resources.  See below image for the view I'm referencing.

Rather than use a `_find` query we should use the ResourcesService to avoid performance issues with our devices in the field.

image

paulbert:


paulbert:
Want to test this on the RaspberryPi.  We also need to test on a center since one of the uses of DialogsList is to list the communities attached to a nation.

paulbert:
@dogi @singharpita @yuviii 
I tested on a center (center.media.mit.edu) & a RaspberryPi and it is working.  Please review.

Issue #2164, opened 2018-09-20
Duplicate configuration


lmmrssa:
We are still seeing multiple configuration of same planet code. Its not happening when you try to configuration planet multiple times. Instead it seems like when user press **Submit** multiple times quickly its able to submit multiple request.

paulbert:
This is unrelated to quick clicks.  The configuration added to the nation has a different `_id` than the one on the community, so when that configuration is replicated it adds a second doc rather than updating.  This community was created with the branch on #2198, and the corresponding doc on dev has a different id.

![image](https://user-images.githubusercontent.com/9203229/46121450-bf888e00-c1e1-11e8-97fe-7fecbb7f7cf7.png)

![image](https://user-images.githubusercontent.com/9203229/46121465-cb745000-c1e1-11e8-9bb2-ea00717f832d.png)

lmmrssa:
@paulbert it was also duplicating on dev when doing new setup.

PR #2198, opened 2018-09-25
[#2164] duplicate configuration (connects #2164)

singharpita:
basically to prevent multiple clicking, i added the spinner in both new and updated configurations.

### DIFF
--- a/src/app/configuration/configuration.service.ts
+++ b/src/app/configuration/configuration.service.ts
@@ -96,9 +96,10 @@ export class ConfigurationService {
   }
 
   postConfiguration(configuration) {
-    return this.couchService.post('configurations', configuration).pipe(
-      switchMap(() => this.updateAutoAccept(configuration.autoAccept))
-    );
+    return forkJoin([
+      this.couchService.post('configurations', configuration),
+      this.updateAutoAccept(configuration.autoAccept)
+    ]);
   }
 
   updateAutoAccept(autoAccept) {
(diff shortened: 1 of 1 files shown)


src/app/configuration/configuration.component.ts
```diff
@@ -235,22 +235,29 @@ export class ConfigurationComponent implements OnInit {
...
-          this.planetMessageService.showMessage('Configuration Updated Successfully');
-        }
-      );
-    } else {
-      const admin = Object.assign(credentials, this.contactFormGroup.value);
-      this.configurationService.createPlanet(admin, configuration, credentials).subscribe((data) => {
-        this.planetMessageService.showMessage('Admin created: ' + credentials.name);
-        this.router.navigate([ '/login' ]);
-      }, (error) => this.planetMessageService.showAlert('There was an error creating planet'));
+    if (this.contactFormGroup.valid) {
```
paulbert: This is unnecessary.  allValid already checks the validity of all of the form groups.

src/app/configuration/configuration.component.ts
```diff
@@ -235,22 +235,29 @@ export class ConfigurationComponent implements OnInit {
...
+        );
+      } else {
+        const admin = Object.assign(credentials, this.contactFormGroup.value);
+        this.configurationService.createPlanet(admin, configuration, credentials).subscribe((data) => {
+          this.planetMessageService.showMessage('Admin created: ' + credentials.name);
+          this.router.navigate([ '/login' ]);
+        }, (error) => this.planetMessageService.showAlert('There was an error creating planet'));
+      }
+    }else {
+      Object.keys(this.contactFormGroup.controls).forEach(field => {
```
paulbert: Will not be reached because of the above.

paulbert:
See my comment on the issue, please #2164.  The issue is not related to quick clicking, so this will require some revision

PR #2227, opened 2018-09-27
Fix multiple answer correct choices tracking (fixes #2223)

### LINKED ISSUE #2223: Multiple choice multiple answer error
When trying to create survey/exam with multiple correct choice, it does not submit. There is validation error. Its working with single correct choice

paulbert:
#2223


lmmrssa:
It fixed course exam but survey is still having issue

lmmrssa:
Able to create exams and survey with all types

PR #2240, opened 2018-09-30
Update PouchDB Service and auth 

lawrencek0:
I had to make some changes to PouchDB Auth to make it work it with new PouchDB 7. I also updated the service to make it more sensible to deal with Pouch's API.


lawrencek0:
I had to make a lot of changes because I was facing issues with these changes as I was working on the Feedback Component (#2241)

Issue #2161, opened 2018-09-20
Mark read


lmmrssa:
We might also want to just have mark read option for notifications. Especially on **dev** we are gathering lot of notification and we need to click each on which navigates them to different page

paulbert:
We can call this "Mark all as read"

yuviii:
@paulbert @lmmrssa so is it making the background gray or something like for the notification row which is visited ?

PR #2242, opened 2018-10-01
[#2161]mark as read added(connects #2161)

yuviii:


src/app/notifications/notifications.component.ts
```diff
@@ -54,6 +56,13 @@ export class NotificationsComponent implements OnInit, AfterViewInit {
     }, (err) => console.log(err.error.reason));
   }
 
+  markRead(notification) {
```
lmmrssa: this function is not required as it is already done by readNotification function. You just need a way to call readNotification without triggering link

src/app/notifications/notifications.component.ts
```diff
@@ -13,12 +14,13 @@ import { MatTableDataSource, MatPaginator, PageEvent } from '@angular/material';
...
   @ViewChild(MatPaginator) paginator: MatPaginator;
   notifications = new MatTableDataSource();
-  displayedColumns = [ 'message' ];
+  displayedColumns = [ 'message', 'read' ];
   private onDestroy$ = new Subject<void>();
   emptyData = false;
 
   constructor(
     private couchService: CouchService,
+    private planetMessageService: PlanetMessageService,
```
lmmrssa: not required anymore

PR #2245, opened 2018-10-01
Code of conduct

empeje:
Requires votes

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -17,7 +17,7 @@ Examples of behavior that contributes to creating a positive environment include
 Examples of unacceptable behavior by participants include:
 
 * The use of sexualized language or imagery and unwelcome sexual attention or advances
-* Trolling, insulting/derogatory comments, and personal or political attacks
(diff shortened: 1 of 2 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,46 @@
...
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery and unwelcome sexual attention or advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
```
yuviii: it would be better if we include  **racism** also as it would be more specific to address the people of all kind of race with more honour.

empeje: I agree with racism included here, what do you think @dogi?

empeje:
Will merge this one after https://github.com/open-learning-exchange/planet/pull/2239

Issue #2251, opened 2018-10-02
Your progress on nation course


lmmrssa:
When looking after course list from nation, we should not have progress.
![screen shot 2018-10-02 at 08 15 43](https://user-images.githubusercontent.com/2909490/46326259-afc2dc80-c61b-11e8-9baf-7ce93ae5518d.png)

PR #2266, opened 2018-10-02
[#2251] your progress on nation (connects #2251)

singharpita:


lmmrssa:
Please check parent no planettype.

PR #2285, opened 2018-10-04
[#2156] Pass count data for resources into dialog prompt

### LINKED ISSUE #2156: deletion count
screenshot 2018-09-20 20 21 54

It would be better to show the total count of selected resources.

jimkosh:
Will show number of resources before deleting in dialog prompt.
Fixes #2156 
image


jimkosh:
@paulbert should I pass in deletion-count data for courses and meetups as well?

jimkosh:
@paulbert sweet, thanks paul.  should I apply these changes for other deletion dialog prompts?

Issue #2263, opened 2018-10-02
planet-markdown-textbox does not show error on submit


paulbert:
When clicking submit with a required `planet-markdown-textbox` empty, the textbox is not highlighted as an error field.  You still need to click on the textbox for the error to show.

The error state should activate on the textbox when we submit a form if applicable.

PR #2292, opened 2018-10-08
[#2263] fix planet-markdown-textbox does not show error on submit (connects#2263)

lucy811:


lucy811:
@paulbert please review the change

Issue #2288, opened 2018-10-04
Filter out satellite user in reports


paulbert:
It appears that the `satellite` user is being counted in reports in the user count.  We should filter that user out since it does not correspond to an actual registered member.

nadaa:
Hi @paulbert, I will work on this, if possible?

nadaa:
is there a component for repots?

nadaa:
@Paul, can you please add more how to reproduce this issue?
What is a satellite user?

PR #2314, opened 2018-10-09
Filter satellite user from report

nadaa:
issue #2288 
Remove satellite user from the **Total Users** in the community report.

### DIFF
--- a/src/app/manager-dashboard/reports/reports.service.ts
+++ b/src/app/manager-dashboard/reports/reports.service.ts
@@ -40,7 +40,7 @@ export class ReportsService {
     const obs = planetCode ? this.couchService.findAll('_users') :
       this.couchService.findAll('child_users').pipe(map((users: any) => users.filter(user => user.planetCode === planetCode)));
     return obs.pipe(map((users: any) => {
-      users = users.filter(user => user.name !== 'satellite')
+      users = users.filter(user => user.name !== 'satellite');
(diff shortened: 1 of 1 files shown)


lmmrssa:
@nadaa we do not want satellite user on any report. So we can just filter it when we are filtering user for planetcode.

nadaa:
Thank @lmmrssa , you mean I should filter it here:     
```
this.couchService.findAll('child_users').pipe(map((users: any) => users.filter(user => user.planetCode === planetCode)));
```

src/app/manager-dashboard/reports/reports.service.ts
```diff
@@ -41,7 +41,7 @@ export class ReportsService {
       this.couchService.findAll('child_users').pipe(map((users: any) => users.filter(user => user.planetCode === planetCode)));
     return obs.pipe(map((users: any) => {
       return ({
-        count: users.length,
+        count: users.filter(user => user.name !== 'satellite').length,
```
paulbert: Rather than filter just for the count property of the return statement, you should add a new line between lines 42 and 43 to filter the users variable `users = users.filter(...)`.  That way both `count` and `byGender` properties to not include the satellite user.

nadaa: Ok, got it. Thanks @paulbert!

Issue #2308, opened 2018-10-09
resources_activities does not count resource views in courses


paulbert:
Viewing a resource from a course step does not count as a visit to a resource.  This should be counted.

paulbert:
Simplest way is maybe to post the new document to resources_activities in ResourcesViewer instead of ResourcesView

PR #2319, opened 2018-10-11
[#2308] resource count (connects #2308)

singharpita:


paulbert:
Realized in updating that we can just call the notification function from `setResource`.  Waiting for build to complete then will merge

PR #2329, opened 2018-10-12
change toggle button color (fixes #2325)

### LINKED ISSUE #2325: Submission: Exam and Survey toggle
Exam and Survey toggle button when active should be highlighted 
screenshot from 2018-10-11 19-48-46

svlesiv:
Fixes #2325

## Description
Added style in styles.scss file which overwrites Angular Material style for toggle button color.

## Screenshots
screen shot 2018-10-12 at 10 54 03 am

screen shot 2018-10-12 at 10 54 36 am

### DIFF
--- a/src/app/_variables.scss
+++ b/src/app/_variables.scss
@@ -3,6 +3,7 @@
 $primary-hue: 500;
 $accent-hue: 500;
 $warn-hue: 500;
+$light-hue: 100;
 
 $planet-primary: $mat-blue;
 $planet-accent: $mat-amber;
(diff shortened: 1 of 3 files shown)


paulbert:
@svlesiv Generally we try to avoid directly changing Material styles (with the `mat-` prefix), but in this case there isn't an alternative so good job.  I just moved it to the `planet-mat-theme.scss` file for better organization and used a lighter hue to contrast it from the toolbar just above it.

@singharpita I'd also like to clarify the original issue.  The selected button is already highlighted by default in gray. Are you saying the gray color is too subtle?

Issue #2312, opened 2018-10-09
Ignore blank data on reports


paulbert:
For ratings & resource_activities blank is log without `title` and for login_activities blank is log without `user` field.

nadaa:
@paulbert, I am trying to produce this issue, I opened the report from the admin account, this is how it looks like:
![](https://i.imgur.com/xcES5Al.png)

there are some fields which are empty like (average rating, highest rate,..etc), shall we add a value such as 'no rating' or 0 instead?

paulbert:
What I'm referring to is the below, where there are lines in the reports with visits/rating on the right but blank on the left.  We should filter out the data where `user` or `title` is blank so those lines do not show up.

![image](https://user-images.githubusercontent.com/9203229/46886663-89761b80-ce29-11e8-8d79-35b431c1a285.png)

PR #2332, opened 2018-10-14
Exclude blank data from report

nadaa:
Some resources appear without titles and users without names as it was mentioned here #2312 .


src/app/manager-dashboard/reports/reports-detail.component.ts
```diff
@@ -37,7 +37,7 @@ export class ReportsDetailComponent {
 
   getLoginActivities() {
     this.activityService.getLoginActivities(this.planetCode).subscribe(visits => {
-      this.reports.visits = visits.slice(0, 5);
+      this.reports.visits = visits.filter( visit => visit.user !== '').slice(0, 5);
```
paulbert: After moving to `reports.service.ts` we should also add the condition `visit.user !== undefined`. Same for line 53.

Issue #2335, opened 2018-10-16
Sending course do not send related resource/exam


lmmrssa:
When sending course from nation to child. It only sends course not resources or exams that are related to it.

lmmrssa:
PR created https://github.com/open-learning-exchange/planet/pull/2342

paulbert:
PR merged

PR #2342, opened 2018-10-17
Course with resource and exam (connects #2335)

lmmrssa:


lmmrssa:
@paulbert thinking to just have one button to get it all. Wanted to discuss will we require those getting individually or not?

Issue #2328, opened 2018-10-12
Sending survey includes password info in submissions doc


paulbert:
Submissions db is insecure, so we should make sure the fields `password_scheme`, `iterations`, `derived_key`, and `salt` are not added as part of the user object on the submission doc

paulbert:
More specifically, these fields should not be saved on the StateService for the `_users` database.

PR #2347, opened 2018-10-18
[#2328] sending survey prevent password(connects #2328)

yuviii:


src/app/surveys/surveys.component.ts
```diff
@@ -84,9 +84,10 @@ export class SurveysComponent implements OnInit, AfterViewInit, OnDestroy {
 
   openSendSurveyDialog(survey) {
     forkJoin([
-      this.dialogsListService.getListAndColumns('_users'),
```
paulbert: Rather than changing this here, we should change the behavior in `DialogsListService`

src/app/shared/dialogs/dialogs-list.service.ts
```diff
@@ -47,8 +49,8 @@ export class DialogsListService {
 
   getListAndColumns(db: string, selector?: any, planetField: string = 'local') {
     selector = selector || this.defaultSelectorFunctions()[db];
-    const fields = db === '_users' || db === 'child_users' ? this.userService.userProperties : [];
-
+    const fields = db === '_users' || db === 'child_users' ? this.couchService.post('_users/_find', { 'selector': { } }) : [];
```
paulbert: You should return a post to the `db + /_find` endpoint if `db === '_users'`.  Remember to include `this.userService.userProperties` as `fields` for that request.

PR #2366, opened 2018-10-22
Update (reduce) font sizes across application

paulbert:
Checking out the idea of reducing font sizes across the application.

### DIFF
--- a/src/planet-mat-theme.scss
+++ b/src/planet-mat-theme.scss
@@ -1,10 +1,30 @@
 @import '~@angular/material/theming';
 // Plus imports for other components in your app.
 
+// Typography settings
+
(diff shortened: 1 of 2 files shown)


lmmrssa:
@paulbert as we decrease fontsize i see we have more height on browser so we can increase size of container as well. right now its not using whole height
screen shot 2018-10-29 at 20 50 46

paulbert:
@lmmrssa More worrying about the above is that the font has changed to Times New Roman on the left panel.  Will need to fix that before merging.

I'm not sure why the panel isn't filling out the whole container there.  Will have to look into that as well.

jessewashburn:
Probably not worth the headache? Users can just zoom in.

Issue #2362, opened 2018-10-20
Create ng serve commands for translations


paulbert:
We have npm commands in our `package.json` for serving the application in development with translations on, but those no longer work with the latest version of the Angular CLI.

The best approach seems to be to create different configurations for the `ng serve` command in `angular.json` which will build the app with different translations.

PR #2367, opened 2018-10-22
Add configuration for serving languages in dev environment

paulbert:
Command to serve with different language in development: `ng serve --configuration=language`

As it is written to change language you need to change the `angular.json` file.  There might be a way to do that with another custom flag which would be better.


paulbert:
Connected to #2362

paulbert:
Closing in favor of #4032

Issue #2370, opened 2018-10-23
"undefined" showing on User/Survey Edit profile


singharpita:
![screenshot from 2018-10-23 07-25-07](https://user-images.githubusercontent.com/27276132/47329571-f5ac1700-d694-11e8-8926-3716fba5ed34.png)
![screenshot from 2018-10-23 07-25-25](https://user-images.githubusercontent.com/27276132/47329579-fba1f800-d694-11e8-9294-311810800bee.png)

singharpita:
its showing in most of the form.
![screenshot from 2018-10-23 07-42-38](https://user-images.githubusercontent.com/27276132/47330246-402e9300-d697-11e8-9c70-011891fac3d9.png)

lmmrssa:
@paulbert looks like {{number}} portion you moved to end of message is adding it to all as *undefined*

paulbert:
@lmmrssa Yes it looks that way.  Forgot that in JavaScript if you have string concatenation and a variable is not defined is converts it to the string `undefined`.  

Simply doing `number || ''` won't work either since that will convert 0 to an empty string.  Can't think of the most concise way to write that in JavaScript off the top of my head.

nadaa:
@singharpita, @paulbert , can I work on this?

lmmrssa:
sure @nadaa

PR #2385, opened 2018-10-24
Remove undefined error msg

nadaa:
Refer to issue #2370


nadaa:
undefined is removed here, but if the error is 0 it will show an empty string, I will think about a solution.

PR #2393, opened 2018-10-25
No attachment hide resource button (fixes #2304)

### LINKED ISSUE #2304: Courses: resource view which doesn't have file in it
if choose the resources which doesn't have any file in it. opening up that resources result in page not found.

It should be redirected to the home page or the option should not be provided in Add resource list in case there is no file in the resources.

screenshot from 2018-10-09 08-09-17

svlesiv:
Fixes #2304 

## Description
I adjusted courses-view.component.html file to hide a button for resources which don't have attachments or to exclude those resources from the list of resources.

## Screenshots
When a step has only 1 resource with no attachments in it:
screen shot 2018-10-25 at 12 07 45 am

When a step has 2 resources and one doesn't have attachments:
screen shot 2018-10-25 at 12 08 42 am

Only resource with an attachment will appear in the list.
screen shot 2018-10-25 at 12 08 10 am

### DIFF
--- a/src/app/courses/view-courses/courses-view.component.html
+++ b/src/app/courses/view-courses/courses-view.component.html
@@ -45,10 +45,12 @@
             </mat-expansion-panel-header>
             <td-markdown [content]="step.description"></td-markdown>
             <mat-action-row>
-              <a *ngIf="step?.resources?.length === 1" mat-raised-button color="accent" class="margin-lr-10" [href]="resourceUrl(step.resources[0])" target="_blank" i18n>View Resource</a>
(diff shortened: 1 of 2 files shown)


lmmrssa:
@singharpita we had some talk on it and as user may update resource after it has been attached to course we wanted it to hide button. So no matter when attachment is removed it will have same effect.

lmmrssa:
@svlesiv instead of checking attachment on course html file we should filter it from component so all previous logics works

svlesiv:
@paulbert, @singharpita, and @lmmrssa thanks for your advice and review.

PR #2422, opened 2018-10-30
ole/stretch64 0.5.3

dogi:


lmmrssa:
node_modules is not found as this folder is not created so it gives error. After creating node_modules folder it seems to work fine
screen shot 2018-10-30 at 16 21 20

Issue #2376, opened 2018-10-23
Sorting in feedback list on date not working


yuviii:

paulbert:
The error means we likely need to make sure we check the type in the SortingDataAccessor
Error:
![image](https://user-images.githubusercontent.com/9203229/47334277-59fab680-d654-11e8-862d-2def04a2c84f.png)

PR #2428, opened 2018-10-30
[#2376] date sorting in feedback list (connects #2376)

leahphuong:


src/app/feedback/feedback.component.ts
```diff
@@ -70,7 +70,7 @@ export class FeedbackComponent implements OnInit, AfterViewInit, OnDestroy {
     this.user = this.userService.get();
     this.getFeedback();
     this.feedback.filterPredicate = composeFilterFunctions([ filterDropdowns(this.filter), filterSpecificFields([ 'owner', 'title' ]) ]);
-    this.feedback.sortingDataAccessor = (item, property) => item[property].toLowerCase();
+    this.feedback.sortingDataAccessor = (item, property) => item[property].toString().toLowerCase();
```
yuviii: item[property] return date which is long int so it don't have toLowerCase() method. just remove this method.
this.feedback.sortingDataAccessor = (item, property) => item[property];

paulbert: See the CommunityComponent `sortingDataAccessor` function.  That checks the type of `item[property]` and applies the `toLowerCase` method if it is a string.

Rather than duplicate, you can add that function to `src/app/shared/table-helpers.ts` and use in both places.

Issue #2414, opened 2018-10-30
ResourceView: Refine detail properties shown


singharpita:
Level is missing from resource detail page.
![screenshot from 2018-10-30 07-19-06](https://user-images.githubusercontent.com/27276132/47689974-52b74800-dc14-11e8-81f8-6492de0ea787.png)
![screenshot from 2018-10-30 07-19-31](https://user-images.githubusercontent.com/27276132/47689981-58149280-dc14-11e8-9569-da914f8244ba.png)

Also In subject label, can the value be separated by space after comma?

paulbert:
Updated the title and putting the two todos together so they're easier to read
* Add Level as another property shown
* Put spaces in between the subjects

PR #2429, opened 2018-10-30
Add level properity and spacing in resource details

nadaa:
issue #2414 
Added the following to resources view:
-  Level property
-  Space between multiple items


paulbert:
@nadaa Simpler to use [`.join`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join).  Will update to use that and merge shortly.  Good job.

Issue #2417, opened 2018-10-30
If already enrolled in a course, "enroll in selected" should be disabled


BhargaviNadendla:
![image](https://user-images.githubusercontent.com/25775298/47691107-b44cc780-dbbe-11e8-9378-14977575658d.png)

paulbert:
Let's try making the enroll in selected number also only count selected items that the user hasn't already enrolled in, too

PR #2430, opened 2018-10-31
[#2417] disable button if already enrolled in a course (connects # 2417)

lucy811:


lucy811:
@paulbert  please review the change

PR #2458, opened 2018-11-06
Add Total Visits and Total Resource Views to report detail

paulbert:


src/app/manager-dashboard/reports/reports-detail.component.ts
```diff
@@ -48,6 +48,7 @@ export class ReportsDetailComponent {
 
   getLoginActivities() {
     this.activityService.getLoginActivities(this.planetCode).subscribe(({ byUser, byMonth }: { byUser: any[], byMonth: any[] }) => {
+      this.reports.totalMemberVisits = byUser.reduce((total, resource: any) => total + resource.count, 0);
```
lmmrssa: could we change resource to member :)

PR #2461, opened 2018-11-06
Update text of course

lmmrssa:


lmmrssa:
@paulbert there is term `Enroll` on course titlebar so shall we also change that to `join` or change everything to `enroll`

PR #2466, opened 2018-11-07
Full screen toggle (fixes #2420)

### LINKED ISSUE #2420: Full screen on editor
There is full screen mode on markdown editor which if confusing to use.
screen shot 2018-10-30 at 07 59 13
Button on image leads to look like this on next image. And its not clear how to return back. You have to press **Esc** key get back
screen shot 2018-10-30 at 07 59 32

So we might want to clean some of these action

svlesiv:
fixes #2420 

## Screenshots
screen shot 2018-11-06 at 9 43 32 pm

screen shot 2018-11-06 at 9 43 20 pm

screen shot 2018-11-06 at 9 43 45 pm

### DIFF
--- a/src/app/shared/forms/planet-markdown-textbox.scss
+++ b/src/app/shared/forms/planet-markdown-textbox.scss
@@ -26,12 +26,18 @@ planet-markdown-textbox {
   }
 
   .CodeMirror.CodeMirror-fullscreen {
-    padding-top: 30px;
+    padding-top: 70px;
     padding-left: 80px;
   }
 
   .CodeMirror.CodeMirror-fullscreen .CodeMirror-scroll{
     height: 100vh;
     min-height: 100vh;
+  }
+
+  .editor-toolbar.fullscreen{
+    z-index: 1000;
+    top:70px;
+    left: 65px;
   }
 }
--- a/src/app/shared/forms/planet-markdown-textbox.scss
+++ b/src/app/shared/forms/planet-markdown-textbox.scss
@@ -24,4 +24,14 @@ planet-markdown-textbox {
     height: 15vh;
     min-height: 15vh;
   }
+
+  .CodeMirror.CodeMirror-fullscreen {
+    margin-top: 10px;
+    margin-left: 70px;
+  }
+
+  .CodeMirror.CodeMirror-fullscreen .CodeMirror-scroll{
+    height: 100vh;
+    min-height: 100vh;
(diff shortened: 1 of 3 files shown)


yuviii:
Left margin should be reduced i guess
@paulbert @lmmrssa 

screenshot 2018-11-07 13 21 33

src/app/shared/forms/planet-markdown-textbox.scss
```diff
@@ -26,8 +26,8 @@ planet-markdown-textbox {
   }
 
   .CodeMirror.CodeMirror-fullscreen {
-    margin-top: 10px;
-    margin-left: 70px;
+    padding-top: 30px;
+    padding-left: 80px;
```
yuviii: padding left seems to be big. 10 px looks ok.

lmmrssa:
@svlesiv nice work, looks good. There are also other action next to it. 
`Toggle Preview` `Toggle Side by Side` could you also test with these.

svlesiv:
@lmmrssa thanks!) 
`Toggle Preview` and `Toggle Side by Side` doesn't work well. I will try to fix it.

src/app/shared/forms/planet-markdown-textbox.scss
svlesiv: You are right.  Probably I would need to add some media queries for that.

svlesiv:
screen shot 2018-11-07 at 5 15 17 pm

screen shot 2018-11-07 at 5 15 35 pm

svlesiv:
screen shot 2018-11-07 at 5 53 19 pm

screen shot 2018-11-07 at 5 53 55 pm

paulbert:
You can reference the way the component creators implemented it here:
https://teradata.github.io/covalent/#/components/text-editor

svlesiv:
Hi @paulbert. Thanks for the link. I will take a look and try to implement that technic.

paulbert:
I looked at this a little more and found what does it: check out the `text-editor-demo` element in their implementation.  It uses `position: absolute` to stretch the view to fill the screen and `transform: translateY(0px)` to set that as the containing element for the fixed position editor.  Works with the two navigation layouts as is.

svlesiv:
Thanks for your advice @paulbert. I looked on a demo. Seems like planet code doesn't generate classes for animation. So I overwrote the values of existing classes.

Issue #2453, opened 2018-11-06
Updated on field is not updating: Submissions


yuviii:
http://localhost:3200/eng/submissions

paulbert:
See #2471 for why this was closed

PR #2471, opened 2018-11-07
[2453] lastupdated time fixed: connects(#2453)

yuviii:


lmmrssa:
@yuviii I don't think we need to update last updated time on grading. last updated should only update when you update any answer submission. which includes continuing exam or resubmitting same survey and that seems to be happening.
@paulbert please confirm. I feel we can close this issue #2453

paulbert:
@lmmrssa Yes I agree with what you wrote.  Closing the issue as well

PR #2478, opened 2018-11-09
Load resources on login

paulbert:
Loads the resources into StateService to reduce the load time of the resources table is reduced.


yuviii:
I got Error message 
screenshot 2018-11-12 10 24 47
screenshot 2018-11-12 10 24 21

PR #2479, opened 2018-11-09
Add dialog for labels (connects #2502)

paulbert:
Adds a dialog for some preset labels to add to resources.

Currently just showing "Subjects", but will add "Language", "Level", "Media", & "Resource Type"


paulbert:
Before switching to this branch to test: Make sure you have created a few labels on resources (you can do this from adding or managing a resource).

Two things to test:
1. Filtering by label on the resources list page
2. Adding and removing labels from the resources add or manage pages

Next steps:
- Include functionality to add a label (see #2501)
- Some linking between labels so each label has a list of linked labels below.  Clicking on the linked label would filter/add both labels to the currently selected list.

lmmrssa:
Looks nice
screen shot 2018-11-15 at 16 24 16

PR #2480, opened 2018-11-09
Vbguest autoupdate false

lmmrssa:


paulbert:
Tested `prod` first, which worked fine.

Error on first try with `dev`:
image

paulbert:
@lmmrssa Uncommented the npm install line in dev in this PR rather than making a new one

PR #2484, opened 2018-11-11
Disabling delete file if no file(fixes #2086)

### LINKED ISSUE #2086: delete file option is visible though there is not any file
screenshot 2018-09-13 20 28 06

> Delete file should is visible though there is not any file uploaded. After clicking on it it says resources updated successfully.
It would be possibly more accurate  either by disabling the option or notify user with no file to delete.

BhargaviNadendla:


paulbert:
Note: continuation of #2293 

@BhargaviNadendla Ideally you should just create one PR so we can see the history of the conversation while reviewing.

paulbert:
Now the "Delete File" checkbox is always disabled when loading the view, even if you are updating a resource which already has a file attached.

PR #2530, opened 2018-11-16
Fix console error on resources page (fixes #2278)

### LINKED ISSUE #2278: Parent resource error
CoursesService subscribes to ratings changes, and it looks like there is an error if the user goes to the resources page before viewing courses in `courses.service.ts`.  Doesn't seem to be breaking, but we should fix regardless.

image

paulbert:
To test: load resources or parent resources from manager dashboard before loading courses.  There should not be new errors in the console.


paulbert:
#2278

Issue #2523, opened 2018-11-15
Error when loading resources on nation


paulbert:
![image](https://user-images.githubusercontent.com/9203229/48561091-3de9de00-e8bd-11e8-90cb-1ec89697211f.png)

paulbert:
The related issue was fixed in #2531

PR #2531, opened 2018-11-16
Show parent labels when viewing parent (connects #2523)

paulbert:
To test: go to parent resources and click filter labels.  Should see labels from the parent planet.

Found the issue in testing #2523


yuviii:
@paulbert agreed with @singharpita , label should be made case insensitive in filter field.

Issue #2526, opened 2018-11-15
Submission send survey: response is not being updated


singharpita:
when a survey is send to the user, that survey can be submitted multiple number of times.
but the response is not being updated in Submission survey list.

PR #2533, opened 2018-11-19
[#2526] submission send survey update response (connects #2526)

singharpita:
when there is previous submission, its putting that answer as string so checkbox was not being able to update that answer.
and everytime as we provide the new answer to be submitted for survey hence i removed the old answer in case of survey.

@paulbert  do you have any better feedback?


lmmrssa:
@paulbert it was the case from the beginning. So if we just need to move condition to not have join in case of 'take'

PR #2548, opened 2018-11-20
Fix first logout not working (fixes #2425)

### LINKED ISSUE #2425: Logout doesn't work first time
Initial login to new planet logout doesn't work and has the below message in the console.

image

The leading theory is that the issue is with the `login_activities` database.  When the admin logs in there should be a document created on that database, but that doesn't appear to happen:

image

paulbert:
@yuviii While checking out another issue, I realized this may be due to the configuration on StateService not getting updated when we do the initial setup.  I made a few tweaks and I think it's working now.

#2425

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -1,7 +1,8 @@
 import { Component } from '@angular/core';
 import { DomSanitizer } from '@angular/platform-browser';
 import { MatIconRegistry } from '@angular/material';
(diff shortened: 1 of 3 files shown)


lmmrssa:
@paulbert still getting error when configuring and then login and logout.
screen shot 2018-11-21 at 12 44 45

paulbert:
@lmmrssa You're right, not sure why it worked for me yesterday.  Check now, it's working for me with this change.

PR #2550, opened 2018-11-20
Course progress leader overview

paulbert:
Change the progress view to start with an overview of all course steps.  You can click on the chart or use the menu in the top right to view a detail on the exams in the course.


paulbert:
There is a course on `dev` that you can pull to test (the issue # is in the title & instructions in the description).  It's best to test with multiple users taking the exams and clicking the wrong answer a random number of times to populate the charts.

PR #2555, opened 2018-11-21
Adding tooltip to show users selected (fixes #2399)

### LINKED ISSUE #2399: Using filter does not clean selection
When we try to send survey. After filtering user list, and submitting.
It makes feel user that he/she is submitting only those which are visible but it is submitting all selected records.
https://drive.google.com/open?id=1ICOA8kR2qCV4HFRRO0gW8DvsWsat-WFa

BhargaviNadendla:


src/app/shared/dialogs/dialogs-list.component.html
```diff
@@ -30,6 +30,11 @@
   [pageSizeOptions]="[5, 10, 20, 50, 100, 200]">
 </mat-paginator>
 <mat-dialog-actions>
+  <button mat-raised-button
```
paulbert: This should not be a button because there is no action when clicked.  Should just be text

src/app/shared/dialogs/dialogs-list.component.html
```diff
@@ -30,6 +30,11 @@
...
+          [matTooltip]=selections()
+          matTooltipClass = "tooltip">
```
paulbert: HTML should not have spaces between the equals sign.

src/app/shared/dialogs/dialogs-list.component.html
```diff
@@ -30,6 +30,11 @@
   [pageSizeOptions]="[5, 10, 20, 50, 100, 200]">
 </mat-paginator>
 <mat-dialog-actions>
+  <button mat-raised-button
+          [matTooltip]=selections()
```
paulbert: This needs to have quotes.  Also we should not have a function assigned to this as the Angular change detection will just run this function every time.  You should have a string on the component which is updated when the selection changes.

src/app/shared/dialogs/dialogs-list.component.ts
```diff
@@ -20,6 +20,10 @@ import { composeFilterFunctions, filterDropdowns } from '../table-helpers';
       overflow-y: auto;
       height: calc(100% - 160px);
     }
+    ::ng-deep .tooltip {
```
paulbert: `ng-deep` is deprecated.  We should avoid using it if possible.

BhargaviNadendla:
@paulbert ::ng-deep has been deprecated but since then there was no alternative to it put forth by angular team. Let us use it until there comes an alternative.

paulbert:
@BhargaviNadendla A note about the `Array.filter()` function since your use of it as of your last commit was not good:

`Array.filter()` should return a comparison in the predicate function.  The result will be an array that is a subset of the original array with only items that returned true.

In bc3bf19 you have a filter function which does not return a value and only performs side effect actions, which are difficult to catch when reading and raise the complexity for future developers trying to understand the code.  For all of the built in array functions we should avoid side effects with the exception of `forEach`

BhargaviNadendla:
@paulbert Thank you for letting me know.

PR #2556, opened 2018-11-21
Mutli Resource Visits Entries- 2415

### LINKED ISSUE #2415: Report Detail: Most visited resource count is increased by 2
Most Visited Resources count increased by 2 rather than 1 on single visit to the particular resource.

nadaa:
fixes #2415 

I solved this by adding a variable to control updating the resource.

### DIFF
--- a/src/app/resources/view-resources/resources-viewer.component.ts
+++ b/src/app/resources/view-resources/resources-viewer.component.ts
@@ -45,7 +45,7 @@ export class ResourcesViewerComponent implements OnChanges, OnDestroy {
 
   ngOnChanges() {
     if (this.shownResourceId === this.resourceId) {
-        this.resourcesService.requestResourcesUpdate(this.parent, this.fetchRating);
+      this.resourcesService.requestResourcesUpdate(this.parent, this.fetchRating);
     }
   }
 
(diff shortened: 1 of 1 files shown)


lmmrssa:
@nadaa It is still updating resource_activities for image. For other it is updating once but for image its still updating twice.

nadaa:
@lmmrssa, yes, this is not the optimal solution.

src/app/resources/view-resources/resources-viewer.component.ts
```diff
@@ -37,12 +38,15 @@ export class ResourcesViewerComponent implements OnChanges, OnDestroy {
   ) {
     this.resourcesService.resourcesListener(this.parent).pipe(takeUntil(this.onDestroy$))
       .subscribe((resources) => {
+        this.isResource = false;
```
paulbert: Save resource id here (maybe as `shownResourceId`)

src/app/resources/view-resources/resources-viewer.component.ts
```diff
@@ -37,12 +38,15 @@ export class ResourcesViewerComponent implements OnChanges, OnDestroy {
...
         this.setResource(resources.find((r: any) => r._id === this.resourceId));
       });
   }
 
   ngOnChanges() {
-    this.resourcesService.requestResourcesUpdate(this.parent, this.fetchRating);
+    if (this.isResource) {
```
paulbert: Check if resource id equals `this.resourceId` here

paulbert:
Ooops, I gave bad advice before.  Since `requestResourceUpdate` is started asynchronously in the parent component, `shownResourceId` is not set before the initial `ngOnChanges` and the activity is still logged twice (or three times in some of my testing).

Needed to check after the request completes.

nadaa:
@paulbert, I checked before submitting, it was working!

Issue #2569, opened 2018-11-22
Auto focus password on Admin password confirmation


lmmrssa:
When we ask for **Admin Confirmation** we need to click on input field to be able to type.
It will be better if that field is auto focused and user can just type

PR #2573, opened 2018-11-23
[#2569] auto focus to admin password (connects #2569)

singharpita:


src/app/shared/dialogs/dialogs-form.service.ts
```diff
@@ -11,11 +11,11 @@ import {
 export class DialogsFormService {
   constructor(private dialog: MatDialog, private fb: FormBuilder) { }
 
-  public confirm(title: string, fields: any, formGroup: any): Observable<boolean> {
+  public confirm(title: string, fields: any, formGroup: any, autoFocus?: false): Observable<boolean> {
```
paulbert: @singharpita This syntax is incorrect.  To set a default value for a function parameter you should use an `=` and no `?` as in:

```
autoFocus = false
```

Issue #2549, opened 2018-11-20
Send course to community


lmmrssa:
This is the data received by community for course sent by nation which seems off so need to check.
```
{"docs":[
{"_id":"2ad8b5978d5d2acd7d9ef71e68081ae5","_rev":"1-5df0fcdd1f10b902f983e9d5e501d07f","db":"exams","sendTo":"peacebuildernp","item":{"_id":"2ad8b5978d5d2acd7d9ef71e68052a66","_rev":"1-77c1b3fefec2da9b822fee0065c32fd7","name":"Week 1","passingPercentage":100,"questions":[{"title":"Did you complete this week's assignments?","body":"","type":"select","correctChoice":["_zcy663vu6"],"marks":1,"choices":[{"text":"Yes","id":"_zcy663vu6"},{"text":"No","id":"_giie2q4lu"}]}],"type":"courses","totalMarks":1}},
{"_id":"2ad8b5978d5d2acd7d9ef71e680823e0","_rev":"1-19e666d3470e8e13dc2094c2f81d6edd","db":"exams","sendTo":"peacebuildernp","item":{"_id":"2ad8b5978d5d2acd7d9ef71e680534da","_rev":"1-a69254f5966210034e0aa9061ddc1e54","name":"Week 2","passingPercentage":100,"questions":[{"title":"Did you complete this week's assignments?","body":"","type":"select","correctChoice":["_oyp5uimyj"],"marks":1,"choices":[{"text":"Yes","id":"_oyp5uimyj"},{"text":"No","id":"_pwpn43sam"}]}],"type":"courses","totalMarks":1}},
{"_id":"2ad8b5978d5d2acd7d9ef71e68082e49","_rev":"1-72a95dbb9b585e28cda49146db341c5c","db":"exams","sendTo":"peacebuildernp","item":{"_id":"2ad8b5978d5d2acd7d9ef71e68054433","_rev":"1-1867912dfcb318dfb25b9623b5410776","name":"Week 3","passingPercentage":100,"questions":[{"title":"Did you complete this week's assignments?","body":"","type":"select","correctChoice":["_6egkffeh1"],"marks":1,"choices":[{"text":"Yes","id":"_6egkffeh1"},{"text":"No","id":"_tzfd9d1z1"}]}],"type":"courses","totalMarks":1}},
[...]

PR #2577, opened 2018-11-26
Send course to child (connects #2549)

lmmrssa:
Disabled Send course to parent till we figure out resolving conflict.
Remove empty exams while sending courses

### DIFF
--- a/src/app/courses/courses.component.ts
+++ b/src/app/courses/courses.component.ts
@@ -322,7 +322,7 @@ export class CoursesComponent implements OnInit, AfterViewInit, OnDestroy {
       const resourcesToSend = [].concat.apply([], coursesToSend.map(course =>
           [].concat.apply([], course.steps.map(step => step.resources))
         ));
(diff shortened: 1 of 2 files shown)


src/app/courses/courses.component.ts
```diff
@@ -322,7 +322,7 @@ export class CoursesComponent implements OnInit, AfterViewInit, OnDestroy {
       const resourcesToSend = [].concat.apply([], coursesToSend.map(course =>
           [].concat.apply([], course.steps.map(step => step.resources))
         ));
-      const examsToSend = [].concat.apply([], coursesToSend.map(course => course.steps.map(step => step.exam)));
+      const examsToSend = [].concat.apply([], coursesToSend.map(course => [].concat.apply([], course.steps.map(step => step.exam || []))));
```
paulbert: We should refactor to avoid so much nesting.

lmmrssa: updated now to use flatmap

PR #2578, opened 2018-11-26
[#2546] Show question title in exams-view component

### LINKED ISSUE #2546: Show question title in ExamsView
Right now we are in the process of creating courses and exams for Somalia.  Because it's not documented well the tests only have a `title` property and no `body`, making the question look odd (with just the radio buttons or input & nothing else).

Rather than trying to clean this data we should show the exam title on this view.

svlesiv:
fixes #2546 

## Screenshots

screen shot 2018-11-26 at 1 31 54 pm

screen shot 2018-11-26 at 1 32 37 pm

### DIFF
--- a/src/app/exams/exams-view.component.html
+++ b/src/app/exams/exams-view.component.html
@@ -12,6 +12,7 @@
   <button mat-icon-button [disabled]="questionNum === maxQuestions" (click)="nextQuestion(questionNum)"><mat-icon>navigate_next</mat-icon></button>
   </mat-toolbar>
   <div class="view-container">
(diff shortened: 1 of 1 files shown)


paulbert:
Before I just had a `<b>` tag, but without the <p> below it that is a bit squished.

I think we should move towards using the Material Typography (started in #2366).  Considering that, I think we should just use the `.mat-h2` class instead of an HTML tag.

svlesiv:
Thanks for reviews and advice. With `.mat-h2` class the title looks not so bulky.

screen shot 2018-11-28 at 9 50 43 pm

PR #2587, opened 2018-11-27
Prevent exam to close with unanswered questions

### LINKED ISSUE #2570: Issue with pre taken exam
When you take exam that you have already taken. And you jump questions and submit. You will see **Continue Exam** And later even if you complete whole set of questions, it still gives you **Continue Exam**
screen shot 2018-11-22 at 22 44 20
screen shot 2018-11-22 at 22 44 58

nadaa:
fixes #2570 

The learner can take the exam multiple times, and he is also able to move between questions without answering them, the issue is, if he reaches to the last question and submits the answer, the exam will be closed.
The solution here did the trick by taking track all answers, and not to close the exam until all answers were submitted.

It will be better to persist answers in the textbox of each question so learners can review their answers.


yuviii:
@nadaa not working 
 - i skipped few answers, got back and again press continue exam.
 It shows me the counter like though this step has only 3 questions.
@lmmrssa @paulbert when i press back, the answer i submitted before get removed, isn't it would be good if we keep the answer when user click continue exam that have beed submitted before.

screenshot 2018-11-27 23 54 05

PR #2588, opened 2018-11-27
Disable button if already joined in a meetup (fixes #2547)

### LINKED ISSUE #2547: Join count not changing: Meetups
How i tested:
 1. select the checkbox of any meetup
 2. show the count 1 in join
 3. from more menu in meetup i select join
4. the meetup is joined and join changed to leave but it still shows the count 1 in in join in toolbar.

Should be disabled when it is joined.
screenshot 2018-11-20 10 12 13
screenshot 2018-11-20 10 13 48

lucy811:


lucy811:
this issue is similar with #2430 

How I tested:
1: join angular hangout and happy hacking hangout.

2: after step 1, there are two situations:
2.1 if you only select angular hangout, toolbar join will disabled. toolbar join show join(0)


2.2 if you select some courses, for example, select angular hangout and demo, the toolbar join will enable, show join(1), because demo hasn’t joined yet in step 1.

paulbert:
@singharpita Working for me, too.  Could you take another look?

PR #2606, opened 2018-11-28
Some kind of text message of button: suggestions(fixes #2600)

### LINKED ISSUE #2600: Some kind of text message of button: suggestions
I think it would be better to show some text message for not having any question rather than blank view toggle after clicking it.

screenshot 2018-11-28 08 35 56

lucy811:


svlesiv:
Hi @lucy811, also it could be nice to use `resource` instead of `resources` and `question` instead of `questions` when we have only 0 or 1  question/resource.

lucy811:
@svlesiv Improve it according to your comment. Thanks.

lucy811:
add situation that those steps which has description ,exams and resource.
qwer

lucy811:
@paulbert

PR #2608, opened 2018-11-29
Fix exam question validation (fixes #2594)

### LINKED ISSUE #2594: Course exam: no validation on single multiple choice(single & multiple answer)
screenshot from 2018-11-28 07-22-40

paulbert:
Currently validation works as expected, but I need to implement better changes to the UI so the person creating an exam can see what questions are invalid.

#2594

### DIFF
--- a/src/app/exams/exams-add.component.html
+++ b/src/app/exams/exams-add.component.html
@@ -25,8 +25,9 @@
         </mat-form-field>
       </div>
       <planet-step-list class="exam-questions" [steps]="examForm.controls.questions.value" [nameProp]="'title'" (stepClicked)="stepClick($event)">
-        <planet-step-list-item *ngFor="let question of examForm.controls.questions.value; index as i">
-          <span>{{question.title || 'Question ' + (i+1)}}</span>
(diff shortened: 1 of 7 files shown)


paulbert:
Refactors the exam questions so the validation works again.  The fields that are required are:

- Question Body
- Each question choice (if any) must have a value
- If the exam is for courses (not survey) at least one correct choice must be selected

Now there is also a message which will popup on any question which has an error as well to direct course creators to the issues.

svlesiv:
`delete` and `move` buttons don't work for me also when questions are valid or not valid. But I'm not getting console error.

paulbert:
@singharpita I'm not able to replicate that console error, either.  Could you send some more detail on the exam you are updating to create that error?

I fixed delete & move, but please test again and look out for an intermittent errors I noticed that I'm not sure I fully fixed:

Sometimes when changing the order of the last question the question values change so, for example rather than
```
1, 2, 3, 4 => 1, 2, 4, 3
```
it would change to
```
1, 2, 3, 4 => 1, 2, 4, 4
```

After this happens adding a question would sometimes produce a similar error where the new question already has a value either of an existing question or one that was previously deleted.

Please check to see if you can replicate the above errors.

PR #2631, opened 2018-12-02
Update members profile heading

### LINKED ISSUE #2613: Users/Members: different headings for list,view and edit pages
for showing list of users the heading given is "Members".
screenshot from 2018-11-29 20-36-36

next if we go to the particular member i.e Detail page, heading shows "Users".
Detail page heading could be "Member Profile"
screenshot from 2018-11-29 20-36-48

When we edit the user, heading is "Profile".
Edit user heading could be "Edit Member Profile"

screenshot from 2018-11-29 20-38-00

nadaa:
Fixes #2613


paulbert:
@nadaa Please don't use the keyword "Fixes #..." in the comments, just put the issue # in the comment.  Fixes automatically closes the issue, but sometimes we want to keep issues open if we think there's more to do on them.

Thanks!

PR #2633, opened 2018-12-03
[#2611] CourseProgress: Prevent showing of the button during loading time

### LINKED ISSUE #2611: Course-your progress: Show progress button shown until the page loads
show progress button should not be loaded when page load.
It will disappear after the page loads or course progress is shown.

svlesiv:
fixes #2611

## Description
I removed initialization of `selectedStep` variable inside ngOnInit(). Thus, "Show full course" button is not visible during loading. 
Not sure if it is a good practice to remove initialization from ngOnInit(). But it seems that the variable is initialized successfully inside functions of the component.


svlesiv:
It doesn't show when I load the page. Thanks for the review

PR #2638, opened 2018-12-04
Add new label (fixes #2501)

### LINKED ISSUE #2501: Popup to create new label

paulbert:
Allows adding new labels from the dialog in the resource page.

Please test to make sure you can create a label and assign it to resources.

Next TODOs:
- Adding a label which is a "sublabel" or attached to other labels doesn't do anything at this moment.  Need to show this in the filter view.
- Tags on the resource list show the id for a split second before changing to the text.  Need to find ways to improve this.
- A manager list view of labels where they can edit them after they have been created.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,7 +94,7 @@ jobs:
       script:
         - set -e
         - echo  "Waiting for couchdb to start"; WAIT_TIME=0; until curl http://127.0.0.1:5984/_all_dbs || [ $WAIT_TIME -eq 180 ]; do echo "..."; sleep 5; WAIT_TIME=$(expr $WAIT_TIME + 5); done
-        - i=$(curl -X GET http://127.0.0.1:5984/_all_dbs | jq length); if [ $i -ne 26 ]; then exit 1; fi
+        - i=$(curl -X GET http://127.0.0.1:5984/_all_dbs | jq length); if [ $i -ne 27 ]; then exit 1; fi
         - ng e2e
     - stage: docker-release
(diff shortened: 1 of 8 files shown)


lmmrssa:
Creating new tag is not adding it to resource immediately. It closes the dialog as well so user has to click again to add tag.
Also adding tag to resource is adding ID of tag as shown in image below
screen shot 2018-12-04 at 18 26 21

lmmrssa:
Add label is not selecting old tags so when submitted it removes old tags if not selected.
screen shot 2018-12-04 at 18 27 03

paulbert:
@lmmrssa 

1. Yes that was another TODO that I forgot.
2. Could you send some more info about the resource which does not have its old tags selected?  On my machine all tags remain selected when opening the dialog.

lmmrssa:
Regarding point 2. It seems like label is added but it was the issue of checkbox not being selected that when you select and unselect checkbox it removes label and you wouldn't know have you removed any previous label or not as all previous label were not already checked.

Label already added
screen shot 2018-12-04 at 20 54 46
Add label popup without old label checked
screen shot 2018-12-04 at 20 58 39
Checking new label confusing as its not clear which labels are there in resource right now
screen shot 2018-12-04 at 20 54 36

lmmrssa:
Label auto selection is working now

paulbert:
@lmmrssa I don't think the video not showing is related as videos work locally for me.  We can discuss at the meeting later.

Issue #2650, opened 2018-12-05
Password visibility


lmmrssa:
Can we also have password visibility toggle on change password
![screen shot 2018-12-05 at 07 48 56](https://user-images.githubusercontent.com/2909490/49485285-7a787c00-f862-11e8-8c59-d14826a3ebae.png)

PR #2658, opened 2018-12-05
[#2650] change password toggle visibility (connects #2650)

singharpita:


src/app/shared/dialogs/change-password.directive.ts
```diff
@@ -11,32 +11,33 @@ import { CustomValidators } from '../../validators/custom-validators';
...
-  {
-    'label': 'Confirm Password',
-    'type': 'textbox',
-    'inputType': 'password',
-    'name': 'confirmPassword',
-    'placeholder': 'Confirm Password',
-    'required': true
-  }
-];
+function getFormFields() {
```
paulbert: @singharpita This doesn't need to be changed to a function.  Could you revert this back to a variable?

lmmrssa: @paulbert dialog is updating the variable on its close with last state change making visibility of password field to be visible when we open change password again. So either we will have to move const into opendialog function or we need to reinitialize variable. That's why it is converted to function to reset the variable

paulbert: That's because you are modifying the `field` object directly in `DialogsFormComponent` when toggling the password view.  I don't think that's a great idea, this object should not be modified, properties on the component should be modified instead.

src/app/shared/dialogs/dialogs-form.component.html
```diff
@@ -4,7 +4,10 @@ <h1 mat-dialog-title><span><mat-icon>create</mat-icon>{{title}}</span></h1>
     <div *ngFor="let field of fields">
       <!-- input field -->
       <mat-form-field *ngIf="field.type === 'textbox'" class="full-width">
-        <input matInput placeholder="{{field.placeholder}}" type="{{field.inputType || 'text'}}" formControlName="{{field.name}}" required="{{field.required}}">
+        <input matInput placeholder="{{field.placeholder}}" type="{{(field.togglePassword && field.inputType == 'password') ? 'password' : 'text'}}" formControlName="{{field.name}}" required="{{field.required}}">
+          <button *ngIf="field.inputType" type="button" mat-icon-button matSuffix (click)="field.togglePassword = !field.togglePassword">
```
paulbert: This shows the toggle password icon button for any `<input>` field in the form.  It should only be for passwords.

singharpita:
@paulbert @lmmrssa Please review changes

PR #2660, opened 2018-12-05
Fix label filter (fixes #2652)

### LINKED ISSUE #2652: filter label: add resources
creating the error when filtering the labels.
screenshot 2018-12-05 08 14 44

paulbert:
Had referenced the incorrect property of the tag object in the filtering function.

To test try typing into the "Filter labels" input and see if the label list is filtered.  Also there should not be errors in the console as you type.

### DIFF
--- a/src/app/shared/forms/tags.service.ts
+++ b/src/app/shared/forms/tags.service.ts
@@ -31,7 +31,7 @@ export class TagsService {
   }
 
   filterTags(tags: any[], filterString: string): string[] {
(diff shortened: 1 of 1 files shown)


svlesiv:
Hi @paulbert, in production I'm getting the same error which you're trying to fix:

screen shot 2018-12-05 at 5 21 38 pm

Unfortunately, I cannot test your fix in dev. I cannot create a tag or filter tags. I'm receiving the next errors in a console:
screen shot 2018-12-05 at 5 28 51 pm

Issue #2662, opened 2018-12-05
Filtering labels: hide labels with no more results


paulbert:
In the dialog to filter by label we should update the count of items that would be selected & hide labels that would result in 0 items shown.

Issue #2661, opened 2018-12-05
Label/collections management page


paulbert:
Need to have a page to edit labels to update names & linking.

PR #2663, opened 2018-12-05
Resources: Filter with sublabels

paulbert:
To test: create a new label and select some values for "Sublabel of...".  Add this label & the parent label to one or many resources.  Back on the resource view click "Filter labels" and you will see a similar view to the below image.  Clicking on a label with an L arrow will filter by both that label and the parent above it.

Note there are still a few more steps for the labels which relate here
#2661 - Would allow us to attach already created labels to new labels
#2662 - Right now if the label & sublabel have no matching resources you could click the sublabel and see no resources


paulbert:
Also added a commit for sublabels for the multi-select so when adding to a resource you can just click the sublabel and then it will select the connected parent.  Note if a label is a sublabel of multiple parents it will only select the one you clicked under.  Clicking on the sublabel again will only deselect the sublabel, so you will have to click the parent label again if you want to deselect that as well.

lmmrssa:
@paulbert Yes there still seems that few things need to be fixed. But it looks really nice that sublabel filter is working

PR #2664, opened 2018-12-06
Add community grouping

paulbert:


src/app/community/community-table.component.ts
```diff
@@ -0,0 +1,190 @@
...
+        height: '500px',
+        width: '600px',
+        autoFocus: false
+      });
+    });
+  }
+
+  addHubClick(planetCode, hubName) {
+    const hub = this.hubs.find((hb: any) => hb.name === hubName);
+    hub.attached.push(planetCode);
```
lmmrssa: looks like its pushing community to attached of hub but not updating other hub. So its resulting to have different **attachedTo** and **children**

paulbert: Oh, the problem is that `children` shouldn't be saved to the DB, it's just for display. If the community is updated we want to show up to date data so that data is fetched every time we load the table.  `attached` is the only field that should be stored.

paulbert:
Fixed.  The only fields saved to the database are `name` and `attached`.

Issue #2680, opened 2018-12-06
Exam texteditor: image as answer submission


singharpita:
is it possible to submit the answer of an exam as an image?
because in texteditor for the type Textarea, image option has been provided.

but in submission its not shown.
![screenshot from 2018-12-06 21-04-42](https://user-images.githubusercontent.com/27276132/49593104-e5c56980-f99a-11e8-8525-56e85d944977.png)

same for the survey as well.

PR #2686, opened 2018-12-07
[#2680] Hide add image (connects #2680)

singharpita:


yuviii:
@singharpita I think its good idea to hide the image option. I guess it would be better to remove image icon from rest as well. As adding the image in step update and new step yields the following output now. 
screenshot 2018-12-10 23 30 22

**After adding the image** 

screenshot 2018-12-10 23 29 56

screenshot 2018-12-10 23 27 46

singharpita:
@paulbert @yuviii There is no image option showing on editor. Could you please make sure you tested on this branch or 
image

PR #2694, opened 2018-12-10
Username/Pswd error message fadeout(fixes # 2566)

BhargaviNadendla:


paulbert:
This makes both message service methods identical.  The alert should stay on screen longer.

Issue #2679, opened 2018-12-06
Submission survey detail enhancement


singharpita:
In survey detail, if "Submitted by", "Status" and "updated on" label is included like others page (resources) it would be great for user to have whole list of information in one page i.e submission survey detail page.

![screenshot from 2018-12-06 20-44-12](https://user-images.githubusercontent.com/27276132/49592462-42c02000-f999-11e8-8747-f4dbfd413930.png)
![screenshot from 2018-12-06 20-44-46](https://user-images.githubusercontent.com/27276132/49592468-46ec3d80-f999-11e8-91d3-05ac4d25ca3d.png)

PR #2699, opened 2018-12-11
[#2679] Survey detail enhancement (connects #2679)

singharpita:


src/app/exams/exams-view.component.html
```diff
@@ -12,56 +12,63 @@
...
+          [disabled]="answer === undefined || answer === '' || grade === undefined || grade === null"
+          i18n>
+          Submit Answer
+        </button>
+        <span class="small warn-text-color v-align-center" *ngIf="incorrectAnswer">
+          <mat-icon class="margin-lr-3">error</mat-icon><span i18n>Incorrect answer, please try again</span>
+        </span>
+      </div>
     </div>
+    <div *ngIf="examType == 'surveys'" class="right-tile">
```
paulbert: We should not have this information showing when a person is taking a survey.

I think showing Submitted by & Submitted on is good for both survey & exam review (`mode === 'grade'), but we can place it in the toolbar next to the question number.

Issue #2701, opened 2018-12-11
CoursesView: Hide/Show Details button


paulbert:
Like on ResourcesView, we should have a hide/show details button which toggle the grey detail section.  It should start at hidden when the page is loaded.

PR #2704, opened 2018-12-11
[2701]-show-hide-course-view (connects #2701)

yuviii:


yuviii:
@paulbert change request is fixed.

PR #2705, opened 2018-12-11
Add versions endpoint with APK versions (fixes #2688)

### LINKED ISSUE #2688: Meta for myplanet integration
We need to add last version of myplanet that planet supports (tested)
Also we need to mention which version of planet we do not support anymore

paulbert:


paulbert:
To test: try on production `localhost/versions`

Should see a json output

paulbert:
Live here: https://paulbert.media.mit.edu/versions

Issue #2748, opened 2018-12-13
Team detail click member name to see profile


singharpita:
should we add link to the members?
since user might want to visit the member profile.
![screenshot from 2018-12-13 21-35-00](https://user-images.githubusercontent.com/27276132/49950061-33a01b80-ff1f-11e8-800b-64786043f5e1.png)

PR #2757, opened 2018-12-14
[2748]-user-link-added-mmetup(connects #2748)

yuviii:


src/app/meetups/view-meetups/meetups-view.component.html
```diff
@@ -34,8 +34,10 @@
       <ng-template #memberList>
         <ul>
           <li *ngFor="let mem of members">
-            <span *ngIf="mem === currentUserName; else otherMember">{{mem}} (You)</span>
-            <ng-template #otherMember><span>{{mem}}</span></ng-template>
+            <a routerLink="/manager/users/profile/{{mem}}">
```
lmmrssa: we should try to not use manager route as even non admin should be looking into profile. manager route is for admin only.

yuviii:
@lmmrssa change request done, please review the latest commit.

PR #2766, opened 2018-12-14
PlanetStepList: Include next & previous buttons (fixes #2702)

### LINKED ISSUE #2702: PlanetStepList: Include next & previous buttons
When viewing a particular step (within the `<ng-container *ngIf="!stepMode">...` element), we should also include a previous step (if not the first step), and a next step that changes to add new step when on the last step button.

These should appear to the right of the "Return to List" button.

lucy811:


lucy811:
I added the next & previous  buttons are shown below:

paulbert:
Also these buttons should be on the reusable component, PlanetStepListComponent.

lucy811:
@paulbert 
1) when on the last step, I don’t know how to only modify PlanetStepListComponent to create a new step.
2）In my latest code, question-list only work on when examForm is not valid. if the examForm is valid, when click next and previous button, sometimes the form will submit automatically. I don’t know why.

paulbert:
After seeing it in action, I think having the "Next" button add a new step is a bad idea, so I removed that functionality.

Issue #2709, opened 2018-12-12
View user profile from child on nation


lmmrssa:
Nation should be able to look into user profile of child communities. Users of community are on `child_users` database

PR #2768, opened 2018-12-17
[#2709] List users of child communities (connects #2709)

singharpita:


singharpita:
@paulbert can you please test it again. Have made some changes

Issue #2765, opened 2018-12-14
Reports: Visits by member & resource visits recent instead of total


paulbert:
We should do last 30 days in reports since it is more relevant for an overview than all time total.

PR #2769, opened 2018-12-17
[#2765] latest report (connects #2765)

singharpita:


src/app/manager-dashboard/manager.service.ts
```diff
@@ -77,7 +77,7 @@ export class ManagerService {
     return this.couchService.post('admin_activities', { ...log, type });
   }
 
-  getLogs() {
+  getLogs(tillDate?: number) {
```
paulbert: You've added this parameter to the function but it's never used.  I think this is not complete.

PR #2772, opened 2018-12-17
Feedback search by status (fixes #2750)

### LINKED ISSUE #2750: Feedback search by 'status'
It would be better if we could search the feedback by status since there will be many instances of it at the Admin side.

image

lucy811:


lucy811:
@paulbert Added a feedback status dropdown, but don't know how to let Open show both "Open" and "Reopened" feedback.

Issue #2805, opened 2018-12-20
Leave selected is highlighted with count though it is not joined.


yuviii:
<img width="1424" alt="screenshot 2018-12-20 20 39 20" src="https://user-images.githubusercontent.com/6964191/50291806-a49f8000-0497-11e9-82b2-0204c35405a3.png">

lmmrssa:
I dont see this happening. It looks like click is wrong. If it was not joined then Join needed to highlight.
There is another issue with on **Leave**
When **7** courses are selected count on Leave is showing **1** which is correct as there is only one course user has joined.
![screen shot 2018-12-20 at 20 49 33](https://user-images.githubusercontent.com/2909490/50292489-54c1b880-0499-11e9-88cc-a37ec7d36355.png)
But after clicking Leave it shows message that **7 courses are removed** which is not correct.
![screen shot 2018-12-20 at 20 49 41](https://user-images.githubusercontent.com/2909490/50292490-555a4f00-0499-11e9-8ade-6083466a101e.png)

paulbert:
This is the case when the course has no steps, which I forgot about for the PR.  Need to exclude courses with no steps from the counts & exclude them from join/leave when clicking the button.

PR #2819, opened 2018-12-23
[#2805] leave count on course with no step (connects #2805)

singharpita:


BhargaviNadendla:
@lmrssa @singharpita When working on resources multi-select and "remove from library" issue, I came across a bug in this recently merged pull request, when selecting the courses already joined, it is not enabling "leave"

image

Issue #2783, opened 2018-12-19
Choice text should be unique in multiple choice multiple answer and multiple choice single answer: AddExam


yuviii:
<img width="845" alt="screenshot 2018-12-19 07 59 47" src="https://user-images.githubusercontent.com/6964191/50194786-72413600-0364-11e9-95b4-23e0898faf2c.png">

PR #2820, opened 2018-12-23
[2783] duplicate-answer-validator-added(connects #2783)

yuviii:


src/app/validators/custom-validators.ts
```diff
@@ -31,6 +31,30 @@ export class CustomValidators {
     };
   }
 
+  static isUniqueAnswer(uniqueRequired: boolean) {
```
lmmrssa: Clean the code and also it is validating choices still exam form submits and there is no message showing. Please work on that

singharpita:
@yuviii could you please confirm fixes

yuviii:
Not relevant anymore

PR #2821, opened 2018-12-24
[2812] filter-added-in-notification(connects #2812)

yuviii:


singharpita:
@yuviii CI test is failing can you check it.

Issue #2887, opened 2019-01-02
Navigation Survey: When survey is opened from main dashboard clicking in back should redirect to dashboard.


yuviii:
Currently it is redirecting to manager page every time when opened.

PR #2892, opened 2019-01-02
[#2887] survey back navigation (connects #2887)

singharpita:


paulbert:
`parent` doesn't exist on any of the routes for the survey so this always leads to the dashboard.  I changed it to a relative route which gives the same result with fewer lines.

I think it's redundant to have this linked in the main navigation and in the Manager Dashboard, so I'm going to just remove the button.

Note relative routing is buggy with this version of Angular which means the `../` route change would not work if we also include the survey routes as children of the manager.  It may be working in `7.2+`, so we'll see when we upgrade.

PR #2895, opened 2019-01-02
Fix multiple community report overview

paulbert:


paulbert:
Error is found on the "Reports" page: no data for Resource Views, Member Visits, Last Admin Login, Last Upgrade, Last Sync because the json sent with the `_find` requests were invalid.

Issue #1732, opened 2018-07-24
Approval option for resource from community


lmmrssa:
There should be option that requires approval or auto approved for any resources, courses coming from community to nation.

i5o:
In my opinion there **MUST** be an approval to get resources. Mostly because all communities can access to these if they list the resources.

We could add something like "pre-approved communities to send resources" or so.
But I was surprised that resources were approved by default when we tested it

PR #2914, opened 2019-01-03
Send from child to parent (connects #1732)

lmmrssa:
@paulbert There are still some works to be done.  But like you to go through it and provide some feedback we it can be improved.


jessewashburn:
Not sure what this is about, but I doubt anyone wants to pick up where he left off on 16 files changed.

Issue #2880, opened 2019-01-02
meetup titlebar space before number


lmmrssa:
Add space between button label and number
![screen shot 2019-01-02 at 07 28 57](https://user-images.githubusercontent.com/2909490/50578305-52c80400-0e60-11e9-86d6-3ed3bb930407.png)

PR #2919, opened 2019-01-04
[#2880] add space between button and count in resource and meetup (connects #2880)

singharpita:
added for resource as well

### DIFF
--- a/src/app/resources/resources.component.html
+++ b/src/app/resources/resources.component.html
@@ -25,16 +25,16 @@
       </button>
       <button mat-button [disabled]="!selection.selected.length || selectedNotAdded === 0" (click)="libraryToggle(selection.selected, 'add')">
         <mat-icon aria-hidden="true" class="margin-lr-3">library_add</mat-icon><span i18n>Add to myLibrary</span>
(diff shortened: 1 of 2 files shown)


src/app/resources/resources.component.html
```diff
@@ -25,16 +25,16 @@
       </button>
       <button mat-button [disabled]="!selection.selected.length || selectedNotAdded === 0" (click)="libraryToggle(selection.selected, 'add')">
         <mat-icon aria-hidden="true" class="margin-lr-3">library_add</mat-icon><span i18n>Add to myLibrary</span>
-        <span *ngIf="selection?.selected?.length">({{selectedNotAdded}})</span>
+        <span *ngIf="selection?.selected?.length"> ({{selectedNotAdded}})</span>
```
yuviii: I think rather than adding the space it would be better to style with margin left by adding the class like
 `        <span *ngIf="selection?.selected?.length" class="margin-lr-3">({{selectedNotAdded}})</span>
`

paulbert: @yuviii For block elements or block like inline elements (like buttons), yes, but with inline elements of text spaces are better.

PR #2924, opened 2019-01-04
Collection List : case insensitive in collection text (fixes #2905)

### LINKED ISSUE #2905: Collection List : case insensitive in collection text
The collection text should be case insensitive. Currently it is allowing to add the text like 'test' and 'Test'

lucy811:


lucy811:
I am not sure where Lowercase Directive needs to be moved.

PR #2977, opened 2019-01-15
Add home link to top navigation (fixes #2926)

### LINKED ISSUE #2926: Add "Home" link to navigation bar
We think it will help usability for people if there is a "Home" link that leads to the dashboard as the first link in the navigation bar.  Please make this change for the wider, text version of the navigation bar only.

paulbert:
#2926

### DIFF
--- a/src/app/home/home.component.html
+++ b/src/app/home/home.component.html
@@ -7,6 +7,9 @@
     </a>
   </span>
   <span class="navbar-center" *ngIf="layout === 'classic' && !forceModern">
+    <a mat-button routerLink="/"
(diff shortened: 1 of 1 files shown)


src/app/home/home.component.html
```diff
@@ -7,6 +7,9 @@ <h1 i18n>Planet Learning</h1>
     </a>
   </span>
   <span class="navbar-center" *ngIf="layout === 'classic' && !forceModern">
+    <a mat-button routerLink="/"
```
lmmrssa: Need to fix title :)

Issue #2540, opened 2018-11-20
Password change issues: change password


yuviii:
Entering the wrong old password it prevents from changing the password, but eventually displays wrong message box #2053
 
<img width="1440" alt="screenshot 2018-11-20 08 23 43" src="https://user-images.githubusercontent.com/6964191/48748161-f50b9e00-ec9d-11e8-82a3-632215459f31.png">
<img width="1440" alt="screenshot 2018-11-20 08 23 49" src="https://user-images.githubusercontent.com/6964191/48748184-09e83180-ec9e-11e8-9f33-0731d4b34deb.png">

paulbert:
Related to #1527

Issue #2053, opened 2018-09-06
Change password with a wrong old password


BhargaviNadendla:
Though given a wrong old password, it says the new password is successfully updated though it is not. When tried to login again, it is only the old password that works

![image](https://user-images.githubusercontent.com/25775298/45167378-ae3dfa00-b1be-11e8-99c4-fce1226519e2.png)

![image](https://user-images.githubusercontent.com/25775298/45167554-fc52fd80-b1be-11e8-8bd2-662012bc8b8b.png)

jiajunc:
I don't have this problem with wrong old password. But I cannot change it to a new password when I type in the correct old password.

jessewashburn:
This has been fixed it looks like. Cannot replicate.

Mutugiii:
Cannot replicate

PR #3042, opened 2019-01-25
Fix lack of errors from couchdb

paulbert:
We had been catching all errors from CouchDB which meant in cases (like incorrect passwords) there was no error in the app when a request failed.

This fixes that and will actually fix the validator for #2540 and #2053


lmmrssa:
screen shot 2019-01-28 at 15 57 17

PR #3043, opened 2019-01-25
Refactor DialogsForm (fixes #1527)

### LINKED ISSUE #1527: Refactor DialogsFormComponent & DialogsFormService

paulbert:
Changes DialogsForm so the dialog does not close automatically when the form is submitted.

This allows us to receive a response from the server before closing the form in case of errors.

Implemented for changing passwords so far, will need to implement for other forms as well.  Note the change password error requires #3042 so I will put this on hold.

#1527

### DIFF
--- a/src/app/shared/dialogs/change-password.directive.ts
+++ b/src/app/shared/dialogs/change-password.directive.ts
@@ -1,15 +1,15 @@
 import { Directive, HostListener, Input } from '@angular/core';
 import { forkJoin, of } from 'rxjs';
-import { switchMap, catchError } from 'rxjs/operators';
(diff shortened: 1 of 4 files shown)


paulbert:
I implemented this for the `change-password` directive, but it's in there as a fallback since the validator will check if the password is correct or not.  Oddly even though the same requests are made, the validator does not invalidate the session but if I remove the validator submitting the wrong password will invalidate the session.

As it stands this will keep the dialog open if the HTTP request to change the password fails, which is a step in the right direction.

PR #3069, opened 2019-01-30
Submission count(exams not yet graded) badge on dashboard(fixes #3025)

### LINKED ISSUE #3025: Submissions: notification
It could be nice to have a notification for submissions similarly to what we have for the surveys.

screen shot 2019-01-22 at 6 28 17 pm

BhargaviNadendla:
@paulbert,  As suggested by Laxman, the exams which are not graded, yet complete are taken into consideration for the badge. Please let me know if there are changes or additions to be made.

### DIFF
--- a/src/app/dashboard/dashboard.component.ts
+++ b/src/app/dashboard/dashboard.component.ts
@@ -36,9 +36,10 @@ export class DashboardComponent implements OnInit {
   }
 
   ngOnInit() {
-    const userShelf = this.userService.shelf;
+
     this.getSurveys();
     this.getExams();
+    this.initDashboard();
 
     this.couchService.post('login_activities/_find', findDocuments({ 'user': this.userService.get().name }, [ 'user' ], [], 1000))
       .pipe(
@@ -49,6 +50,11 @@ export class DashboardComponent implements OnInit {
         this.visits = res.docs.length;
       });
 
+  }
+
+  initDashboard() {
+
+    const userShelf = this.userService.shelf;
     if (this.isEmptyShelf(userShelf)) {
       this.data = { resources: [], courses: [], meetups: [], myTeams: [] };
       return;
--- a/src/app/dashboard/dashboard.component.ts
+++ b/src/app/dashboard/dashboard.component.ts
@@ -21,6 +21,7 @@ export class DashboardComponent implements OnInit {
   dateNow: any;
   visits = 0;
   surveysCount = 0;
+  exams = [];
 
   constructor(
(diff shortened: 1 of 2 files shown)


src/app/dashboard/dashboard.component.ts
```diff
@@ -92,13 +93,20 @@ export class DashboardComponent implements OnInit {
...
 
-  getSurveys() {
-    this.submissionsService.getSubmissions(findDocuments({
-      'user.name': this.userService.get().name,
-      type: 'survey',
-       status: 'pending'
-    })).subscribe((surveys) => {
-      this.surveys = surveys;
+  getSurveysSubmissions() {
+    forkJoin([
```
paulbert: There's no need to put these requests together in one function or `forkJoin` observable.  You should split this up into two functions which will also allow for clearer variable and function names.

src/app/dashboard/dashboard.component.ts
```diff
@@ -92,13 +93,20 @@ export class DashboardComponent implements OnInit {
...
+      this.submissionsService.getSubmissions(findDocuments({
+        'user.name': this.userService.get().name,
+        type: 'survey',
+        status: 'pending'
+      })),
+      this.submissionsService.getSubmissions(findDocuments({
+        type: 'exam',
+        status: 'complete'
+      }))
+    ]).subscribe(results => {
```
paulbert: When we know the result of a request it's better to use a descriptive name rather than a generic `result` (i.e. `surveys` and `exams`)

BhargaviNadendla:
@paulbert @lmmrssa Updated the badge to be shown only for admins

src/app/dashboard/dashboard.component.ts
```diff
@@ -96,9 +99,18 @@ export class DashboardComponent implements OnInit {
...
       'user.name': this.userService.get().name,
       type: 'survey',
-       status: 'pending'
+      status: 'pending'
     })).subscribe((surveys) => {
       this.surveys = surveys;
     });
   }
+
+  getExams() {
```
paulbert: This is a near duplicate of the `getSurveys()` function, and it is not querying for exams.  You should make a couple of changes here:

1. The `type` in the query needs to be 'exams' when querying for exams.  `status` should be `complete` because this is the status for an exam that is ready to be graded.
2. Please try to refactor `getExams()` and `getSubmissions()` into one function that is called once for exams and once for surveys in `ngOnInit()`.  Note that surveys should only query for submissions that match the username

Issue #3057, opened 2019-01-29
Achievements: When user hasn't added anything should show message


paulbert:
When there is no achievements doc for a user the screen is blank and an error message shows.

We should catch the error and if it is due to document not found we should:

* Gracefully continue the observable sequence without throwing an error so the error message does not show up.
* Set the component to show "You haven't entered any achievements yet. Click Edit Achievements to add your achievements!"

![image](https://user-images.githubusercontent.com/9203229/51947806-261df280-23f4-11e9-9af4-66a35b433dd5.png)

PR #3077, opened 2019-01-31
[3057]-empty-achievement-managed (connects #3057)

yuviii:


src/app/users/users-achievements/users-achievements.component.html
```diff
@@ -41,5 +41,6 @@ <h4 *ngIf="achievements[infoType]">{{infoType}}</h4>
         </mat-list>
       </div>
     </div>
+    <div *ngIf="!achievements">You haven't entered any achievements yet. Click Edit Achievements to add your achievements!</div>
```
paulbert: On a slow connection this message will show while the request to get the achievements doc completes.  We should not show the message until after the request completes.

src/app/users/users-achievements/users-achievements.component.ts
```diff
@@ -42,7 +42,6 @@ export class UsersAchievementsComponent implements OnInit {
         )
       });
     }, (error) => {
-      this.planetMessageService.showAlert('There was an error getting achievements');
```
paulbert: We should not remove the error message entirely.  If the error is for something other than the document does not exist we should show that.

yuviii:
@paulbert @lmmrssa please review the update

PR #3099, opened 2019-02-01
Fix for sync when _users have _attachments (fixes #3097)

### LINKED ISSUE #3097: Sync failing when any user has profile image

paulbert:
Removes profile images from member information syncing to nation.


singharpita:
works fine when the role is admin and has profile image.
But when the role is manager and irrespective of profile image, Run Sync shows [Object] [Object].
screenshot from 2019-02-03 20-55-02

Issue #3087, opened 2019-01-31
Achievements: There is no way to see the full description if a description is long.


yuviii:
<img width="1438" alt="screenshot 2019-01-31 20 36 51" src="https://user-images.githubusercontent.com/6964191/52062432-98c7d300-2598-11e9-9df0-7ff4e1aeb2cc.png">

paulbert:
https://stackoverflow.com/questions/42023559/angular-2-material-show-full-text-for-the-list-items

I think this change will fix it in a way that still looks good.  Note we should make sure this CSS change is limited to this component.

PR #3102, opened 2019-02-03
[3087]long-text-managed-matline(connects #3087)

yuviii:


lmmrssa:
This is working but just adding
```
    .mat-list-item {
      height: initial !important;
    }
    .mat-line {
      white-space: normal !important;
    }
```
is enough. First style adds spacing and second removes wrapping.

PR #3125, opened 2019-02-06
FeedbackView: change date format (fixes #3124)

### LINKED ISSUE #3124: FeedbackView: Use generic date pipe
In FeedbackView we should use a generic date type for the pipe so the date format matches the locale used in the deployment.

svlesiv:
screen shot 2019-02-06 at 12 49 06 pm

### DIFF
--- a/src/app/feedback/feedback-view.component.html
+++ b/src/app/feedback/feedback-view.component.html
@@ -43,7 +43,7 @@
     <div class="chat-list" #chatList>
       <div class="chat-list-item" *ngFor="let message of feedback.messages; trackBy: feedbackTrackByFn">
         <div class="chat-list-user">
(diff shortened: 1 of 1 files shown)


svlesiv:
If exact time doesn't matter, we could use 'mediumDate' instead of 'medium'. 
Please let me know if I understood this issue correctly and no farther altering of Time Zones needed.

yuviii:
@svlesiv yeah you are right
'medium': equivalent to 'MMM d, y, h:mm:ss a' (Jun 15, 2015, 9:03:01 AM). and
'mediumDate': equivalent to 'MMM d, y' (Jun 15, 2015).

[Date pipe angular documentation](https://angular.io/api/common/DatePipe)

PR #3149, opened 2019-02-08
AchievementsUpdate edits (fixes #3147)

### LINKED ISSUE #3147: AchievementsUpdate edits
Feedback from RRR
- Modify some button text
- Move buttons to bottom of list
- Add summaries of the different sections to explain what they are

paulbert:


yuviii:
@paulbert 
I think it would be better to add the title in the achievement list as well like Personal informations and instead of `Enter an achievement` it would be good to label as `Add an achievement`.

I am approving this pr because it is working as it needs to.

screenshot 2019-02-09 18 03 14

PR #3151, opened 2019-02-08
Update README.md

YuserahN:
To access localhost:3000, you need to run 'vagrant dev up' rather than 'vagrant up'.


lmmrssa:
@YuserahN yes please change all vagrant commands.

YuserahN:
I've made the suggested changes.

Issue #3142, opened 2019-02-07
Achievements: View child achievements


paulbert:
After syncing from a community the link to view an achievement on the nation does not work.

For child users the link to view the profile includes the full id with the community code in it (i.e. user **paul** on community **pb** would have the id in `child_users` **paul@pb**).  In the component the `@parentCode` is being added again for the queries, which is causing the incorrect id to be queried for with child users (see request in image below).

![image](https://user-images.githubusercontent.com/9203229/52421354-e5903a00-2ac1-11e9-8947-0ad4073a3e14.png)

paulbert:
And to specify, let's change the `AchievementsComponent` to handle this condition.  If the `name` parameter already has an `@` symbol in it we should not add it again.

lmmrssa:
@paulbert  I was able to see user lax@lprod achievement on dev.

paulbert:
This problem only arises when navigating to the user's achievements through the "Associated" list of members.  This includes the admins from the various children attached to the planet.  The admin users should not be used as a "learner", so should not have achievements in the field anyway.  In case they do get used for this purpose, we should still fix the link.

Based on this removing the "priority" label.

PR #3155, opened 2019-02-09
[#3142] view child achievement (connects #3142)

singharpita:


src/app/users/users-achievements/users-achievements.component.ts
```diff
@@ -67,9 +67,10 @@ export class UsersAchievementsComponent implements OnInit {
   }
 
   initUser(name, planetCode) {
+    const achievementName = name.indexOf('@') > -1 ? name : name + '@' + planetCode;
```
yuviii: Need to handle if the name is null. Cannot read property 'indexOf' of null

singharpita: @yuviii couldn't replicate the case of null name.
Can you please specify the steps or screenshot the indexOf null console error?

paulbert: @yuviii If name is null, this function will not be called (see if statement at line 36).  Could you provide more detail on how you saw that error?

Issue #3135, opened 2019-02-07
Achievement Add Resources: UI fixes


singharpita:
Can Edit, Delete and move button is aligned properly and next to Achievement title?
![screenshot from 2019-02-07 20-00-17](https://user-images.githubusercontent.com/27276132/52417144-4d22a580-2b13-11e9-81ef-9045ef931c2a.png)

singharpita:
Also spacing between last Achievement's Resource tag and Add Other Information.
![screenshot from 2019-02-07 20-08-19](https://user-images.githubusercontent.com/27276132/52417565-46486280-2b14-11e9-9ede-9accc01d6a57.png)

paulbert:
For the first, alignment question we can try vertically aligning these buttons to the top, but it may mess up the alignment when there is only one line of text.  I think it may be best to leave it aligned in the center.

For the second comment:
1. Let's make sure to not simply arbitrarily increase the spacing there.  If we change this it should match the spacing between the first achievement and the Add Achievement button
2. We can add a little more space between the list items (0.25rem) so the buttons no longer touch.

PR #3158, opened 2019-02-11
[#3135] achievement UI fixes (connects #3135)

singharpita:


src/app/users/users-achievements/users-achievements.scss
```diff
@@ -28,6 +28,7 @@
 
     .achievement-button {
       grid-area: btn;
+      margin: 0.25rem;
```
paulbert: This also adds left/right margin.  Please set that to 0 with `margin: 0.25rem 0;`

src/app/users/users-achievements/users-achievements-update.scss
```diff
@@ -12,6 +12,11 @@
     height: inherit;
     padding: 0.25rem 0;
     white-space: pre-wrap;
+    margin-bottom: 0.25rem;
+  }
+
+  planet-step-list .mat-list-base .mat-list-item:last-child {
+    padding: 0.25rem;
```
paulbert: There is already top & bottom padding on `planet-step-list .mat-list-base .mat-list-item`.  All this does is add left & right padding to the last `.mat-list-item`, which creates a horizontal toolbar (see image).  Please remove this style.

image

PR #3161, opened 2019-02-11
Achievements: change text alignment (fixes #3062)

### LINKED ISSUE #3062: Achievements: better formatting
It would be nice to have better formatting for lists and quotation.

screen shot 2019-01-29 at 5 30 12 pm

svlesiv:
Maybe we could change text alignment. In that way, we will eliminate problems with lists and quotations, and overall it might look better.

screen shot 2019-02-11 at 2 30 18 pm

vs 

screen shot 2019-02-11 at 2 29 51 pm


paulbert:
I think the center alignment and name at type are important for this resume like view, so I changed to align the lists only to the left and center the rest of the text.

PR #3162, opened 2019-02-11
Session: increase session timeout (fixes #3119)

### LINKED ISSUE #3119: Session time out when creating long courses, exam or surveys
If you are not navigating between pages for long time then even if you are performing some activities within a page your session expires.
Such as creating a survey, exam or courses with lot of details on it.

svlesiv:


lmmrssa:
@svlesiv This is not what we were planning as a solution.
@paulbert If I am correct we were trying to have a way to reactivate session if user is actively using the application. Also making sure when session is activated data user was filling still remains on that page.

PR #3177, opened 2019-02-13
(Fixes #3137) Changes Button to Edit Achievement Button

### LINKED ISSUE #3137: Achievements: change title edit to add achievement when achievement has not added.
I think it would be better the change the button label `edit achievement` to `Add achievement` when user has not entered any achievement. Also the icon + instead of pen icon in add like as in resources and courses.

YuserahN:
Fixes #3137 
Changes the Edit Achievements Button to Add Achievements when no achievements have been added.

### DIFF
--- a/src/app/users/users-achievements/users-achievements.component.html
+++ b/src/app/users/users-achievements/users-achievements.component.html
@@ -9,9 +9,9 @@
     <span>{{user.firstName.concat(' ', user.lastName).trim() || user.name}}</span>
     <span class="toolbar-fill"></span>
     <a mat-raised-button color="accent" i18n [routerLink]="['/myAchievements/update']"><mat-icon>mode_edit</mat-icon>
-	  <span *ngIf="achievementNotFound">Add Achievements</span>
-	  <span *ngIf="!achievementNotFound">Edit Achievements</span> 
-	</a>
+      <span *ngIf="achievementNotFound">Add Achievements</span>
+      <span *ngIf="!achievementNotFound">Edit Achievements</span>
+    </a>
   </mat-toolbar>
   <div class="view-container">
     <div *ngIf="achievements !== undefined" class="achievements-container">
--- a/src/app/users/users-achievements/users-achievements.component.html
+++ b/src/app/users/users-achievements/users-achievements.component.html
@@ -8,7 +8,10 @@
   <mat-toolbar class="primary-color font-size-1">
     <span>{{user.firstName.concat(' ', user.lastName).trim() || user.name}}</span>
     <span class="toolbar-fill"></span>
-    <a mat-raised-button color="accent" i18n [routerLink]="['/myAchievements/update']" *ngIf="ownAchievements"><mat-icon>mode_edit</mat-icon><span>Edit Achievements</span></a>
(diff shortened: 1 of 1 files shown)


src/app/users/users-achievements/users-achievements.component.html
```diff
@@ -8,7 +8,10 @@
   <mat-toolbar class="primary-color font-size-1">
     <span>{{user.firstName.concat(' ', user.lastName).trim() || user.name}}</span>
     <span class="toolbar-fill"></span>
-    <a mat-raised-button color="accent" i18n [routerLink]="['/myAchievements/update']" *ngIf="ownAchievements"><mat-icon>mode_edit</mat-icon><span>Edit Achievements</span></a>
+    <a mat-raised-button color="accent" i18n [routerLink]="['/myAchievements/update']"><mat-icon>mode_edit</mat-icon>
+		<span *ngIf="!ownAchievements">Add Achievements</span>
+		<span *ngIf="ownAchievements">Edit Achievements</span> 
+	</a>
```
paulbert: Indentation should be at 2 space increments with the spans inside the a element.

yuviii:
@YuserahN 
It supposed to show add achievement button if no achievement exists. Not changing the label.

screenshot 2019-02-15 14 16 15

src/app/users/users-achievements/users-achievements.component.html
```diff
@@ -8,7 +8,10 @@
   <mat-toolbar class="primary-color font-size-1">
     <span>{{user.firstName.concat(' ', user.lastName).trim() || user.name}}</span>
     <span class="toolbar-fill"></span>
-    <a mat-raised-button color="accent" i18n [routerLink]="['/myAchievements/update']" *ngIf="ownAchievements"><mat-icon>mode_edit</mat-icon><span>Edit Achievements</span></a>
+    <a mat-raised-button color="accent" i18n [routerLink]="['/myAchievements/update']"><mat-icon>mode_edit</mat-icon>
+	  <span *ngIf="!ownAchievements">Add Achievements</span>
```
paulbert: I missed this in my review, but the property to test for this is `achievementNotFound` instead of `ownAchievements`.

Also the span tags should be indented two spaces further than the `<a>` tag, and then ending `</a>` tag should be in line with the starting tag.

YuserahN: @paulbert I've made the changes.

src/app/users/users-achievements/users-achievements.component.html
YuserahN: @paulbert I've added indentation at 2 space increments. I am not sure why it won't render the same way when I push the commit.
 [indentation
](url)

paulbert: Ok, that looks good.  I'll push the commit so we can move it along, but we can troubleshoot this tonight.

paulbert: I found the issue.  Looks like there were some tab characters in the indentation.  Github shows that as one space, my editor shows them as two, and your editor shows them as four.

There should be an option in your editor to set the tab key to insert X number of spaces, so we can try to set that tonight.  Or if you want to try out VS code that will automatically have that set for this repository.

PR #3178, opened 2019-02-13
Viewing the selected resources in dialogs-form(fixes #3140)

### LINKED ISSUE #3140: Achievement: show the selected resources in dialog
I think it would be better to show something like `two resources selected` when resources is selected in the dialog.

screenshot 2019-02-07 20 53 30

BhargaviNadendla:


src/app/shared/dialogs/dialogs-form.component.html
```diff
@@ -50,6 +50,10 @@ <h1 mat-dialog-title><span><mat-icon>create</mat-icon>{{title}}</span></h1>
 
       <!-- Dialog Button -->
       <button type="button" mat-stroked-button *ngIf="field.type === 'dialog'" (click)="openDialog(field)">{{field.text}}</button>
+      <span *ngIf="field.type === 'dialog'" class="mat-caption margin-lr-3" [matTooltip]="tooltipText">
```
lmmrssa: right now it always shows resources irrespective of what you send. We should make it similar to https://github.com/open-learning-exchange/planet/blob/658223414ee4959ae82bd964def36c4889d43251/src/app/shared/dialogs/dialogs-list.component.html#L35
Atleast in current case **other** should not say resources.

BhargaviNadendla: @lmmrssa I couldn't find any other component using dialogs-form component that uses dialogs-list component at present. Also in case of other, if a number that is not 0 or 1, it would fall under 'other' category as per my understanding. Please correct me if I am wrong.

BhargaviNadendla:
@paulbert  I am unable to add any achievements right now as it is resulting in a 404 error. I will fix this once this issue is fixed

paulbert:
Closing since this will be irrelevant with #3786

PR #3179, opened 2019-02-14
Users: Only show accepted communities in children list (fixes #3168)

### LINKED ISSUE #3168: Child planet list should only show accepted planets
image

paulbert:
#3168


paulbert:
Since this menu only comes up on a nation with connected communities, you will need to test on a Planet configured as nation.

Issue #3189, opened 2019-02-14
Member with role "Learner" can update the Resource created by admin


singharpita:
right now member with role "Learner" is able to update the Resources created by admin.

paulbert:
We should restrict editing to the resource creator and the admins (users with `isUserAdmin === true`)

PR #3197, opened 2019-02-15
[#3189] member with learner role resstricted to manage (connects #3189)

singharpita:


yuviii:
@singharpita not able to see the manage option if the creator is non admin.

singharpita:
changes made to Resource db. 
added createdBy field in Resource.

PR #3202, opened 2019-02-15
(Fixes #3194) See Profile of User Submitting Exam or Survey

### LINKED ISSUE #3194: See profile of user submitting exam or survey
We currently show user name of person who submitted exam or survey. But there is no way to look for that user from submission section, if manager or leader wants to check for.
It will be nice to make username link that takes to user profile.
screen shot 2019-02-14 at 21 23 48

YuserahN:
Fixes #3194

### DIFF
--- a/src/app/submissions/submissions.component.html
+++ b/src/app/submissions/submissions.component.html
@@ -41,7 +41,7 @@
       <ng-container matColumnDef="user">
         <mat-header-cell *matHeaderCellDef mat-sort-header i18n>Submitted by</mat-header-cell>
         <mat-cell *matCellDef="let element" (click)="$event.stopPropagation()">
-          <span *ngIf="element.submittedBy !== ''; else unknownLabel" [routerLink]="['/users/profile/',element.submittedBy]">{{element.submittedBy}}</span>
+          <span *ngIf="element.submittedBy !== ''; else unknownLabel" [routerLink]="user !== '' ? ['/users/profile/',element.submittedBy]:[]">{{element.submittedBy}}</span>
           <ng-template #unknownLabel><span i18n>Unknown</span></ng-template>
         </mat-cell>
       </ng-container>
--- a/src/app/submissions/submissions.component.html
+++ b/src/app/submissions/submissions.component.html
@@ -41,7 +41,7 @@
       <ng-container matColumnDef="user">
         <mat-header-cell *matHeaderCellDef mat-sort-header i18n>Submitted by</mat-header-cell>
         <mat-cell *matCellDef="let element" (click)="$event.stopPropagation()">
-          <span *ngIf="element.submittedBy !== ''; else unknownLabel" [routerLink]="user !== '' ? ['/users/profile/',element.submittedBy]:[]">{{element.submittedBy}}</span>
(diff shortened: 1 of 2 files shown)


src/app/submissions/submissions.component.html
```diff
@@ -40,8 +40,8 @@
       </ng-container>
       <ng-container matColumnDef="user">
         <mat-header-cell *matHeaderCellDef mat-sort-header i18n>Submitted by</mat-header-cell>
-        <mat-cell *matCellDef="let element">
-          <span *ngIf="element.submittedBy !== ''; else unknownLabel">{{element.submittedBy}}</span>
+        <mat-cell *matCellDef="let element" (click)="$event.stopPropagation()">
+          <span *ngIf="element.submittedBy !== ''; else unknownLabel" [routerLink]="['/users/profile/',element.submittedBy]">{{element.submittedBy}}</span>
```
lmmrssa: Link needs to be submission.user.name if it exist. Else it will not require link.
Submission can be done by
* Submission by valid user logged in
* Submission by custom detail (user does not exist)
* Submission without user detail

Hence only link should be when there is `user.name`

src/app/submissions/submissions.component.html
```diff
@@ -52,7 +52,7 @@
         </mat-cell>
       </ng-container>
       <mat-header-row *matHeaderRowDef="displayedColumns" class="hide"></mat-header-row>
-      <mat-row *matRowDef="let row; columns: displayedColumns;" (click)="submissionAction(row)"></mat-row>
+      <mat-row *matRowDef="let row; columns: displayedColumns;" (click)="submissionAction(row)"></mat-row>
```
lmmrssa: also please remove extra space added here.

src/app/submissions/submissions.component.html
YuserahN: Alright. I'll check it out.

src/app/submissions/submissions.component.html
```diff
@@ -40,8 +40,8 @@
       </ng-container>
       <ng-container matColumnDef="user">
         <mat-header-cell *matHeaderCellDef mat-sort-header i18n>Submitted by</mat-header-cell>
-        <mat-cell *matCellDef="let element">
-          <span *ngIf="element.submittedBy !== ''; else unknownLabel">{{element.submittedBy}}</span>
+        <mat-cell *matCellDef="let element" (click)="$event.stopPropagation()">
+          <span *ngIf="element.submittedBy !== ''; else unknownLabel" [routerLink]="name !== '' ? ['/users/profile/',element.submittedBy]:[]">{{element.submittedBy}}</span>
```
paulbert: You should use `element.user.name` (in a mat-table, element is the value of each element of the array/iterable supplied to the table) and check it does not equal `undefined`.

This code will always end up as "true" because for the value of `name` it will check the component, see there is no property `name` and the final comparison will be `undefined !== ''`.

On the `user` property of each submission, `name` will be a string if the user has created a profile or undefined if there is no profile.

YuserahN: I've made the changes. When I ran it with ```user``` and ```name```, it seemed to have worked. In other words, I was only able to view the user profiles if the ```user !== ''```. Maybe this worked due to some loophole. Either way, I've used ```element.user.name``` this time and that works.

PR #3204, opened 2019-02-18
(Fixes #2743) Submissions: Clarify Grading Status

### LINKED ISSUE #2743: Submissions: Clarify grading status
Right now we have three stages for an exam and two for a survey
Exams: pending -> complete -> graded
Survey: pending -> complete

I think the exam stages should be clearer, especially the second because it requires an action from a leader.  Let's change them to:
```
Exams: in progress -> requires grading -> complete
Survey: in progress -> complete
```

This will require a change to the code where we set the status that will need to check if (a) it is an exam and (b) it has ungraded questions to set the status to "requires grading"

YuserahN:
Fixes #2743

### DIFF
--- a/src/app/submissions/submissions.component.html
+++ b/src/app/submissions/submissions.component.html
@@ -31,12 +31,10 @@
       </ng-container>
       <ng-container matColumnDef="status">
         <mat-header-cell *matHeaderCellDef mat-sort-header i18n>Status</mat-header-cell>
-        <mat-cell *matCellDef="let element">
-          <mat-chip-list *ngIf="this.filter.type === 'exam' && element.status === 'complete'"><mat-chip selected color="accent" i18n>complete</mat-chip></mat-chip-list>
-          <mat-chip-list *ngIf="this.filter.type === 'survey' && element.status === 'complete'"><mat-chip selected il8n>complete</mat-chip></mat-chip-list>
-          <mat-chip-list *ngIf="element.status ==='graded'"><mat-chip selected i18n>graded</mat-chip></mat-chip-list>
-          <mat-chip-list *ngIf="element.status==='pending'"><mat-chip i18n>pending</mat-chip></mat-chip-list>
-          <mat-chip-list *ngIf="element.status==='requires grading'"><mat-chip selected color="accent" i18n>requires grading</mat-chip></mat-chip-list> 
+        <mat-cell *matCellDef="let element" [ngSwitch]="element.status">
+          <mat-chip-list *ngSwitchCase="'pending'"><mat-chip i18n>pending</mat-chip></mat-chip-list>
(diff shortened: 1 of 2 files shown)


YuserahN:
I had to make multiple commits to fix the issue it was having with indentation.

lmmrssa:
@YuserahN looks like you are making changes on github web portal. You should try to work on local copy of code. So you will be able to work on multiple files also you can test linting issue that way.

YuserahN:
@lmmrssa That is what I normally do. I use gitbash to work on a localy copy of the code and then push it to the web portal. But I was running into tslint errors and was not sure how to view them using git commands. So I just created the PR using the web portal and was able to view and fix the issues that way. For future reference, how do I view the linting errors using git?

lmmrssa:
@YuserahN you need to run `npm run install-hooks`. once that is installed every push will check for linting issues. And you will not be able to push if there is error. And it will also show you which lines has issue.

https://github.com/open-learning-exchange/planet/blob/master/Style-Guide.md this page gives you more detail on it.

YuserahN:
@lmmrssa Yes, I ran ```npm run install-hooks``` a few days ago. That's why it checked for linting errors. But I wasn't sure what the issues were, until I used the web portal. I'll check out the provided link for clarification.

YuserahN:
@lmmrssa Why does it still give me errors even when I am using 2 pace indentation? Doesn't that just mean I am supposed to press the spacebar twice?

lmmrssa:
@YuserahN there are different linting rules we have set not just two whitespace rule. So please check the linting fail. That will give detail about which line and and column there is error along with which error has occured.

src/app/submissions/submissions.component.html
```diff
@@ -36,6 +36,7 @@
           <mat-chip-list *ngIf="this.filter.type === 'survey' && element.status === 'complete'"><mat-chip selected il8n>complete</mat-chip></mat-chip-list>
           <mat-chip-list *ngIf="element.status ==='graded'"><mat-chip selected i18n>graded</mat-chip></mat-chip-list>
           <mat-chip-list *ngIf="element.status==='pending'"><mat-chip i18n>pending</mat-chip></mat-chip-list>
+          <mat-chip-list *ngIf="element.status==='requires grading'"><mat-chip selected color="accent" i18n>requires grading</mat-chip></mat-chip-list>
```
singharpita: @YuserahN can we make status to be one word while saving.

YuserahN: @singharpita You mean, 'requires grading' should be something like, 'Submitted'?

singharpita: may be require_grading or something but not separated by space

paulbert: @singharpita @YuserahN  Let's leave it with a space, please.

Issue #3190, opened 2019-02-14
Do not show unaccepted community in survey send list


lmmrssa:
![screen shot 2019-02-14 at 20 54 25](https://user-images.githubusercontent.com/2909490/52796247-8a4ee080-309b-11e9-888d-d011e231f627.png)

lmmrssa:
@BhargaviNadendla found you have assigned few issues to yourself to assigning it to arpita.

PR #3207, opened 2019-02-19
[#3190] hide unaccepeted community registration request from send sur… (connects #3190)

singharpita:
…vey list

### DIFF
--- a/src/app/shared/dialogs/dialogs-list.service.ts
+++ b/src/app/shared/dialogs/dialogs-list.service.ts
@@ -51,12 +51,7 @@ export class DialogsListService {
   getListAndColumns(db: string, selector?: any, planetField: string = 'local') {
     selector = selector || this.defaultSelectorFunctions()[db];
     if (db === '_users') {
-      return this.couchService.findAll('_users', findDocuments({
-        '$or': [
-          { 'isUserAdmin': true },
-          { 'roles': { '$in': [ 'learner', 'leader', 'manager' ] } }
-        ]
-      }, this.userService.userProperties))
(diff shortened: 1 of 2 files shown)


src/app/shared/dialogs/dialogs-list.service.ts
```diff
@@ -51,7 +51,12 @@ export class DialogsListService {
   getListAndColumns(db: string, selector?: any, planetField: string = 'local') {
     selector = selector || this.defaultSelectorFunctions()[db];
     if (db === '_users') {
-      return this.couchService.findAll('_users', findDocuments({}, this.userService.userProperties))
+      return this.couchService.findAll('_users', findDocuments({
```
paulbert: @singharpita This users for communities are found in the `child_users` database on the nation.  This change will not affect the users from communities which have not been accepted.

We need to check the `communityregistrationrequests` database and filter if the community matching the `planetCode` has not been accepted when running `getListAndColumns` for `child_users`

lmmrssa: @paulbert child_users will not be synced until we approve community right. But we do send users to users database. And list is checking planetcode from both _users and child_users to show list.

singharpita:
@paulbert there is `defaultSelectorFunctions` shall we add filter in this function? Is there any place (listview) which will require to show inactive users?

paulbert:
@singharpita You can combine this filter with what is already there for `defaultSelectorFunctions` for `child_users`, but you will need to get the data on `communityregistrationrequests` to know what to filter each time.

Issue #3192, opened 2019-02-14
Record survey incomplete submission


lmmrssa:
When user goto **Record survey** and do not answer all questions. Survey stil gets submitted.
![screen shot 2019-02-14 at 21 12 01](https://user-images.githubusercontent.com/2909490/52797287-ad7a8f80-309d-11e9-84c0-7022d6102797.png)
But when going to submision list it will show pending. And there is no way to get back to that survey to complete it as it is not associated to any user.
Nor can we see submission being pending status.
![screen shot 2019-02-14 at 21 11 27](https://user-images.githubusercontent.com/2909490/52797282-a94e7200-309d-11e9-845e-db84b5867f01.png)

paulbert:
Two things we can do to help fix this:
1. Number of Times Taken could only count complete surveys
2. Filter out pending surveys with no user

Calling this one low priority since the main functionality used in Planet is the "Send Survey"

PR #3208, opened 2019-02-19
[#3192] incomplete record survey with no user fixes (connects #3192)

singharpita:


src/app/surveys/surveys.component.ts
```diff
@@ -72,7 +72,30 @@ export class SurveysComponent implements OnInit, AfterViewInit, OnDestroy {
 
   getSubmissions() {
     // get the no of submisson for each test from submisson table
-    return this.couchService.findAll('submissions', { 'selector': { 'type': 'survey' } });
+    return this.couchService.findAll('submissions', { 'selector': {
```
paulbert: Instead of using the long find selector, lets do 
```
this.couchService.findAll('submissions').pipe((submissions) => ...
```
and then filter the submissions with a javascript filter function.  That will be more readable, and I don't think this selector will perform well on the RasPi.

singharpita:
@paulbert made requested changed please review

PR #3226, opened 2019-02-20
Achievements: remove personal Info (fixes #3222)

### LINKED ISSUE #3222: Achievements: Remove "Personal Info"
We are going to consolidate all achievement information into one list, so need to remove the "Personal Info" section

svlesiv:


yuviii:
@svlesiv Cannot see the option of adding personal information and the one added previously is also hidden. If it is the thing we need to do then working fine in my case.
@paulbert Is this the case we want or show the list in different way ? 

screenshot 2019-02-20 20 49 34

svlesiv:
Thank for the review @yuviii. I will try to fix it within 2 hours.

PR #3233, opened 2019-02-21
Achievements: Rearrange to include title & date (fixes #3232)

### LINKED ISSUE #3232: Update Achivements view for title & date

paulbert:
#3232 editing the achievements view to include the new fields.

Click on an achievement to show description and a button for each resource.  Clicking again closes the achievement.  Only on achievement should be open at a time.


lmmrssa:
Minor improvement here can be making achievement clear that its collapsable list.

Last one had attachment which is hidden. But it is not visible that we can click on achievement to see attachment inside it.
screen shot 2019-02-21 at 13 30 02
screen shot 2019-02-21 at 13 30 09

PR #3252, opened 2019-02-21
(Fixes #3238) Achievement: Label in Dialog Box

### LINKED ISSUE #3238: Achievement: Label in dialogBox
After clicking the enter the reference button, The dialog it shows has label `Add achievement` it should be `Add Reference`

Also in phone placeholder in same dialog it displays `phone#` should be sample phone number 

screenshot 2019-02-21 20 14 31

YuserahN:
Fixes #3238


YuserahN:
PR #3266 resolves this issue as well because it fixes it on a wider scale, which means this PR (#3252) is no longer needed. I would close it, but someone might want to review both just case in case.

paulbert:
@YuserahN Saw your comment after approving.  I'll close this in favor of the other PR, then.

PR #3257, opened 2019-02-22
Dialog: enable submit button for resources  (fixes #3241)

### LINKED ISSUE #3241: Submit disabled on editing 'Achievements' and updating resources
After editing an achievement, unable to update the resources

image

image

svlesiv:


svlesiv:
Thanks, @paulbert, for the advice. This is a much more elegant way to mark for as dirty without creating an extra variable.

Issue #3221, opened 2019-02-20
Back button reusable component


singharpita:
- Go to any feedback view of resource/course/teams.
- Click on resource/course/teams icon present on top right hand side.
- it will take to the particular course, resource or team detail page
- click the back button.
- It will redirect to the list page rather than feedback page.

It should be same as of browser back button.

paulbert:
I'm open now to changing the back button so it always mimics the browser back.  The main reason we changed it was to avoid a logged in user ending up on the login page, but that is no longer possible.

We should create this as a reusable component which will save some duplication across our other components.

yuviii:
@ArashDai I think it supposed to take me to the feedback list as i navigated through the feedback, but it is routing to the list for eg if it is course detail then to course list or same case for resource as well.

[Screen navigating video here on this link ](https://www.useloom.com/share/f4195306f85a41e1b61fd586adaa576e)

paulbert:
Closing.  We found another special case that I think is a little more important than this where after completing the "Add Course/Resource/etc" page, the back button would then go to the form.  In that situation it's better to go up a level.

When we have some more time to add polish, we can revisit.

PR #3271, opened 2019-02-26
3221 back button

ArashDai:
having an async issue where previousRoute remains undefined. solves #3221

### DIFF
--- a/src/app/shared/back/back.component.ts
+++ b/src/app/shared/back/back.component.ts
@@ -1,0 +1,26 @@
+import { Component } from '@angular/core';
+import { Router, RoutesRecognized } from '@angular/router';
+import {filter, pairwise} from 'rxjs/operators';
+
+@Component({
+  selector: 'app-back',
+  templateUrl: './back.component.html',
+  styleUrls: ['./back.component.scss']
+})
+export class BackComponent {
+
+  previousRoute: any;
+
+  constructor(private router: Router) {
+    this.router.events.pipe(
+      filter(e => e instanceof RoutesRecognized),
+      pairwise()
+    ).subscribe((event: any[]) => {
+      this.previousRoute = event[0].urlAfterRedirects;
+    });
+  }
+
+  goBack(){
+    this.router.navigate([this.previousRoute]);
+  }
+}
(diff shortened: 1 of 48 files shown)


yuviii:
@ArashDai I think it supposed to take me to the feedback list as i navigated through the feedback, but it is routing to the list for eg if it is course detail then to course list or same case for resource as well.

[Screen navigating video here on this link](https://www.useloom.com/share/f4195306f85a41e1b61fd586adaa576e)

src/app/shared/back/back.component.ts
```diff
@@ -0,0 +1,26 @@
...
+  selector: 'planet-back',
+  templateUrl: './back.component.html',
+  styleUrls: [ './back.component.scss' ]
+})
+export class BackComponent {
+
+  previousRoute: any;
+
+  constructor(private router: Router) {
+    this.router.events.pipe(
```
paulbert: @ArashDai Is there a reason you are subscribing to router events here?

There is a `Location` service built in to Angular with a `back()` method which should do the same.

paulbert:
@yuviii The back component is only implemented on the `courses/view` route for this PR (we can implement across other routes in another PR).

src/app/shared/back/back.component.ts
ArashDai: @paulbert I guess not, I didn't realize the Location service had a back method, I thought I had to go through the router to get to the last page.

src/app/shared/back/back.component.ts
```diff
@@ -0,0 +1,19 @@
+import { Component } from '@angular/core';
+import { Location } from '@angular/common';
+
+@Component({
+  selector: 'planet-back',
+  templateUrl: './back.component.html',
```
lmmrssa: It is single line in template so you should have inline templating.

src/app/shared/back/back.component.ts
```diff
@@ -0,0 +1,19 @@
...
+  styleUrls: [ './back.component.scss' ],
```
lmmrssa: It is empty style file so please remove this file

ArashDai:
I implemented the back button throughout the app, I left old goBack() functions alone if they were used by other pieces of code to not break things. also I left it out of users-achivements because it would take users back into the form if they made an edit and submitted it, then try to go back

paulbert:
I'm going to close this.  Unfortunately it's not a priority to figure out how best to handle the back button navigation in special situations, and we don't have a good solution right now.

Issue #3280, opened 2019-02-27
ResourcesView: Show collections


singharpita:

paulbert:
Add collections in the details section on the left:

![image](https://user-images.githubusercontent.com/9203229/53464491-70f15100-3a18-11e9-8e70-fc84094cab55.png)

PR #3284, opened 2019-02-27
[#3280]-collection-list-shown (connects #3280)

yuviii:


yuviii:
@lmmrssa the collection was shown in plain text as per suggestion of @paulbert.

lmmrssa:
Updated collection to use chip. @paulbert may be we can add links on this which will take to collection filtered resource list.

paulbert:
@lmmrssa The chips don't look very good in this view, which is why I did not and do not want them here.  I think adding links would be a larger change (in the resource table view selected collections aren't initialized based on the url or parameters), so we can add that in another PR

Issue #3281, opened 2019-02-27
Achievements: Sort by date


paulbert:
There is a request to have achievements sorted automatically by date.  I think we should implement this as follows:

1. A button which will sort the achievements by date
  - If the achievements are already sorted it will toggle ascending/descending
  - If they are not sorted the button should sort ascending
2. When the achievements are sorted already, when adding or editing an achievement we should automatically sort the achievements so they remain sorted
3. When the member first creates their achievements they should start sorted ascending by default

lmmrssa:
@paulbert Does date here means date of achievement? If that's the case then Date of achievement is optional.

paulbert:
@lmmrssa Yes good point let's have achievements with no date at the bottom sorted alphabetically by title.

PR #3290, opened 2019-02-27
[#3281] Sort achievement view (connects #3281)

singharpita:
@paulbert  added sorting of achievement on achievement view.
Is the implementation correct?
will further work on achievement update.


paulbert:
@singharpita This actually should be implemented on AchievementsUpdate.  The learner creating the document creates an order for their achievements that is static for the people viewing the achievements.

singharpita:
@paulbert got the concept will make adjustment to my pull request.
One question, when user comes back to update achievement how to identify how he has sorted it before. Do we save sorting order?

paulbert:
@singharpita Yes let's save this as the `dateSortOrder` property with possible values 'asc', 'desc', or 'none'

singharpita:
@paulbert made changes on update achievement and removed sorting from achievement view.
Added three order of sorting **asc**, **desc**, **none**. Default will be none.

PR #3293, opened 2019-02-28
Achievements pdf export

paulbert:
Note the showdown package is a dependency for the markdown viewer so it shouldn't add too much overall to the app.

Still a work in progress

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "chart.js": "^2.7.3",
     "cropperjs": "^1.3.4",
     "export-to-csv": "^0.2.1",
+    "jspdf": "^1.5.3",
     "jszip": "^3.1.5",
     "material-design-icons": "^3.0.1",
     "material-icons": "^0.3.1",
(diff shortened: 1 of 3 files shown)


singharpita:
@paulbert made PDF based on HTML element. Regarding styling found that using html2canvas could just print pdf as its being displayed. Shall i try that approach to see how it will look?

paulbert:
@singharpita I tried html2canvas and had trouble getting it working. Also, because the HTML is converted to vector there is no way to copy from the outputted PDF, which I do not think is ideal.

Sans serif is a family of fonts rather than a specific font. `Times` is a serif font.  It's likely `Arial` or `Helvetica` are available sans serif options. See http://web.mit.edu/jmorzins/www/fonts.html for a list of a few more possible sans serif fonts

paulbert:
Trying to pull this back in as a solution for health/surveys, but there are errors with the build

paulbert:
Closing since we are using a different javascript to PDF library

PR #3307, opened 2019-03-01
[Fixes #1301] Grading Status of Auto Correct Exams

### LINKED ISSUE #1301: Change final status to graded on all auto correct exam

YuserahN:
Fixes #1301


YuserahN:
@paulbert It was giving me tslint errors involving variable shadowing with ```answer```. So I changed it to ```submission.answer```.

PR #3309, opened 2019-03-01
[Fixes #1301] Auto Correct Grading Status

### LINKED ISSUE #1301: Change final status to graded on all auto correct exam

YuserahN:
Fixes #1301


YuserahN:
@paulbert I tried pushing (through Git) the commit we worked on during the meeting. However, it gave me tslint errors due to variable shadowing of ```answer```. So I changed it to ```eachAnswer```, and it is still giving me errors. I figured I'd push the commit using Github so that you can take a look at it.

PR #3310, opened 2019-03-01
[Fixes #1301] Grading Status of Auto Correct

### LINKED ISSUE #1301: Change final status to graded on all auto correct exam

YuserahN:
Fixes #1301

### DIFF
--- a/src/app/submissions/submissions.service.ts
+++ b/src/app/submissions/submissions.service.ts
@@ -91,7 +91,7 @@ export class SubmissionsService {
     const oldAnswer = submission.answers[index];
     submission.answers[index] = this.newAnswer(answer, oldAnswer, correct);
     const nextQuestion = this.nextQuestion(submission, index, 'passed');
-    if (submission.answers.findIndex(answer => answer.grade === undefined) === -1) {
+    if (submission.answers.findIndex(ans => ans.grade === undefined) === -1) {
       submission.status = 'graded';
     }
     if (correct !== undefined) {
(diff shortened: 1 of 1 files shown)


YuserahN:
@paulbert This is the code we worked on during the meeting. However, it is giving me tslint errors due to variable shadowing of ```answer``` when I try to push it through Git. So I pushed the commit through Github so that you can take a look.

lmmrssa:
@YuserahN made the changes to fix linting.
On local you should see what is the linting issue and fix that.
`npm run v-lint` to check linting issue and `npm run v-lint-fix` to fix it **(not all issues are auto fixable)**

src/app/submissions/submissions.service.ts
```diff
@@ -91,6 +91,9 @@ export class SubmissionsService {
     const oldAnswer = submission.answers[index];
     submission.answers[index] = this.newAnswer(answer, oldAnswer, correct);
     const nextQuestion = this.nextQuestion(submission, index, 'passed');
+    if (submission.answers.findIndex(ans => ans.grade === undefined) === -1) {
```
paulbert: @YuserahN To explain a little more detail: the linting rule that was broken here was for a "shadowed variable".  Specifically, there is a variable `answer` defined as a function parameter on line 89.  In the findIndex if you declare another `answer` variable within that function scope there could be confusion over which `answer` variable is being referred to.

Here's the specific documentation for the rule.  Let us know if you have any more questions!

https://palantir.github.io/tslint/rules/no-shadowed-variable/

PR #3312, opened 2019-03-03
[Fixes #2904] hover on pending submission

### LINKED ISSUE #2904: hover on pending submission
Pending submissions can not be clicked. So hovering over it should not show **hand-pointer**
screen shot 2019-01-03 at 20 44 14

coder8102:
[Fixes #2904] hover on pending submission

### DIFF
--- a/src/app/submissions/submissions.component.html
+++ b/src/app/submissions/submissions.component.html
@@ -52,7 +52,7 @@
         </mat-cell>
       </ng-container>
       <mat-header-row *matHeaderRowDef="displayedColumns" class="hide"></mat-header-row>
-      <mat-row *matRowDef="let row; columns: displayedColumns;" (click)="submissionAction(row)"></mat-row>
+      <mat-row *matRowDef="let row; columns: displayedColumns;" (click)="submissionAction(row)" [ngClass]="{'mat-row-pointer': row.status=='complete'}"></mat-row>
     </mat-table>
     <mat-paginator #paginator
       [pageSize]="50"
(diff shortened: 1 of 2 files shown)


src/app/submissions/submissions.component.html
```diff
@@ -52,7 +52,7 @@
         </mat-cell>
       </ng-container>
       <mat-header-row *matHeaderRowDef="displayedColumns" class="hide"></mat-header-row>
-      <mat-row *matRowDef="let row; columns: displayedColumns;" (click)="submissionAction(row)"></mat-row>
+      <mat-row *matRowDef="let row; columns: displayedColumns;" (click)="submissionAction(row)" [ngClass]="{'mat-row-pointer': row.status=='complete'}"></mat-row>
```
paulbert: A few notes:
* There is a global `cursor-pointer` class that would be better in this case.
* We want there to be a pointer if the submissions status does not equal pending, not only when it equals complete.  There are three states, so this will remove the pointer for the third as well.
* Please always use `===` or `!==` instead of `==` or `!=`

src/app/submissions/submissions.component.ts
```diff
@@ -16,7 +16,7 @@ import { DialogsLoadingService } from '../shared/dialogs/dialogs-loading.service
     .mat-column-name {
       max-width: 25vw;
     }
-    mat-row {
+    .mat-row-pointer {
```
paulbert: After converting to use to the global class you can remove this definition entirely.

coder8102:
Committed requested changes

PR #3316, opened 2019-03-05
New monitor role & directive for hiding content based on roles

paulbert:


paulbert:
It's easier to view the changes without the whitespace:

https://github.com/open-learning-exchange/planet/pull/3316/files?w=1

Issue #3274, opened 2019-02-27
Achievement: Add Reference email validation if email is not empty


singharpita:
since email is not mandatory field, so when email is not empty can there be validation of this field?
![screenshot from 2019-02-27 07-27-55](https://user-images.githubusercontent.com/27276132/53459304-65764800-3a61-11e9-97f5-8e1f0ecbf606.png)

paulbert:
I think that would be a good thing to add, but it's a low priority change

PR #3366, opened 2019-03-11
[#3274]email-validation-added-in-achievement-reference(connects #3274)

yuviii:


src/app/users/users-achievements/users-achievements-update.component.ts
```diff
@@ -148,7 +149,7 @@ export class UsersAchievementsUpdateComponent implements OnInit, OnDestroy {
       this.fb.group({
         relationship: '',
         phone: '',
-        email: '',
+        email: [ '', Validators.email ],
```
lmmrssa: validation does not work. It will be overwritten by references.

PR #3383, opened 2019-03-13
[Fixes #3379] Remove 'Not Graded' From Survey Status Options

### LINKED ISSUE #3379: Submissions: Survey status options
Surveys will never have the "requires grading" status, so that option should not be shown for Surveys.  Only Exams.

YuserahN:
Fixes #3379


YuserahN:
> 
> 
> @YuserahN In surveys only able to click in the `complete` survey but not in `pending` one.

@yuviii I am able to click on 'pending' in the list of filter options for survey. Are you saying that you are unable to view surveys that have a status of 'pending'? I think that's a different issue.

PR #3391, opened 2019-03-13
Resources: add many collections (fixes #3385)

### LINKED ISSUE #3385: Resources: Add collections to many
On resource list add functionality to select many resources and add to collections

paulbert:
Currently can only add collections.  Will try to add full collection management (list initializes with existing collections of selected so you can add/remove) tonight.

### DIFF
--- a/src/app/resources/resources.component.html
+++ b/src/app/resources/resources.component.html
@@ -22,12 +22,14 @@
         </button>
       </span>
       <span *ngIf="!parent">
-        <mat-form-field class="font-size-1 margin-lr-3 mat-form-field-type-no-underline mat-form-field-dynamic-width">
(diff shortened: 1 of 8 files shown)


paulbert:
@yuviii I think I see the problem.  Clicking directly on the checkbox changes the state of the checkbox, but also runs the code which changes the state stored in the component.  Each are independent of one another.  I think you end up with the UI checked, but the internal component stores that as unchecked.

lmmrssa:
@paulbert There is one issue when trying to add collections. If you select resource with collection and without collection. Then click on change selection. After that without making any changes close dialog and unselect resource with collection. Now try to change selection and add different collection to those resources.
It will add new collection as well as collection from one you had selected before. Dialog will only show new collection selected but on saving it saves both collections.

paulbert:
@lmmrssa That was even a wider issue, the selection was still holding on to anything previously selected in the background.  Fixed with my last commit.

Issue #2906, opened 2019-01-03
Notification bell menu: See All pinned to bottom


paulbert:
It would be nice to have the "See All" button pinned to the bottom instead of having to scroll through a long list of notifications to get there.

![image](https://user-images.githubusercontent.com/9203229/50644959-592dad00-0f40-11e9-8cdd-78973e39d0cf.png)

PR #3394, opened 2019-03-14
Notification See all pin to bottom (connects #2906)

singharpita:


lmmrssa:
Screen Shot 2019-03-14 at 11 47 32
Screen Shot 2019-03-14 at 11 51 40

PR #3415, opened 2019-03-15
add surveys to courses fixes #3328

### LINKED ISSUE #3328: Add a survey to a course step
Similar to the 'Add Resources' button, but only allow one survey to be added.  Note that surveys are in the `exams` database with `type: 'surveys'`

ArashDai:
todo:
prevent adding more than one survey to a course

### DIFF
--- a/src/app/courses/add-courses/courses-step.component.ts
+++ b/src/app/courses/add-courses/courses-step.component.ts
@@ -89,20 +89,11 @@ export class CoursesStepComponent implements OnDestroy {
   }
 
   addExam(type) {
-    if (type === 'exam') {
-      this.coursesService.stepIndex = this.activeStepIndex;
-      if (this.activeStep.exam) {
-        this.router.navigate([ '/courses/update/exam/', this.activeStep.exam._id ]);
-      } else {
-        this.router.navigate([ '/courses/exam/' ]);
-      }
-    } else if (type === 'survey') {
-      this.coursesService.stepIndex = this.activeStepIndex;
-      if (this.activeStep.exam) {
-        this.router.navigate([ '/courses/update/survey/', this.activeStep.exam._id, { 'type': 'surveys' } ]);
-      } else {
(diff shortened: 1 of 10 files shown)


src/app/courses/add-courses/courses-step.component.ts
```diff
@@ -88,12 +88,21 @@ export class CoursesStepComponent implements OnDestroy {
...
   }
 
-  addExam() {
-    this.coursesService.stepIndex = this.activeStepIndex;
-    if (this.activeStep.exam) {
-      this.router.navigate([ '/courses/update/exam/', this.activeStep.exam._id ]);
-    } else {
-      this.router.navigate([ '/courses/exam/' ]);
+  addExam(type) {
+    if (type === 'exam') {
```
paulbert: This should be refactored to reduce complexity.  I think you don't need to add the extra route, you can just pass the `type` function parameter as a route parameter.

ArashDai:
sorry i guess i mixed up the fact that its in the exams database with you wanting an implementation like resources, I can change it if you want

src/app/courses/add-courses/courses-step.component.ts
ArashDai: much cleaner

src/app/exams/exams-add.component.ts
```diff
@@ -114,7 +115,7 @@ export class ExamsAddComponent implements OnInit {
     this.couchService.updateDocument(this.dbName, { createdDate: date, ...examInfo, updatedDate: date }).subscribe((res) => {
       this.documentInfo = { _id: res.id, _rev: res.rev };
       let routerParams = {};
-      if (this.examType === 'courses') {
+      if (this.examType === 'courses' || this.isCourseContent) {
         this.appendExamToCourse(examInfo);
```
paulbert: We should change this to take the `type` route parameter that you're passing so the survey goes on the `survey` property of the course object.  Otherwise we'd overwrite any exams.

Issue #3399, opened 2019-03-14
ourNews: Delete or edit by creator of post


singharpita:
1. Can there be delete action for every post which can be deleted by admin and creator of that post?
2. Can there be edit action for every post which can be modified by creator only?

paulbert:
I think adding the functionality for the creator of a post is OK, but including for admins we need to consider the implications of.

paulbert:
For edit: Open a dialog to edit the post
For delete: Open a dialog to confirm the user wants to delete the post

PR #3416, opened 2019-03-15
News edit and delete (connects #3399)

singharpita:


singharpita:
@yuviii If you check issue #3399 it is mentioned to just allow **Edit** **Delete** to creator

PR #3429, opened 2019-03-18
[#3410] Prevent joining old meetup

### LINKED ISSUE #3410: Meetups : Hide the meetups from the list if its end date is behind the current date
Screenshot 2019-03-14 21 33 52

YuserahN:
Partially fixes #3410 
I've modified joinMeetup() to ensure that user cannot join an old meetup. But this only works when the user tries to join from the meetups view page. If they try to join from the meetups page, where they can select multiple meetups and click on 'join selected', they are still able to join old meetups. I will modify meetupsToggle() to ensure that the user is prevented from joining. Or I will make the changes in ```meetups.component.html``` within ```[disabled]="selectedNotJoined === 0"``` if possible. I tried making these changes through attendMeetups() so that I would not have to make these changes in two separate functions, but that didn't seem to work. Any suggestions are helpful.

### DIFF
--- a/src/app/meetups/view-meetups/meetups-view.component.ts
+++ b/src/app/meetups/view-meetups/meetups-view.component.ts
@@ -84,12 +84,16 @@ export class MeetupsViewComponent implements OnInit, OnDestroy {
   }
 
   joinMeetup() {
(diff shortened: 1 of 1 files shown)


paulbert:
Fixed in #3518

PR #3430, opened 2019-03-18
Course & resource: add edit button (fixes #2719)

### LINKED ISSUE #2719: Course & resource: edit button on View route
We should add an edit button for ResourcesView and CoursesView if the logged in user has editing rights to that resource/course.

svlesiv:
Screen Shot 2019-03-18 at 3 15 20 PM

Screen Shot 2019-03-18 at 3 15 58 PM

### DIFF
--- a/src/app/courses/view-courses/courses-view.component.ts
+++ b/src/app/courses/view-courses/courses-view.component.ts
@@ -54,8 +54,8 @@ export class CoursesViewComponent implements OnInit, OnDestroy {
         this.courseId = params.get('id');
         this.coursesService.requestCourse(
         { courseId: this.courseId, forceLatest: true, parent: this.parent }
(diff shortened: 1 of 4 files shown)


src/app/courses/view-courses/courses-view.component.ts
```diff
@@ -44,11 +47,14 @@ export class CoursesViewComponent implements OnInit, OnDestroy {
       });
       this.progress = progress;
       this.isUserEnrolled = this.checkMyCourses(course._id);
+      this.canManage = this.currentUser.isUserAdmin;
```
yuviii: @svlesiv Here it is checking the user is admin or not. `addedBy` is the username of creator of resources.

svlesiv:
Thanks for pointing this out. I will fix it.

PR #3446, opened 2019-03-20
[Fixes #3432] Add Edit Date to News Post

### LINKED ISSUE #3432: ourNews: Edit on post should change wrote on date?
can wrote one date should be changed to updated date when user update their post?
Screenshot from 2019-03-19 07-41-52

YuserahN:
Fixes #3432 
Displays edit date 

newsPostDates

### DIFF
--- a/src/app/news/news.component.html
+++ b/src/app/news/news.component.html
@@ -24,8 +24,10 @@
                 item.user.name}}
             </a>
           </mat-card-title>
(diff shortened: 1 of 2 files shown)


YuserahN:
The first commit displays both ('wrote on' date as well as edit date, if it exists). I pushed a second commit, which replaces 'wrote on' date with edit date, if edit date exists. 
I was not sure which one is preferable.

lmmrssa:
@paulbert made some changes to make sure that ourNews page could use the available space and Also not hide the story section. Please review.

PR #3447, opened 2019-03-20
Surveys: delete surveys (fixes #2751)

### LINKED ISSUE #2751: Survey: ability to select and delete surveys
Could be nice to have ability to select and delete surveys.

screen shot 2018-12-13 at 11 03 39 am

svlesiv:
Screen Shot 2019-03-20 at 2 56 05 PM

### DIFF
--- a/src/app/shared/utils.ts
+++ b/src/app/shared/utils.ts
@@ -32,7 +32,3 @@ export const styleVariables: any = (() => {
 export const filterById = (array = [], id: string) => array.filter(item => item._id !== id);
 
 export const itemsShown = (paginator: any) => Math.min(paginator.length - (paginator.pageIndex * paginator.pageSize), paginator.pageSize);
-
-export const postToDeleteItems = (dbService: any, dbName: string, deleteArray: any) => {
-  return dbService.post(dbName + '/_bulk_docs', { docs: deleteArray });
-};
--- a/src/app/shared/utils.ts
+++ b/src/app/shared/utils.ts
@@ -28,3 +28,11 @@ export const styleVariables: any = (() => {
     return { ...styleObj, [prop.trim()]: value };
   }, {});
 })();
+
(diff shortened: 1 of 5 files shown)


svlesiv:
It failed **codeclimate** because of exceeding allowed amount of function in the component (which probably is not bad) and `similar blocks of code found in other locations`. 

This functionality was implemented in many other components, and repetition is consistent through the app. Should I try to fix it?

yuviii:
> It failed **codeclimate** because of exceeding allowed amount of function in the component (which probably is not very bad) and `similar blocks of code found in other locations`.
> 
> This functionality was implemented in many other components, and repetition is consistent through the app. Should I try to fix it? Move repetitive functions into **utils** ?

Yes u are right. Need to sort out this issues. help needed @paulbert

svlesiv:
I fixed almost all issues, except the one related to repetition items in the constructor.

src/app/shared/utils.ts
```diff
@@ -28,3 +28,11 @@ export const styleVariables: any = (() => {
     return { ...styleObj, [prop.trim()]: value };
   }, {});
 })();
+
+export const filterById = (array = [], id: string) => array.filter(item => item._id !== id);
+
+export const itemsShown = (paginator: any) => Math.min(paginator.length - (paginator.pageIndex * paginator.pageSize), paginator.pageSize);
+
+export const postToDeleteItems = (dbService: any, dbName: string, deleteArray: any) => {
```
paulbert: This function should be included in CouchService directly.  Having a function which always requires the same service as the first parameter 

I think that would increase the number of functions on CouchService beyond the Code Climate limit, but I think more functions on CouchService is preferable to this.

Please call the function `bulkDocs`.

svlesiv: Thanks for the advice. I will fix that.

PR #3448, opened 2019-03-21
Additional resource filters

paulbert:
A set of filters (click "Show Filters" on resources page to display them) for resource Subject, Level, Medium, and Language.  Populates the list based on currently selected resources.

One known issue: if you select some of these filters and then select a collection it is possible to end up with 0 resources.  This is more an issue with the collection list & I more suitable for a future PR on that feature.

It is possible to end up with no resources left with a selection filter & text search as well, but I don't think that's avoidable.

### DIFF
--- a/src/app/resources/resources.component.html
+++ b/src/app/resources/resources.component.html
@@ -71,7 +71,7 @@
     </ng-container>
   </mat-toolbar>
 
-  <div class="view-container view-full-height view-table" *ngIf="!emptyData; else notFoundMessage">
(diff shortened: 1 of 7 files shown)


lmmrssa:
Testing, Collection also seems to be filtered based on Show Filter. And show filter is filtered based on collection. Right now there is one issue that Collection filter only works when you are trying to have **Multiple selection**, on simple selection (single collection) it does not filter the collection.

But it looks good right now. Also right now filter has fixed height and it scroll when there is lot of items. Was wondering if we can have filter section to be something like notification or profile menu. So that the resource list section does not look hidden.

Approving this PR as it looks good.

Issue #3444, opened 2019-03-20
News: modify layout


svlesiv:
It might be inconvenient to scroll over posts using a small window. Maybe it would be better not to confine posts into a container with a defined size.

![Screen Shot 2019-03-20 at 11 20 43 AM](https://user-images.githubusercontent.com/35533801/54709188-4ddf2c00-4b02-11e9-8621-c693860d2181.png)

It might look something like this:
![Screen Shot 2019-03-20 at 11 31 41 AM](https://user-images.githubusercontent.com/35533801/54710084-2b4e1280-4b04-11e9-97a3-f68eca81299d.png)

paulbert:
I think we could have a link that stays at the top to scroll back to post.  If this list gets very long it might be challenging for the learner to make it back to the top.

PR #3454, opened 2019-03-21
[#3444] split layout of ournews

lmmrssa:


src/app/news/news.scss
```diff
@@ -1,14 +1,12 @@
-$max-width: 640px;
+@import '../variables';
 
 .view-container {
   display: grid;
-  grid-template-rows: auto 1fr;
   grid-template-areas: 'form' 'news';
```
paulbert: For `grid-template-areas` rows are separated by quotes and columns by spaces within quotes.  Here you've defined two rows as the template, but you should change it to `'form news'`

src/app/news/news.scss
```diff
@@ -1,14 +1,12 @@
...
-  justify-content: center;
-  grid-template-columns: minmax(auto, $max-width);
+  grid-template-columns: 1fr 2fr;
 
   .new-post-form {
-    grid-area: form;
+    grid-area: 'form';
```
paulbert: This is not valid CSS.  `grid-area` should not have quotes.

image

src/app/news/news.scss
```diff
@@ -1,14 +1,12 @@
-$max-width: 640px;
+@import '../variables';
```
paulbert: Doesn't look like any of the variables are used here, so this can be removed

paulbert:
I'm going to close this since I think we're going in a different direction with the Community news

Issue #3450, opened 2019-03-21
Resources: Adding My Resource filter in resource list


yuviii:
I think it would be good to have the My Resource filter button in the toolbar, so it would allow the user to see the resource list added only by them.

paulbert:
Also on the dashboard when the learner clicks "myLibrary" they should see this view with only the resources they have selected

paulbert:
Lets have a `mat-stroked-button` next to the Library heading that switches between the two views

PR #3463, opened 2019-03-22
[#3450] myLibrary Filter

singharpita:


src/app/dashboard/dashboard.component.html
```diff
@@ -30,7 +30,7 @@ <h2>{{displayName}}</h2>
   [cardTitle]="'myLibrary'"
   class="planet-library-theme"
   [itemData]="data.resources"
-  [emptyLink]="'/resources'"
```
paulbert: We need to have two link `@Input` variables here.

`link` should be for clicking **myLibrary**, **myCourses**, etc.
`emptyLink` should still be for clicking **Add item to myLibrary**, etc when myLibrary is empty.

PR #3466, opened 2019-03-24
acheivement sort breaking fixed (connects #3408)

### LINKED ISSUE #3408: Achievements: new entries break sorting
In the below video "Test 2" and "Test 3" were already on the achievements doc while "Test" & "Test 4" were just created.  The date is in a different format than what is saved to the database.

This change: https://github.com/open-learning-exchange/planet/commit/4e6f44d7aca5227614b8e376fc4ee69418d127f3 removed the `.toJSON()` method which would fix this but was causing another error.  We need to modify this line to fix sorting AND avoid #3342 

https://www.useloom.com/share/a5faa96cfba44a019608cfbb13e54085

yuviii:
fixes #3408

### DIFF
--- a/src/app/users/users-achievements/users-achievements-update.component.ts
+++ b/src/app/users/users-achievements/users-achievements-update.component.ts
@@ -131,7 +131,8 @@ export class UsersAchievementsUpdateComponent implements OnInit, OnDestroy {
         date: [ achievement.date, null, ac => this.validatorService.notDateInFuture$(ac) ]
       }),
       { onSubmit: (formValue, formGroup) => {
-        const achievedAt = (formGroup.controls.date.value instanceof Date) ? formGroup.controls.date.value.toISOString() : formGroup.controls.date.value;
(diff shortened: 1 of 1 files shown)


svlesiv:
Works for me, but I didn't have achievements with the old date format to catch the error.

yuviii:
@paulbert Please review the update.

paulbert:
@yuviii You did not fix the error.  Create an achievement with a date & save all achievements.  Then go back in to update that achievement but do not change the date (change a different field).  You will see the error from my previous post.

The problem is not that the function is called when `formGroup.controls.date.value` does not exist, it is that `formGroup.controls.date.value` is sometimes a string and sometimes a Date object with the `toISOString()` method.  You need to check if `formGroup.controls.date.value.toISOString` exists.  This is what I tried to say in the meeting, but I guess I was not clear enough.

lmmrssa:
@paulbert can you check if changes made address the issue?

PR #3469, opened 2019-03-25
[Fixes #3460] Add Complete Button to Exam/Survey

### LINKED ISSUE #3460: ExamsView: Add "Complete Survey/Exam" button
For exams and surveys we should have an additional button so there's a little less confusion if questions are answered out of order.

The view should use the following logic:
If there are other unanswered questions (not counting the one currently being submitted), there should be only one "Submit Answer" button which uses the current functionality (goes to the next unanswered question)

If every other question has a submitted answer, the "Submit Answer" button should progress to the next question and there should be an additional "Complete Survey" or "Complete Exam" button so the learner can exit the exam if they choose.  If the current question is also the last question in the exam, the "Submit Answer" button should be hidden.

YuserahN:
Fixes #3460 (or partially fixes #3460) 
The 'Submit Answer' button is replaced by 'Complete' when the user is on the last question.

### DIFF
--- a/src/app/exams/exams-view.component.html
+++ b/src/app/exams/exams-view.component.html
@@ -61,10 +61,9 @@
         color="primary"
         (click)="nextQuestion()"
         [planetSubmit]="spinnerOn"
-        [disabled]="!answer.valid || grade === undefined || grade === null"
-        i18n>
-        <ng-container *ngIf="(maxQuestions-1) === answerCount">Complete</ng-container>
-        <ng-container *ngIf="(maxQuestions-1) !== answerCount">Submit Answer</ng-container>
+        [disabled]="!answer.valid || grade === undefined || grade === null">
+        <ng-container *ngIf="(maxQuestions-1) === answerCount" i18n>Complete</ng-container>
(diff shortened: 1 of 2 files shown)


paulbert:
@YuserahN That error is found in surveys or if you browser refresh on a course exam.

The problem is that `SubmissionsService.submission` won't be set until after an async request as part of the `SubmissionsService.openSubmission` function.  You should initialize `answerCount` with 0 and lines 64 & 70 should just occur once in the `subscribe` portion of `setSubmissionListener`.  I can go over it in more detail at the meeting tomorrow.

YuserahN:
I've removed the WIP label because technically issue #3460 is resolved through this pull request. While it doesn't work entirely the way we'd want it to, this is due to another issue. The survey/exam should go to the next unanswered question upon clicking the 'Submit Answer' button. So if I skip question #2 and answer question #3, clicking 'Submit Answer' should take me back to question #2. Instead, it proceeds to question #4 and lets me "complete" the exam/survey without answering all of the questions. I will go ahead and open a new issue for this portion.

PR #3471, opened 2019-03-25
[Fixes #3401] ourNews: user with no role with access to post

### LINKED ISSUE #3401: ourNews: user with no role with access to post
should user with no role can post in ourNews?

coder8102:
[Fixes #3401] ourNews: user with no role with access to post

**Learner**
issue#3401_learner

**No role**
issue#3401_no-role

### DIFF
--- a/src/app/news/news.component.html
+++ b/src/app/news/news.component.html
@@ -7,7 +7,7 @@
     <mat-toolbar-row class="primary-color font-size-1"></mat-toolbar-row>
   </mat-toolbar>
   <div class="view-container view-full-height">
-    <div class="new-post-form">
+    <div *planetAuthorizedRoles="'learner'" class="new-post-form">
       <mat-form-field class="full-width mat-form-field-type-no-underline">
         <planet-markdown-textbox class="full-width" [(ngModel)]="newMessage" i18n-placeholder placeholder="What would you like to share?"></planet-markdown-textbox>
       </mat-form-field>


coder8102:
@paulbert How to allow leader and manager having access to POST? Should I just add roles into condition as *planetAuthorizedRoles="'learner',''leader','manager'"

src/app/news/news.component.html
```diff
@@ -7,7 +7,7 @@
     <mat-toolbar-row class="primary-color font-size-1"></mat-toolbar-row>
   </mat-toolbar>
   <div class="view-container view-full-height">
-    <div class="new-post-form">
+    <div *planetAuthorizedRoles="'learner'" class="new-post-form">
```
yuviii: @coder8102 `<div *ngIf="currentUser.isUserAdmin || currentUser.roles.length" class="news-post-form">` will work for this case.

paulbert:
@yuviii That is the way we would do this before, but we should use the custom `*planetAuthorizedRoles` directive instead of `*ngIf`.

Also I'll add an issue related to my message in the gitter channel:
```
As far as leaders & managers not having access to post, we need to make a separate change to how roles are set to fix this. Every activated member should be a learner at a minimum, so setting *planetAuthorizedRoles="'learner'" on an element will show it to every activated member (members can be deactivated by clicking the lock symbol on the Members page)
```

paulbert:
See #3488 as follow up to make this work for leaders as well.  Leader should also be learner always (so we should not allow members with only leader role)

PR #3472, opened 2019-03-26
[Fixes #3324] Notification link should be disable for user with no role

### LINKED ISSUE #3324: Notification link should be disable for user with no role
user with no role has no access to any view (Course, Resource, Team, Meetup, Survey).
there might be case when User (with learner role) has requested to join meetups and then that User role has been updated to no Role.
In such case notification link is still visible and clickable that redirects to the corresponding detail page.
screenshot from 2019-03-06 07-40-12

It might be better if notification link is disabled for user with no role.

coder8102:
[Fixes #3324] Notification link should be disable for user with no role

### DIFF
--- a/src/app/meetups/view-meetups/meetups-view.component.html
+++ b/src/app/meetups/view-meetups/meetups-view.component.html
@@ -20,7 +20,7 @@
   </mat-toolbar>
   <div class="view-container">
     <div class="left-tile">
-      <p *ngIf="meetupDetail.createdBy" i18n><b>Created By:</b> <a [routerLink]="['/users/profile', meetupDetail.createdBy]">{{' ' + meetupDetail.createdBy}}</a></p>
+      <p i18n><b>Created By:</b> <a [routerLink]="['/users/profile', meetupDetail.createdBy]">{{' ' + meetupDetail.createdBy}}</a></p>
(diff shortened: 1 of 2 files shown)


src/app/meetups/view-meetups/meetups-view.component.html
```diff
@@ -20,7 +20,7 @@ <h3 class="margin-lr-3 ellipsis-title">{{meetupDetail.title}}</h3>
   </mat-toolbar>
   <div class="view-container">
     <div class="left-tile">
-      <p i18n><b>Created By:</b> <a [routerLink]="['/users/profile', meetupDetail.createdBy]">{{' ' + meetupDetail.createdBy}}</a></p>
+      <p *ngIf="meetupDetail.createdBy" i18n><b>Created By:</b> <a [routerLink]="['/users/profile', meetupDetail.createdBy]">{{' ' + meetupDetail.createdBy}}</a></p>
```
paulbert: Meetups will always have a `createdBy` when created, so this is unnecessary

coder8102: Thanks Paul. I set condition to show notifications only for user with atleast learner role.

No notification icon for user with no role:
issue-3324

Issue #3475, opened 2019-03-26
Reset selection on myLibrary toggle


lmmrssa:
When myLibrary and Community resources is toggled. Selection done on one list is still set. We need to reset selection when toggle change.
![Screen Shot 2019-03-26 at 07 29 53](https://user-images.githubusercontent.com/2909490/54965554-24f2e700-4f99-11e9-856b-50981e57b7f6.png)
![Screen Shot 2019-03-26 at 07 30 02](https://user-images.githubusercontent.com/2909490/54965555-24f2e700-4f99-11e9-8c4c-00940e8b0d57.png)

PR #3481, opened 2019-03-26
[#3475] Reset selection on mylibrary toggle

lmmrssa:


paulbert:
@lmmrssa I'm leaving this unmerged to make sure you have an opportunity to test before creating the new version

paulbert:
@lmmrssa I rebased again and incorporated #3467 because I think it would be good to try to maintain the parts of the selection we can.

I noticed if the page # is changing because of the filter it still removes the selection, but I think we can try to fix that in another issue.

If the two new commits don't work feel free to remove them and merge this without them.

lmmrssa:
Looks good. Yes when number of page changes it is loosing the selection but as that is something we want to handle on another issue. I would merge this PR.

PR #3487, opened 2019-03-28
GET /update/planetapk

i5o:


lmmrssa:
Right now there is no interface or automatic checking but running `http://localhost/upgrade/fromnation/planetapk` is downloading myplanet from nation

PR #3500, opened 2019-03-29
Update upgrade script

lmmrssa:


lmmrssa:
Changed parentDomain fetch from config.
Right now we are planning to change all parentDomain to use **domain/db** pattern.

lmmrssa:
* https://github.com/open-learning-exchange/planet/blob/9c570ee1e4d99aa384d6e6b807549661689f7a1f/docker/planet/nginx/upgrade_planetapk.sh#L17
This line seems to be downloading myplanet.apk without checking if it is new one or not. It is only checking after it download. So it will consume bandwidth each time we start it. So we should check the sha before we download.

* I tried executing `*.sh` files directly and it seems to run good but nginx url seems to be giving timeout error. I assume its because `*.sh` file takes too long to execute because of the file it needs to download. Also curl of download does not seem to run on background so `*.sh` file needs to wait before returning output for nginx.

* curl seem to have output of progress that is what will be nice to see if we are waiting for download of file. And try to return continuous response of the `*.sh`

i5o:
Looks good

docker/planet/nginx/upgrade_planetapk.sh
```diff
@@ -14,12 +15,14 @@ tag=$(echo "$releaseUrl" | sed -r 's/.*\/(.*)/\1/')
 
 downloadUrl="$repoUrl/releases/download/$tag/$repoApkName"
 
-curl -s "$downloadUrl" -o "$localApkName.tmp" -L
+#curl -s "$downloadUrl" -o "$localApkName.tmp" -L
```
paulbert: Could you remove the commented out lines before merging?

docker/planet/nginx/upgrade_planetapk.sh
```diff
@@ -14,12 +15,14 @@ tag=$(echo "$releaseUrl" | sed -r 's/.*\/(.*)/\1/')
...
 
-if [ -f "$localApkName" ] && echo "$(sha256sum $localApkName | head -c64)  $localApkName.tmp" | sha256sum -s -c -; then
+if [ -f "$localApkName" ] && echo "$(curl -s "$repoUrl/releases/download/$tag/$sha256File" | head -c64)  $localApkName" | sha256sum -s -c -; then
```
paulbert: There's an extra space in this line (after `head -c64)`)

PR #3505, opened 2019-03-30
Fixes# issue-2326

### LINKED ISSUE #2326: Change password as same old password shouldn't work
While changing the password, when given the same old password as new password is working now.

coder8102:
[Fixes #2326 ] Change password as same old password shouldn't work
(old password and password are same)
issue-2326

(old password and password are different)
issue-2326-1


src/app/validators/custom-validators.ts
```diff
@@ -195,6 +195,35 @@ export class CustomValidators {
     };
   }
 
+  // Set this on both password and confirmation fields so it runs when either changes
+  // confirm should be false for the confirmation field validator
+  static unmatchPassword(unmatch: string, confirm: boolean): ValidatorFn {
```
paulbert: This function is a near copy of `matchPassword`.  Please try to refactor to use `matchPassword` for both so we don't have duplication here.

src/app/validators/custom-validators.ts
```diff
@@ -168,25 +168,33 @@ export class CustomValidators {
...
 
-      const matchControl = ac.parent.get(match),
+      const matchControl = ac.parent.get(matchField),
         val1 = ac.value,
         val2 = matchControl.value,
         confirmControl: AbstractControl = confirm ? ac : matchControl;
 
       // If passwords do not match, set error for confirmation field
-      if (val1 !== val2) {
+      if (match && val1 !== val2) {
```
paulbert: Rather than adding another conditional, you only need to check if `match === (val1 !== val2)`.  You would then set an `errorType` variable with `matchPassword` or `unmatchPassword` and use that like so: `setErrors({ [errorType]: false })`.

src/app/validators/custom-validators.ts
```diff
@@ -168,27 +168,30 @@ export class CustomValidators {
...
+        errorType = match ? 'matchPassword' : 'unmatchPassword';
+
+      // If passwords do not match when match=true, set error for confirmation field
+      // If passwords match when match=false, set error for confirmation field
+      if (match === (val1 !== val2)) {
+        confirmControl.setErrors({ [errorType]: false });
+        // If this is set on the confirmation field, also return error
         if (confirm) {
-          return { matchPassword: false };
+          return { [errorType]: false };
```
paulbert: @coder8102 See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Computed_property_names for this syntax (Computed property names).  It's useful here because then we avoid repeating the `match ? ...` conditional.

PR #3506, opened 2019-03-30
Achievements: filter missing resources (fixes #3216)

### LINKED ISSUE #3216: Achievements : Need to handle if the resources is deleted.
I think it is good to show message like `resource not found or deleted`, if the resource added in achievement is deleted.

Steps I did
 1. attach the resources in the achievement 
 2. opened the resources list
 3. deleted the resource file from the attached resource in the achievement.
 3. opened the achievement again

screenshot 2019-02-20 07 35 38

svlesiv:
Screen Shot 2019-03-30 at 1 21 54 PM

Screen Shot 2019-03-30 at 2 59 20 PM

### DIFF
--- a/src/app/users/users-achievements/users-achievements.service.ts
+++ b/src/app/users/users-achievements/users-achievements.service.ts
@@ -18,17 +18,15 @@ export class UsersAchievementsService {
     private resourcesService: ResourcesService,
     private route: ActivatedRoute
   ) {
-    this.resourcesService.resourcesListener(this.parent).pipe(map((resources: any) => {
-      resources.map((res: any) => { this.resourceIds.push(res._id); });
-      return this.resourceIds;
-    })).subscribe((resIds: any) => {
-      this.resourceIds = resIds;
+    this.resourcesService.resourcesListener(this.parent)
+      .subscribe((resources: any) => {
+        this.resourceIds = resources.map((res: any) => res._id);
     });
-    this.resourcesService.requestResourcesUpdate(this.parent);
   }
 
 
   getAchievements(id) {
+    this.resourcesService.requestResourcesUpdate(this.parent);
     return this.couchService.get(this.dbName + '/' + id).pipe(
       map((userAchievements: any) => {
         userAchievements.achievements.forEach((achievement: any) => {
--- a/src/app/users/users-achievements/users-achievements.service.ts
+++ b/src/app/users/users-achievements/users-achievements.service.ts
@@ -1,6 +1,9 @@
 import { Injectable } from '@angular/core';
 import { CouchService } from '../../shared/couchdb.service';
+import { forkJoin } from 'rxjs';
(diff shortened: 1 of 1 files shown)


src/app/users/users-achievements/users-achievements.service.ts
```diff
@@ -1,19 +1,44 @@
...
   constructor(
-    private couchService: CouchService
-  ) {}
+    private couchService: CouchService,
+    private resourcesService: ResourcesService,
+    private route: ActivatedRoute
+  ) {
+    this.resourcesService.resourcesListener(this.parent).pipe(map((resources: any) => {
+      resources.map((res: any) => { this.resourceIds.push(res._id); });
+      return this.resourceIds;
```
paulbert: You don't need the `map` function here.  You can simply run all of the code in the `subscribe` function

src/app/users/users-achievements/users-achievements.service.ts
```diff
@@ -1,19 +1,44 @@
...
 
   constructor(
-    private couchService: CouchService
-  ) {}
+    private couchService: CouchService,
+    private resourcesService: ResourcesService,
+    private route: ActivatedRoute
+  ) {
+    this.resourcesService.resourcesListener(this.parent).pipe(map((resources: any) => {
+      resources.map((res: any) => { this.resourceIds.push(res._id); });
```
paulbert: The only array function which should have side-effects (modifying objects outside of its scope) is `forEach`, and we should avoid using that when possible.

Instead of writing this with side effects you can write it like this (return is implied for one line arrow functions):
```
this.resourceIds = resources.map((res: any) => res._id);
```

src/app/users/users-achievements/users-achievements.service.ts
```diff
@@ -1,19 +1,44 @@
...
+      return this.resourceIds;
+    })).subscribe((resIds: any) => {
+      this.resourceIds = resIds;
+    });
+    this.resourcesService.requestResourcesUpdate(this.parent);
```
paulbert: There's a problem here: this will run once when the application starts.  If you delete a resource and immediately check your achievements, it will still be there.  If you open a different page that calls this, it will run and remove the resource.

I think you need to incorporate this into `getAchievements` so the request is made every time and the resources used is up to date.

svlesiv:
@paulbert, thank you for your advice. I will try to fix it.

paulbert:
@svlesiv That still had an issue where the first time you open the achievements the resource was still there.  I did a little digging into what can be done with rxjs and came up with a solution which makes sure `getAchievements` uses the latest resources.

There was also a small problem I noticed where you were getting `parent` from the `routeParams`.  Services run independently of the components and the routers, so this would initialize to whatever the first route loaded was and never change.  Since we don't have any functionality to view achievements from the parent planet, I just set it to false.

PR #3518, opened 2019-04-02
[#3410] Prevents Joining Old Meetups From View Page

YuserahN:


src/app/meetups/meetups.component.ts
```diff
@@ -56,9 +57,10 @@ export class MeetupsComponent implements OnInit, AfterViewInit, OnDestroy {
     private userService: UserService,
     private meetupService: MeetupService,
     private stateService: StateService,
-    private dialogsLoadingService: DialogsLoadingService
+    private dialogsLoadingService: DialogsLoadingService,
```
paulbert: Extra comma, should be removed.

src/app/meetups/meetups.component.ts
```diff
@@ -203,7 +205,17 @@ export class MeetupsComponent implements OnInit, AfterViewInit, OnDestroy {
...
-    this.selectedJoined = inShelf;
     this.selectedNotJoined = notInShelf;
+    this.selectedJoined = inShelf;
+    let index, meetupDate;
+    for (let i = 0; i < selected.length; i++) {
+      index = this.meetups.data.findIndex(x => x._id === selected[i]);
+      meetupDate = this.meetups.data[index].endDate ? this.meetups.data[index].endDate : this.meetups.data[index].startDate;
+      if (this.dateNow > meetupDate) {
+        this.selectedNotJoined--;
+        selected.splice(i, 1);
```
paulbert: I think this is causing strange behavior.  I can't leave an old meetup (it says "You have left 0 meetups") and after a refresh I'm still a member of that meetup.

Also it's causing the select all button to flicker when select all is clicked

src/app/meetups/meetups.component.ts
```diff
@@ -203,7 +205,17 @@ export class MeetupsComponent implements OnInit, AfterViewInit, OnDestroy {
...
+      } 
+    }
+    this.userService.changeShelf(selected, 'meetupIds', 'remove');
```
paulbert: This doesn't actually do anything, and I think you should remove it entirely anyway.  Let's let the user remove themselves from meetups that are older rather than silently removing them in the background.

YuserahN:
@paulbert So, in ```selected.filter(id =>...)```, since the array just contains the selected ids, I need to access the dates through ```this.meetups.data...``` by getting the id from ```selected``` and matching it with the id in ```meetups```, correct? That is what I tried to do in my latest commit.

PR #3527, opened 2019-04-03
New dashboard layout

paulbert:
Starts the transformation from the old dashboard layout to the new one.

Changes can be viewed in beta mode.


lmmrssa:
looks good.
One issue is that when beta is set to be user preferences and user has not triggered preferences, home page is empty. It only shows after user select some preferences.

PR #3529, opened 2019-04-04
Set /upgrade timeout to 15 minutes (fixes #3528)

### LINKED ISSUE #3528: upgrade: 504 gateway time-out

i5o:


paulbert:
@i5o the upgrade script uses `/upgrade`, not `/latest`.  I don't think we use `/latest` anywhere, in fact.

i5o:
whops whops - @paulbert 

this basically happened because we tested using a different method created to fail :)
then I copied the line to the wrong place -- fixed

good catch

paulbert:
@i5o No worries, thanks for the quick fix!

PR #3549, opened 2019-04-07
[Fixes #3488] Members: Every active member is learner

### LINKED ISSUE #3488: Members: Every active member is learner
Two changes need to be made to bring this into full effect:

1. We should not include the "learner" role in the set roles list of the Member management section.  This role should be always set when the member is "unlocked", and should be removed with every other role when the member is "locked".
2. The one exception is the "managers" (set with **Promote**) which were already implemented to remove the `learner` role.  Since they are additional administrators we can give them broad access similar to the `_admin` role in `AuthorizedRolesDirective`

coder8102:
[Fixes #3488] Members: Every active member is learner

**Role: Leader**
issue-3488 2

**Role: no role**
issue-3488 4

### DIFF
--- a/src/app/shared/authorized-roles.directive.ts
+++ b/src/app/shared/authorized-roles.directive.ts
@@ -15,7 +15,7 @@ export class AuthorizedRolesDirective {
   @Input()
   set planetAuthorizedRoles(rolesString: string) {
     const authorizedRoles = (rolesString || '').split(',').map(val => val.trim());
-    if (this.userService.doesUserHaveRole([ '_admin', ...authorizedRoles ])) {
+    if (this.userService.doesUserHaveRole([ '_admin', 'manager', ...authorizedRoles ])) {
       this.viewContainer.createEmbeddedView(this.templateRef);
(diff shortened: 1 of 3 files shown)


src/app/shared/authorized-roles.directive.ts
```diff
@@ -15,7 +15,7 @@ export class AuthorizedRolesDirective {
   @Input()
   set planetAuthorizedRoles(rolesString: string) {
     const authorizedRoles = (rolesString || '').split(',').map(val => val.trim());
-    if (this.userService.doesUserHaveRole([ '_admin', ...authorizedRoles ])) {
```
paulbert: Please only add `'manager'` to the array.

paulbert:
I also found an issue where a member loses the "learner" role.  See video (there's some unwanted scrolling happening after clicks, but by the end member "q" only has the "leader" and "monitor" roles):

https://www.loom.com/share/0ec50a168d0b4e80bb0f20b1cb28c1fe

PR #3555, opened 2019-04-08
Survey/Exam: fix long title (fixes #3512)

### LINKED ISSUE #3512: Survey: Handling long titles
1. can there be scroll when title of survey is long similar to course description.
Screenshot from 2019-04-02 07-23-40
Screenshot from 2019-04-02 07-23-14
2. No of times taken can be made centered align rather than left aligned.

svlesiv:
I've fixed long title in Exam and Survey view:
Screen Shot 2019-04-08 at 1 09 16 PM

However, I'm not sure if I should fix long title in MySurveys and Surveys views (list of surveys). Maybe better to keep full titles here.

Screen Shot 2019-04-08 at 12 46 23 PM

vs

Screen Shot 2019-04-08 at 12 47 13 PM

### DIFF
--- a/src/app/exams/exams-view.component.html
+++ b/src/app/exams/exams-view.component.html
@@ -6,8 +6,9 @@
 
 <div class="space-container">
   <mat-toolbar class="primary-color font-size-1">
+    <span class="ellipsis-title">{{title ? title + ': ' : ''}}</span>
     <span>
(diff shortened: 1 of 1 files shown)


svlesiv:
Hi @yuviii, I didn't fix long titles in a list of surveys or exams. I just showed how it may look like if it will be truncated. I tried to do it and thought that maybe better to keep a full title there.

I've fixed only in Exam/Survey view. When you click on a survey with a long title, the title supposed to be truncated on a blue background.

PR #3581, opened 2019-04-10
Learner progress updates (fixes #3574)

### LINKED ISSUE #3574: Learner Progress updates
- Total errors should be removed for this view
- Should not fill in blue until step is completed
- Less padding for the numbers
- Implement better scrolling for lots of steps

paulbert:
Still working on the scrolling portion of the issue.  May need to merge without that if it is too difficult.

### DIFF
--- a/src/app/courses/courses.service.ts
+++ b/src/app/courses/courses.service.ts
@@ -112,7 +112,7 @@ export class CoursesService {
     }, { resources: [], exams: [] });
   }
 
-  getCourses({ ids = [], addProgress = false, addRatings = false }, opts?) {
+  getCourses({ ids = [], addProgress = false, addRatings = false }, opts = {}) {
     this.currentParams = { ids, opts };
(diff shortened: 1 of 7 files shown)


paulbert:
This is ready to go.  There is a minor issue with Firefox where the scrollbars are still showing (it's very difficult to remove them), but we can deal with that in another issue.

Changes:
- Learner view:
  - Total errors is no longer there
  - Cells/boxes will only fill with blue if the learner has at least started that step
- Both learner and leader view:
  - Cell size is reduced
  - Left most columns and bottom row are fixed to view.  Scroll bars are removed in Chrome, but you can scroll by click & drag with the mouse or vertically with the mouse wheel

There's a course on dev you can replicate with 20 steps (DEMO - Course with 20 steps) that will help in creating a view to test the scrolling.  Here's a demo I made of the scrolling (showing the leader view because it's easier to create data for that):

https://www.loom.com/share/3501cb32f60f40aebae859c6a0ea375b

PR #3582, opened 2019-04-10
Check sha256sum to download

lmmrssa:


paulbert:
From github script is working well.  Checks sha successfully and downloads apk if there is not a match.

From nation always shows the below and downloads a new apk.

```
Downloading new version

Verifying apk
Download error
Downloading new version
```

lmmrssa:
@paulbert could you please verify this

PR #3596, opened 2019-04-11
Make pre-push hook executable

paulbert:
Changes the permissions on the pre-push hook to avoid git ignoring it.


paulbert:
Oops invalid name for the docker tag.  Will reopen with better branch name

PR #3603, opened 2019-04-12
CourseDetails: add an icon to attempted step (fixes #3587)

### LINKED ISSUE #3587: CourseDetails : indicators for already attempted steps in Stepslist
I think it would be a good idea to color the row or use some kind of icon to distinguish the attempted and unattempted steps in step list as now progress bar shows. Particularly when we have many steps number this can be useful for the user to visualize more clearly which steps are accomplished so far.

Screenshot 2019-04-11 20 27 38

svlesiv:
Screen Shot 2019-04-12 at 1 43 15 PM

### DIFF
--- a/src/app/courses/progress-courses/courses-progress-bar.component.ts
+++ b/src/app/courses/progress-courses/courses-progress-bar.component.ts
@@ -20,6 +20,7 @@ export class CoursesProgressBarComponent implements OnChanges {
     this.steps = this.course.steps.map((step: any, index: number) => {
       const progress = this.courseProgress.find((p: any) => p.stepNum === (index + 1));
       const status = this.progressStatus(progress);
+      step.status = status;
       return { stepTitle: step.stepTitle, status };
     });
   }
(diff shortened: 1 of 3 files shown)


paulbert:
@svlesiv I modified the solution to avoid including a side effect in the reusable component.  Changing an `@Input` variable to modify the parent component isn't the best pattern, so I just added the progress variable to the steps on CoursesView directly

paulbert:
@svlesiv I also made a change to only show the checkmark of the step has been completed. Thinking about this a little more, I think we need a different icon for a step that is in progress (when the `passed` property of the progress doc is false) rather than just leaving those steps blank.

svlesiv:
Thank you @paulbert for suggestions. I tried different Material icons and stoped on a few options:

Screen Shot 2019-04-16 at 4 11 12 PM

Screen Shot 2019-04-16 at 4 38 39 PM

Screen Shot 2019-04-16 at 5 06 27 PM

I like the last one the most. Unfortunately, Material icons library is quite limited.

Issue #3536, opened 2019-04-04
Teams: Button spacing


yuviii:
![Screenshot 2019-04-04 20 57 57](https://user-images.githubusercontent.com/6964191/55567086-b80be980-571c-11e9-963c-d4f5d314a7d4.png)

lmmrssa:
There is `Manage Team` or `View Team` button which was supposed to show there but it looks like `userStatus` is returning a condition which cause this button to be empty. That's why there is button but its showing empty causing the empty space between **Request pending** and **Feedback**

PR #3634, opened 2019-04-17
Team-button-space fixed(connects #3536)

yuviii:


src/app/teams/teams.component.html
```diff
@@ -53,7 +53,7 @@ <h3>
               Request pending
             </button>
           </ng-container>
-          <button mat-raised-button color="primary" [routerLink]="['view', element.doc._id]" i18n>
+          <button mat-raised-button color="primary" [routerLink]="['view', element.doc._id]" *ngIf="element.userStatus!=='requesting'" i18n>
```
lmmrssa: **Manage Team** is seen user is already member. Other than that it should be **View Team** This button needs to be always available. Just the label changes.

src/app/teams/teams.component.html
```diff
@@ -53,7 +53,7 @@ <h3>
...
             { element.userStatus, select, member {Manage Team} unrelated {View Team} }
```
paulbert: In this line we should use the `other` keyword to set **View Team** as the default if the user is not a member of the team, instead of only setting the button if `userStatus` is `unrelated`.

They don't write that in the clearest way, but you can see some examples in the [i18n documentation](https://angular.io/guide/i18n#regular-expressions-for-plurals-and-selections)

Issue #3636, opened 2019-04-17
Update upgrade function on dashboard.


lmmrssa:
* Add upgrade myplanet function on manager dashboard
* For center user `/upgrade/planetapk`
* For other user `/upgrade/fromnation/planetapk`
* Add feature to verify upgrade

PR #3641, opened 2019-04-18
Upgrade function optimize (connects #3636)

lmmrssa:


src/app/manager-dashboard/manager-dashboard.component.html
```diff
@@ -20,7 +20,9 @@
     <ng-container *ngIf="requestStatus === 'accepted'">
       <a routerLink="sync" i18n mat-raised-button>Manage Sync</a>
       <a routerLink="/upgrade" *ngIf="versionLocal !== versionParent" i18n mat-raised-button>Upgrade</a>
+      <a routerLink="/upgrade/myplanet" *ngIf="versionLocal === versionParent" i18n mat-raised-button>Upgrade myPlanet</a>
     </ng-container>
+    <a routerLink="/upgrade/myplanet" *ngIf="" i18n mat-raised-button>Upgrade myPlanet</a>
```
paulbert: This is never shown because of the `*ngIf`, and would be a duplicate of the other button on line 23 anyway.  I think this line can be removed

src/app/upgrade/upgrade.component.html
```diff
@@ -28,6 +28,7 @@
       </mat-card-content>
       <mat-card-actions>
         <button mat-raised-button color="primary" routerLink="/">OK</button>
+        <button mat-raised-button color="primary" (click)="upgrade()" *ngIf="enabled || done">Verfiy</button>
```
paulbert: This button just restarts the process.  I think it is confusing, but are you including this to solve the issue of false successes with the upgrade?

Also the spelling is incorrect.  It should be `Verify`

src/app/manager-dashboard/manager-dashboard.component.html
lmmrssa: yes it was not supposed to be there

src/app/upgrade/upgrade.component.html
lmmrssa: It will make user click verify to make sure upgrade was successful.

PR #3642, opened 2019-04-18
CoursesAdd: Save as draft locally

paulbert:
With this change courses will now save locally with PouchDB as a "draft".  Three actions will remove the draft:

1. Clicking "Submit" will save the course to CouchDB and delete PouchDB draft
2. Clicking "Cancel" will delete PouchDB draft
3. Adding an exam or survey will save the course to CouchDB and delete PouchDB draft

Note I also removed the `id` field for course steps because it is no longer used (we no longer use a `trackBy` function in CoursesAdd)


lmmrssa:
@paulbert course draft seems good. It is temporarily saving course and when trying to add new course it prefill course.
I believe this feature will be more useful even on creating exam and survey as that also takes long time.

Issue #3652, opened 2019-04-18
Collections: List alphabetically


paulbert:
We should make sure the list of collections is alphabetical.  I believe this uses the `StateService` and the functions on that can be supplied a `sort` parameter.

paulbert:
Even if it doesn't use `StateService` we should sort whatever request is made to CouchDB

PR #3656, opened 2019-04-19
Sort tags asc (connects #3652)

singharpita:


singharpita:
@yuviii There is indexing script in couchdb-setup.sh that you need to run.

yuviii:
@singharpita yeah i did it, but wondering is this the case with me only or others too. Cannot get the tags sorted still.

lmmrssa:
@yuviii sorting seems to be working.
@singharpita Looking at picture by @yuviii it looks like sorting is case sensitive. So can you make it case insensitive.
Or @paulbert shall we make tag creation limited to lower case only? Right now we can not create same tag twice even if we change case anyway.

yuviii:
@lmmrssa yeah u are correct. Its because of of case insensitive

Issue #2774, opened 2018-12-18
Fileserver outside couchdb


lmmrssa:
Create a way to save file outside couchdb for large files and serve them.
We should create a new filetype they can choose to upload it outside.

PR #3657, opened 2019-04-19
Upload script on nginx (connects #2774)

lmmrssa:
If requested with multipart-form request to `/upload`
It will save with the filename into fs folder


jessewashburn:
Seems like a very involved and no longer relevant WIP that others are unlikely to take over.

PR #3669, opened 2019-04-22
CoursesAdd: Use ResourcesComponent for Add Resources (fixes #3661)

### LINKED ISSUE #3661: CoursesAdd: Adding resource uses resource list page
Rather than opening the list of titles, adding a resource should open the full resource table.

paulbert:


lmmrssa:
We should also remove this mat-action on dialog box
Screen Shot 2019-04-25 at 21 31 09

Issue #3679, opened 2019-04-23
Resources: Title search as a keyword search


paulbert:
Currently the title text search is a string comparison the full text in the input against each resource title.

We should instead do the same search for each word in the input (split the string by spaces).  It should match resources that match all of the strings in the title of the resource.

You can add the new function for this in `table-helpers.ts`.

paulbert:
Will be similar to `filterSpecificFields` (please allow to specify fields so we can use it with other tables if needed).  You may be able to split the array, and then pass every string in to `filterSpecificFields` also.

PR #3683, opened 2019-04-24
Resource filter (connects #3679)

singharpita:


singharpita:
@paulbert Shall we try to implement it on other pages as well?

Issue #3650, opened 2019-04-18
Course: Survey on delete still shows as Take Survey


singharpita:
Steps to Replicate:
- Add survey from any course.
- Go to Survey and delete that survey from the list.
- Now go to Course . That survey is still shown as Take Survey and user can also submit that course.
- Now manage(edit) the above mentioned course and update that survey. Survey form is shown as empty.

paulbert:
If you actively delete a from the list which is connected to a course, a dialog should popup asking to confirm & if confirmed the course should be modified to remove that survey.

PR #3685, opened 2019-04-24
Survey delete from course on delete(connects #3650)

yubarajpoudel:


src/app/surveys/surveys.component.ts
```diff
@@ -1,7 +1,7 @@
 import { Component, OnInit, ViewChild, AfterViewInit, OnDestroy } from '@angular/core';
 import { Router, ActivatedRoute } from '@angular/router';
 import { MatTableDataSource, MatSort, MatPaginator, MatDialog, MatDialogRef, PageEvent } from '@angular/material';
-import { forkJoin, Subject, of } from 'rxjs';
+import { forkJoin, Subject, of, Observable, observable } from 'rxjs';
```
paulbert: These imports are not used

src/app/surveys/surveys.component.ts
```diff
@@ -145,35 +148,61 @@ export class SurveysComponent implements OnInit, AfterViewInit, OnDestroy {
...
 
+  createCourseArray(surveys) {
+    const courseArr = [];
+    const surveyCourse = surveys.filter((survey) => survey.course);
+    surveyCourse.forEach((element: any) => {
+      const indexAtArr = courseArr.findIndex( (mCourse) => mCourse._id === element.course._id);
+      const course = indexAtArr > -1 ? courseArr[indexAtArr] : element.course;
+      const index = this.findSurveyInSteps(course.steps, element);
+      if (index > -1) {
+        course.steps.splice(index, 1);
```
paulbert: Please remove the survey only, not the entire step

src/app/surveys/surveys.component.ts
```diff
@@ -145,35 +148,61 @@ export class SurveysComponent implements OnInit, AfterViewInit, OnDestroy {
...
+        }
       },
       onError: () => this.planetMessageService.showAlert('There was a problem deleting survey.')
     };
   }
 
   deleteSurvey(survey) {
     const { _id: surveyId, _rev: surveyRev } = survey;
+    if (survey.course) {
+      survey.course.steps.splice(this.findSurveyInSteps(survey.course.steps, survey), 1);
```
paulbert: Please remove the survey only, not the entire step

src/app/surveys/surveys.component.ts
```diff
@@ -102,6 +101,10 @@ export class SurveysComponent implements OnInit, AfterViewInit, OnDestroy {
     this.surveys.paginator = this.paginator;
   }
 
+  findSurveyInSteps(steps, survey) {
+    return steps.findIndex((step: any) => step.survey && step.survey._id === survey._id);
```
paulbert: I think rather than running this function many times you can revert to have it inside the `ngOnInit` and save the stepNum to the survey.

src/app/surveys/surveys.component.ts
```diff
@@ -145,35 +148,61 @@ export class SurveysComponent implements OnInit, AfterViewInit, OnDestroy {
     const selected = this.selection.selected.map(surveyId => findByIdInArray(this.surveys.data, surveyId));
     if (selected.length === 1) {
       const survey = selected[0];
-      this.openDeleteDialog(this.deleteSurvey(survey), 'single', survey.name);
+      this.openDeleteDialog(this.deleteSurvey(survey), 'single',
+       survey.name + ( survey.course ? ' will be deleted from ' + survey.course.courseTitle + ' aswell' : '' ));
```
paulbert: Indentation should be 2 spaces

src/app/surveys/surveys.component.ts
```diff
@@ -145,35 +148,61 @@ export class SurveysComponent implements OnInit, AfterViewInit, OnDestroy {
...
     } else {
       this.openDeleteDialog(this.deleteSurveys(selected), 'many', '');
     }
   }
 
+  createCourseArray(surveys) {
+    const courseArr = [];
+    const surveyCourse = surveys.filter((survey) => survey.course);
+    surveyCourse.forEach((element: any) => {
+      const indexAtArr = courseArr.findIndex( (mCourse) => mCourse._id === element.course._id);
```
paulbert: Extra space between after `findIndex(`

src/app/surveys/surveys.component.ts
```diff
@@ -145,36 +151,61 @@ export class SurveysComponent implements OnInit, AfterViewInit, OnDestroy {
...
+        }
       },
       onError: () => this.planetMessageService.showAlert('There was a problem deleting survey.')
     };
   }
 
   deleteSurvey(survey) {
     const { _id: surveyId, _rev: surveyRev } = survey;
+    if (survey.course) {
+      delete survey.course.steps[this.findSurveyInSteps(survey.course.steps, survey)]['survey'];
```
paulbert: Two things on this line:
- We shouldn't delete from the survey variable directly.  This is a reference to the data object in the table, so if the request fails the data in the table would be updated without the database changing.  In this situation it's not bad since this property isn't shown in the table, but it's more future proof if we avoid this completely.
- You've only completed half of my earlier comment.  This is the other half, where rather than calling `findSurveyInSteps` again you can just reference `survey.stepNum`

Issue #3672, opened 2019-04-23
Add Exam: Add Choice often below scroll for multiple choice


singharpita:
When type is changed from multiple single choice or multiple, the Add choice button is shown in bottom and its kind of hidden so can page be scroll to bottom when type changes to Multiple choice single or multiple answer?

paulbert:
Not sure how best to handle this since.  Maybe having "Add Choice" between title & description is easiest so everything is shown on most screens.

PR #3688, opened 2019-04-25
Move choice on exam (connects #3672)

singharpita:


paulbert:
Showed change to Richard and he chose a different layout.  Closing

PR #3689, opened 2019-04-25
Allow renaming of community/nation on parent (fixes #3684)

### LINKED ISSUE #3684: Change names for child planets
Allow the parent (e.g. Nation) to change the name of the child (e.g. Community) in the "Manage" list.  Shows different name both on this list and in reporting.

paulbert:


lmmrssa:
@paulbert this is creating a different document to rename the file. It is showing the name on the list by this new document but on report view it still shows name of the community not this new name.

paulbert:
@lmmrssa Which report view?  I'll fix tomorrow evening from my home computer since the development environment there is set up as a nation.

paulbert:
Also the intent with creating a new document is so the nation/center is not allowed to change the community/nation configuration directly, but they can add their own name to show in reports.

PR #3704, opened 2019-04-26
Questions: remove a title and display description (fixes #3701)

### LINKED ISSUE #3701: ExamsQuestion: Remove question title
Remove the question title field.  We should also see if we can use the description as the text in the question list instead of "Question 1", "Question 2", etc. so it's easier to manage moving the questions around.

The text should not overflow beyond one line, and we should not see any of the markdown syntax (ideally this is just plaintext).  I'm not sure if this is possible, so if it's taking too long to find a solution for this we can just remove the title and keep searching.

svlesiv:
Screen Shot 2019-04-26 at 11 19 33 AM
Screen Shot 2019-04-26 at 11 20 06 AM

Screen Shot 2019-04-26 at 11 52 46 AM
Screen Shot 2019-04-26 at 11 53 28 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "pouchdb-authentication": "^1.1.3",
     "pouchdb-find": "^7.0.0",
     "rxjs": "6.4.0",
+    "showdown": "1.6.4",
     "zone.js": "~0.8.26"
   },
   "devDependencies": {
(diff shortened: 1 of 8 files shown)


lmmrssa:
Screen Shot 2019-04-29 at 16 05 27

svlesiv:
Thank for the review @lmmrssa. I just checked it again, and I'm not getting console error.

svlesiv:
Thanks for your review and advice @paulbert. I will try to fix it.

Issue #3676, opened 2019-04-23
Course Add : Survey must have options for multiple choice or single choice


yuviii:
Currently the survey can be saved without adding the options as well for both the cases 
 1. Multiple choice multiple selection
 2. Multiple choice single selection

The survey should only be saved if it has options added for both of above mentioned cases.

PR #3705, opened 2019-04-28
Options for choice in survey (connects #3676)

yuviii:


lmmrssa:
@yuviii right now travis is failing because of the data type. So you need to fix that making sure that functions and data are using the correct type.

yuviii:
@lmmrssa @paulbert please review the update

PR #3757, opened 2019-05-04
CoursesView: Remove Method (fixes #3748)

### LINKED ISSUE #3748: CoursesView: Remove "Method"
We do not use "Method" and should remove it.

image

diabagatekelly:
I removed "Method" from CoursesView.


lmmrssa:
@diabagatekelly try not to use `git add .` instead use `git add <filename>` to avoid adding extra files.

diabagatekelly:
> 
> 
> Please make sure you do not touch other files. If have touched, then you need to make sure you roll back those file changes.
> You have added lot of extra new files and also changes on files not related to this issue.
> Also it looks like you do not have installed linting hook. To do that run `npm run install-hooks`

I had not when I worked on this, but I just added it now.

diabagatekelly:
> 
> 
> @diabagatekelly try not to use `git add .` instead use `git add <filename>` to avoid adding extra files.

Sorry about that. I will make sure to only git add the file I work on. Should I recommit and re-do the pull request?

Issue #3721, opened 2019-04-30
Notification: datetime is missing in Survey notification


singharpita:
![Screenshot from 2019-04-30 07-42-21](https://user-images.githubusercontent.com/27276132/56937315-b7982000-6b1b-11e9-880a-3d26f0585546.png)

lmmrssa:
Yes we need to add Date on notification

yuviii:
@lmmrssa @paulbert  There is already `time` field but
in `Submissions.service.ts`
```
const data = {
      'message': `<b>${this.userService.get().name}</b> has
        ${isRecorded ? 'recorded' : 'completed'} the survey <b>${this.submission.parent.name}</b>`,
      'link': '/submissions/exam',
      'linkParams': { submissionId: this.submission._id, questionNum: 1, status: 'complete', mode: 'view' },
      'type': 'survey',
      'priority': 1,
      'status': 'unread',
      'time': this.couchService.datePlaceholder
    };
```
the datePlaceholder is returning the object. Cannot find the problem why this is happening

it is saving in couch db

```
{
  "_id": "8e5806b38978b3c14aa01b3c380234ce",
  "_rev": "1-deffaaec5d25ad0138c12f500f5366ff",
  "message": "<b>asd</b> has\n        completed the survey <b>test1</b>",
  "link": "/submissions/exam",
  "linkParams": {
    "submissionId": "8e5806b38978b3c14aa01b3c38022b07",
    "questionNum": 1,
    "status": "complete",
    "mode": "view"
  },
  "type": "survey",
  "priority": 1,
  "status": "unread",
  "time": {},
  "user": "org.couchdb.user:yuvi"
}
```

lmmrssa:
@yuviii it is using dateplaceholder not the time. placeholder is used for input field to show date. In database it should save timestamp as we do for createdate, updateddate etc

paulbert:
@yuviii You use `couchService.datePlaceholder` in the object, and use `couchService.updateDocument` when adding the document to the database.

Not sure where this is best to document, but we should recommend to the entire team to use `updateDocument` because it automatically converts the `datePlaceholder` to an actual date.

PR #3754, opened 2019-05-03
Datetime issues fixed(connects #3721)

yuviii:

PR #3759, opened 2019-05-06
Time of survey notification (connects #3721)

lmmrssa:


lmmrssa:
Update document approach instead of #3754

Issue #3629, opened 2019-04-16
Create a view to show all myplanet devices connected


lmmrssa:
Show myplanet devices that has connected along with **version** **last upgrade** **last sync** etc

PR #3760, opened 2019-05-06
List myplanet (connects #3629)

lmmrssa:
Add a view on manager dashboard to look into list of myPlanet.
It will still require to group same tablets and may be other detail reports.
This PR just has a starting point to facilitate looking into myPlanet connected.


lmmrssa:
Sketch (17)
Sketch (16)

PR #3776, opened 2019-05-08
Resources: selected collections component (fixes #3699)

### LINKED ISSUE #3699: Resources: Move selected collections text
We have a request to move the selected collections text on the resources view so that it is next to (above or to the right of) the X of Y Resources Shown text here:

resourcesbar

This will be a challenge because the text and the button are part of their own reusable component.  A possible good solution is:

- Move the text into its own reusable component that takes the value of PlanetTagInputComponent as an @Input.
- Add that component in the blue toolbar and use a [template reference variable](https://angular.io/guide/template-syntax#template-reference-variables--var-) to pass the value of PlanetTagInputComponent in to this new component.
- We already have a `helperText` variable available to hide the text if set to `false`
- If this works, we should also use the new component in PlanetTagInputComponent

svlesiv:
Screen Shot 2019-05-08 at 10 52 17 AM

### DIFF
--- a/src/app/resources/resources.component.html
+++ b/src/app/resources/resources.component.html
@@ -26,7 +26,7 @@
     <ng-container *ngIf="isAuthorized">
       <button *ngIf="!parent" mat-mini-fab routerLink="add"><mat-icon>add</mat-icon></button>
       <planet-filtered-amount [table]="resources"></planet-filtered-amount>
-      <planet-tag-selected-input [selectedTags]="tag.value"></planet-tag-selected-input>
+      <planet-tag-selected-input [selectedIds]="tag.value" [parent]="parent"></planet-tag-selected-input>
       <span class="toolbar-fill"></span>
       <span *ngIf="parent">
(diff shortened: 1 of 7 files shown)


src/app/resources/resources.component.html
```diff
@@ -26,6 +26,7 @@
     <ng-container *ngIf="isAuthorized">
       <button *ngIf="!parent" mat-mini-fab routerLink="add"><mat-icon>add</mat-icon></button>
       <planet-filtered-amount [table]="resources"></planet-filtered-amount>
+      <planet-tag-selected-input [selectedIds]="tag.value" [parent]="parent"></planet-tag-selected-input>
```
paulbert: I realize you don't actually need the `#tag` for this, you can just set `selectedIds` to `tagFilter.value`

Also, we're running out of horizontal space in this toolbar.  Can you put this text below the "Showing X of Y resource(s)"?

svlesiv: Thanks for a good advice @paulbert.

svlesiv:
Screen Shot 2019-05-10 at 3 02 56 PM

PR #3782, opened 2019-05-09
Fix test question validation (fixes #3781)

### LINKED ISSUE #3781: Multiple choice option validation
Validation of multiple choice is now stopping even short answer and long answer.

paulbert:
#3781 

Still uses built in validator, but only for the correct question types


lmmrssa:
When adding multiple choice question (new) it is saving without any choices.
But on edit it will show validation for same question.

Issue #3766, opened 2019-05-07
Surveys: Send survey use name set by nation for list


paulbert:
When sending a survey to a community from nation (or nation from Earth), the list should use the name set on the nation instead of the one set on the community.

![image](https://user-images.githubusercontent.com/9203229/57266112-a7ff6680-7048-11e9-8a59-5dd6463dce18.png)

Note to work on this you will need to set up your development environment as a nation or center & set up another environment as a child planet for the development environment.

PR #3792, opened 2019-05-13
[#3766] Show community name on survey list

singharpita:


singharpita:
@paulbert had checked that option, reduceDropdown function on DialogListComponent is using tableData or dropdownSettings. So send planetName on tableData. Otherwise we will have to send the list of values from dropdownSetting.
Do you suggest to send it on dropdownSetting?

paulbert:
One way to do that would be to send in another property, `labels`, in `dropdownSettings` with an object as its value with each key a `planetCode` and each value the corresponding `planetName`.  Then in `reduceDropdowns` you could specify the text for the dropdown if `labels[value]` exists or use `value` if not.

PR #3812, opened 2019-05-15
Fix Course Progress's course chart label (fixes #3590)

### LINKED ISSUE #3590: Course Progress: step's course progress label is misleading
- go to course progress of any step.
- Label Steps should be changed to "Questions" or "No of Questions"
Screenshot from 2019-04-11 21-03-12
It might be misleading because full course also shows same label and it shows count of steps in it while within step's course progress it shows count of question in that particular step.

diabagatekelly:
I fixed Course Progress's chart so the label says 'Questions' instead of 'Steps' when the use navigates to the chart for specific questions.

### DIFF
--- a/src/app/courses/progress-courses/courses-progress-chart.component.html
+++ b/src/app/courses/progress-courses/courses-progress-chart.component.html
@@ -13,7 +13,7 @@
   </ng-container>
 </div>
 <div class="errors errors-index" #errorsIndex>
-  <div class="anchor-label" i18n>Steps</div>
+  <div class="anchor-label" i18n>{{label}}</div>
   <div
     class="border-bottom"
     matTooltip="Click to see test detail"
--- a/src/app/courses/progress-courses/courses-progress-chart.component.html
+++ b/src/app/courses/progress-courses/courses-progress-chart.component.html
@@ -13,7 +13,7 @@
   </ng-container>
 </div>
 <div class="errors errors-index" #errorsIndex>
-  <div class="anchor-label" i18n>{{label}}</div>
+  <div class="anchor-label" i18n>{label, select, Steps {Steps} Quest. {Quest.}}</div>
   <div
     class="border-bottom"
     matTooltip="Click to see test detail"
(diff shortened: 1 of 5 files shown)


src/app/courses/progress-courses/courses-progress-chart.component.html
```diff
@@ -13,7 +13,7 @@
   </ng-container>
 </div>
 <div class="errors errors-index" #errorsIndex>
-  <div class="anchor-label" i18n>Steps</div>
+  <div class="anchor-label" i18n>{{label}}</div>
```
paulbert: Instead of directly showing the label variable in the template, you should change this to use the i18n select syntax.  This will allow the text to be translated.

https://angular.io/guide/i18n#select-among-alternative-text-messages

src/app/courses/progress-courses/courses-progress-leader.component.ts
```diff
@@ -48,6 +49,7 @@ export class CoursesProgressLeaderComponent implements OnInit, OnDestroy {
       this.filterSubmittedExamSteps(submissions);
       this.dialogsLoadingService.stop();
     });
+    this.chartLabel = 'Steps';
```
paulbert: You can just initialize the value to `'Steps'` on line 19 rather than adding this here.

src/app/courses/progress-courses/courses-progress-chart.scss
```diff
@@ -2,7 +2,7 @@
 
 :host {
 
-  $cell-size: 45px;
+  $cell-size: 71px;
```
paulbert: This leaves too much space around the numbers in the chart.  We should either abbreviate question or maybe rotate the text to fit.

src/app/courses/progress-courses/courses-progress-chart.component.html
diabagatekelly: I think I got this right. I wasn't sure if I needed to change anything in the i18n folder so I just left it alone.

src/app/courses/progress-courses/courses-progress-leader.component.ts
diabagatekelly: Got it

Issue #3813, opened 2019-05-15
Issue on sending courses and resources with older tag.


lmmrssa:
We recently change way tag works but sending (fetch) has not been updated to reflect that which is causing issue.

paulbert:
It's not an issue with the tags, but how the data is organized in the table that has changed

The below is an example of a document in `send_docs` created with 0.7.39.  The value in `item.doc` should be the value for `item` for the fetch on the community end to work.
![image](https://user-images.githubusercontent.com/9203229/57864865-9ea19700-77ca-11e9-875d-932034fb0cba.png)

PR #3839, opened 2019-05-17
Send Fetch course resource (connects #3813)

lmmrssa:


lmmrssa:
@paulbert yes i had thought about that as well. So was already looking other way. But trying to look all other aspect as we also have achievement which we sync. So it should work with all and not add extra details on sync item.

Issue #3828, opened 2019-05-16
Courses & Resources: Blue toolbar overflow


paulbert:
We've added many buttons to the blue toolbar of courses & resources so they are now overflowing.  We should combing the manager related ones in to one button (labeled "Manager Actions") which opens a menu with those tasks.  Note while "Change Collections" is restricted to managers we can leave that one out since adding this custom component to a `mat-menu` would be challenging.

For both resources and courses these three buttons can be in that menu:
- Push to ...
- Send ...
- Delete

![image](https://user-images.githubusercontent.com/9203229/57867198-ab27ee80-77ce-11e9-9504-3719aa791998.png)

paulbert:
Note the "Push to community" button should retain its other condition (planet is not a community)

YuserahN:
@paulbert Am I in the right place? 

![planet_learning_dev_resources](https://user-images.githubusercontent.com/21087353/57875162-d6fba200-77d7-11e9-8062-fe955f12b7c6.PNG)

paulbert:
On a community when logged in as administrator you should see in the resources toolbar:
![image](https://user-images.githubusercontent.com/9203229/57934453-c60d6800-788d-11e9-9c70-8f9d662cb789.png)

And on nation you should see:
![image](https://user-images.githubusercontent.com/9203229/57934516-e3423680-788d-11e9-8871-655efb8b1fb9.png)

PR #3845, opened 2019-05-17
Courses & Resources: Combine Toolbar Manager Actions Into Menu

YuserahN:
#3828


YuserahN:
@paulbert Should 'Manager Actions' be disabled if all of the menu options under it are disabled? (And it's enabled if at least one of the menu items is enabled.)

lmmrssa:
@YuserahN when logged in as non-admin and there is no button displayed on resource, Manager action still shows.

Issue #3856, opened 2019-05-21
Manager Actions: In 'Send Resource,' Show Number Of Selected Resources


YuserahN:
Under 'Manager Actions,' 'Send Courses' shows the number of selected courses. The same should be done for 'Send resource.' 

![send_courses_planet](https://user-images.githubusercontent.com/21087353/58062755-28da5a00-7b41-11e9-8b49-e4f66b5509ae.PNG)

In this case, the following should display 'Send Resources (1).' 

![send_resources_planet](https://user-images.githubusercontent.com/21087353/58062760-2e37a480-7b41-11e9-8c64-29dd4d865162.PNG)

paulbert:
Modifying the same area of the code as #3853 

If both are open, feel free to fix in same PR

PR #3868, opened 2019-05-21
[#3856] Send Resource show no of resource (connects #3856)

singharpita:
included this issue as well : https://github.com/open-learning-exchange/planet/issues/3853


paulbert:
Ok, let's only limit resources sent to parent/nation if they do not have an attachment.

singharpita:
@paulbert made changes to only send resource with attachment (to parent). Shall we also implement it on sending it to community (child)?

Issue #3884, opened 2019-05-23
Preview with empty question


lmmrssa:
When we have created multiple questions but have not filled question details on any question. Question description of previous question shows.
May we should only show questions that are valid.

paulbert:
I think this will be fixed in my proposal for #3880, but leaving open in case not.

Best thing would be for preview to only open if the exam is valid.

Issue #3880, opened 2019-05-23
Course manage: Preview Test when empty shows inconsistent UI


singharpita:
![Screenshot from 2019-05-23 20-25-58](https://user-images.githubusercontent.com/27276132/58262017-23774c80-7d99-11e9-84c7-ab7769412bcd.png)
- there should be space between "questionof"
- Question of should be hidden and empty message should be shown.
- Close preview button should be replaced by "X" icon.

paulbert:
This is breaking the UI it appears, and nothing should show, anyway.  If there are no steps or there are form validation errors, we should not allow Preview.

PR #3891, opened 2019-05-25
Preview is Restricted to Valid Exams

YuserahN:
#3880 & #3884


YuserahN:
@paulbert Do you want me to disable the 'Preview' button if the exam is invalid? The current PR allows the user to click on the button, but it doesn't do anything if the exam is invalid. (It's clickable, but the preview is not displayed.)

PR #3926, opened 2019-05-29
Collections: fixed console error when creating a new collection (fixes #3905)

### LINKED ISSUE #3905: Console error when creating a new collection (resources, courses)
Screen Shot 2019-05-27 at 7 26 35 PM

https://www.loom.com/share/e83c8d8ec38c475c904eeec25f6cd65a

https://www.loom.com/share/bcd4a9efb7384cb888b2160c731925e1

svlesiv:


paulbert:
I did a little more reading on `FormControl.reset` and it looks like
1. You can supply a value to set the form control
2. `markAsPristine` and `markAsUntouched` are actually run as a part of that function

Made changes based on these two findings to simplify

See Angular code here:
https://github.com/angular/angular/blob/8.0.0/packages/forms/src/model.ts#L1067-L1092

PR #3947, opened 2019-06-03
Add message to exam preview on completion (fixes #3885)

### LINKED ISSUE #3885: ExamsAdd: Preview message after last question
We need to have some sort of UI feedback when the exam is completed in preview mode in the dialog in ExamsAdd

Maybe where the incorrect answer shows we could just have a message that says "Exam complete" after clicking the button to submit the last answer.

diabagatekelly:
Added a message to the exam preview window when the last question is correct.

### DIFF
--- a/src/app/exams/exams-view.component.html
+++ b/src/app/exams/exams-view.component.html
@@ -66,15 +66,14 @@
         <ng-container *ngIf="isComplete" i18n>Complete</ng-container>
         <ng-container *ngIf="!isComplete" i18n>Submit Answer</ng-container>
       </button>
-      <span class="v-align-center small warn-text-color" [ngSwitch]="examPreviewStatus">
-        <div *ngSwitchCase="'incomplete'">
+      <span class="v-align-center small" [ngSwitch]="statusMessage">
+        <div *ngSwitchCase="'incorrect'" class="warn-text-color">
           <mat-icon class="margin-lr-3">error</mat-icon><span i18n>Incorrect answer, please try again</span>
         </div>
-        <div *ngSwitchCase="'complete'">
-          <mat-icon class="margin-lr-3">error</mat-icon><span i18n>Exam complete, you can close the preview</span>
+        <div *ngSwitchCase="'complete'" class="primary-text-color">
+          <mat-icon class="margin-lr-3">check_circle</mat-icon><span i18n>Exam complete, you can close the preview</span>
         </div>
       </span>
-
     </div>
   </div>
 </div>
--- a/src/app/exams/exams-view.component.html
+++ b/src/app/exams/exams-view.component.html
@@ -66,9 +66,15 @@
         <ng-container *ngIf="isComplete" i18n>Complete</ng-container>
         <ng-container *ngIf="!isComplete" i18n>Submit Answer</ng-container>
       </button>
-      <span class="small warn-text-color v-align-center" *ngIf="incorrectAnswer">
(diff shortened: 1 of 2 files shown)


src/app/exams/exams-view.component.ts
```diff
@@ -118,6 +121,9 @@ export class ExamsViewComponent implements OnInit, OnDestroy {
       return;
     }
     if (this.isDialog) {
+      if (nextQuestion > this.maxQuestions - 1) {
```
paulbert: With the `return` above on line 121, this conditional will actually always be true so you can remove it.  I think I suggested this, but missed that with the return in the if block above the exam will always be complete.

src/app/exams/exams-view.component.ts
```diff
@@ -27,6 +27,7 @@ export class ExamsViewComponent implements OnInit, OnDestroy {
   maxQuestions = 0;
   answer = new FormControl(null, this.answerValidator);
   incorrectAnswer = false;
+  examPreviewStatus = '';
```
paulbert: Let's rename this variable to `statusMessage` and the two possible values for it should be `incorrect` and `complete`

src/app/exams/exams-view.component.html
```diff
@@ -66,9 +66,15 @@
...
-        <mat-icon class="margin-lr-3">error</mat-icon><span i18n>Incorrect answer, please try again</span>
+      <span class="v-align-center small warn-text-color" [ngSwitch]="examPreviewStatus">
+        <div *ngSwitchCase="'incomplete'">
+          <mat-icon class="margin-lr-3">error</mat-icon><span i18n>Incorrect answer, please try again</span>
+        </div>
+        <div *ngSwitchCase="'complete'">
+          <mat-icon class="margin-lr-3">error</mat-icon><span i18n>Exam complete, you can close the preview</span>
+        </div>
       </span>
+
```
paulbert: Please remove the extra line

src/app/exams/exams-view.component.html
```diff
@@ -66,9 +66,15 @@
...
         <ng-container *ngIf="!isComplete" i18n>Submit Answer</ng-container>
       </button>
-      <span class="small warn-text-color v-align-center" *ngIf="incorrectAnswer">
-        <mat-icon class="margin-lr-3">error</mat-icon><span i18n>Incorrect answer, please try again</span>
+      <span class="v-align-center small warn-text-color" [ngSwitch]="examPreviewStatus">
+        <div *ngSwitchCase="'incomplete'">
+          <mat-icon class="margin-lr-3">error</mat-icon><span i18n>Incorrect answer, please try again</span>
+        </div>
+        <div *ngSwitchCase="'complete'">
+          <mat-icon class="margin-lr-3">error</mat-icon><span i18n>Exam complete, you can close the preview</span>
```
paulbert: This line should use the `primary-text-color` and the icon should be `check_circle`.

src/app/exams/exams-view.component.ts
```diff
@@ -104,6 +106,7 @@ export class ExamsViewComponent implements OnInit, OnDestroy {
     obs.subscribe(({ nextQuestion }) => {
       if (correctAnswer === false) {
         this.incorrectAnswer = true;
```
paulbert: `incorrectAnswer` has been replaced by the new status variable, so you can remove it.  Note that you will also have to reset the value of the status variable to an empty string in line 86 of this file.

diabagatekelly:
Hello @paulbert, I've made the required changes.

PR #3949, opened 2019-06-04
3851 resources add file validation

### LINKED ISSUE #3851: ResourcesAdd: Open file validation updates
The "Open which file" field should only apply when a zip file is being uploaded.  We should make some updates to help keep this field valid:

1. If the uploaded file is not a zip file, this should be empty & disabled
2. If the uploaded file is a zip file, the text should match a filename in the zip file

image

ArashDai:
fixes #3851

### DIFF
--- a/src/app/resources/resources-add.component.html
+++ b/src/app/resources/resources-add.component.html
@@ -79,11 +79,12 @@
           <input matInput i18n-placeholder [(ngModel)]="userDetail.name" placeholder="Added By" formControlName="addedBy" readonly>
         </mat-form-field>
         <mat-form-field>
-          <mat-select i18n-placeholder placeholder="Open Which File" formControlName="openWhichFile">
-            <mat-option *ngFor="let file of attachedZipFiles" [value]="file" >
+          <input type="text" i18n-placeholder placeholder="Open Which File" matInput formControlName="openWhichFile" [matAutocomplete]="auto">
+          <mat-autocomplete #auto="matAutocomplete">
+            <mat-option *ngFor="let file of filteredZipFiles | async" [value]="file">
               {{file}}
             </mat-option>
-          </mat-select>
+          </mat-autocomplete>
         </mat-form-field>
         <div class="inner-gaps by-column full-width">
           <label i18n>File Upload:</label>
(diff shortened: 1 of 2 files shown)


src/app/resources/resources-add.component.ts
```diff
@@ -29,6 +29,7 @@ import { TagsService } from '../shared/forms/tags.service';
 export class ResourcesAddComponent implements OnInit {
   constants = constants;
   file: any;
+  attachedZipFiles: Array<String>;
```
paulbert: Let's stick to the syntax we're using in other parts of the app: `string[]`

src/app/resources/resources-add.component.ts
```diff
@@ -257,6 +262,22 @@ export class ResourcesAddComponent implements OnInit {
     this.disableDownload = false;
     this.disableDelete = false;
     this.resourceForm.updateValueAndValidity();
+
+    if (this.file.type === 'application/zip') {
```
paulbert: Please use the `simpleMediaType` method.  Some zip files have different MIME types than this.

src/app/resources/resources-add.component.ts
```diff
@@ -257,6 +262,22 @@ export class ResourcesAddComponent implements OnInit {
...
+      // If the uploaded file is a zip, update attachedZipFiles to show options in openWhichFile dropdown
+
+      this.resourceForm.controls.openWhichFile.enable();
+      const fileObs: Observable<any> = this.createFileObs();
```
paulbert: This unpacks the zip file into memory, which could slow things down for larger files.  The first part of `zipObs` gets a list of the file names, so you could split that into two functions/observables and use only the file name part here.

No need to assign it to a variable, either.

src/app/resources/resources-add.component.html
```diff
@@ -78,7 +78,11 @@
         <input matInput i18n-placeholder [(ngModel)]="userDetail.name" placeholder="Added By" formControlName="addedBy" readonly>
       </mat-form-field>
       <mat-form-field>
-        <input matInput  i18n-placeholder placeholder="Open Which File" formControlName="openWhichFile">
+        <mat-select i18n-placeholder placeholder="Open Which File" formControlName="openWhichFile">
```
paulbert: Trying this out I think it might be to use [autocomplete](https://material.angular.io/components/autocomplete/overview) instead of a select.  Long select lists are a little difficult to navigate.

ArashDai: autocomplete with filter looks good, ill try that out

Issue #3904, opened 2019-05-28
Adding Exams: correctChoice Does Not Update If Choice Removed


YuserahN:
When creating an exam and adding a multiple choice question, removing a choice you've already marked as the correct choice doesn't remove it from the correctChoice object. The question is still counted as having a correct choice. Therefore, it allows you to add the question and exam without selecting another correct choice for that question. (If a new choice is marked as the correct choice, then the previous one is overwritten. But this still doesn't account for the case where someone may not mark a new choice as the correct choice. In that case, the user ends up with a question where the correct choice that's stored in the database is not listed as an option in the UI.) 

https://www.loom.com/share/4beeff4faa5044f59a372d4b0d6f7a20

PR #3957, opened 2019-06-04
Adding Exams: Update correctChoices Upon Removal Of Choice

YuserahN:
#3904

### DIFF
--- a/src/app/exams/exams-question.component.ts
+++ b/src/app/exams/exams-question.component.ts
@@ -55,6 +55,11 @@ export class ExamsQuestionComponent implements OnInit, OnChanges {
   }
 
   removeChoice(index: number) {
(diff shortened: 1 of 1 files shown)


src/app/exams/exams-question.component.ts
```diff
@@ -55,6 +55,11 @@ export class ExamsQuestionComponent implements OnInit, OnChanges {
   }
 
   removeChoice(index: number) {
+    const correctCh = this.questionForm.controls.correctChoice.value;
+    const correctChoiceIndex = correctCh.indexOf(this.choices.value[index].id);
+    if (correctCh[correctChoiceIndex] === this.choices.value[index].id) {
```
paulbert: A simpler condition which is equivalent is `correctChoiceIndex > -1`.

PR #3966, opened 2019-06-04
Added space - Issue#3950

samuelchen1213:
Screen Shot 2019-06-04 at 11 30 35 AM

### DIFF
--- a/src/app/exams/exams-view.component.html
+++ b/src/app/exams/exams-view.component.html
@@ -8,7 +8,7 @@
   <mat-toolbar class="primary-color font-size-1">
     <span class="ellipsis-title">{{title ? title + ': ' : ''}}</span>
     <span>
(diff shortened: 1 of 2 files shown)


samuelchen1213:
Understood, I will fix this!

samuelchen1213:
@paulbert @lmmrssa  Would it be more beneficial to create a complete different class in styles.scss just for `white-space:pre` or can I add it to the `ellipsis-title` class (Or would this interfere with styling with other components)

paulbert:
@samuelchen1213 The only usage that might have an issue is in the CoursesStepViewComponent (click a course to open CoursesViewComponent and then click yellow "View Step" button), so I think adding it to `.ellipsis-title` would work.

samuelchen1213:
@paulbert Perfect! I did not want to cause future issues. My pull request will be on its way in a minute!

PR #3968, opened 2019-06-04
Collections: Update layout

paulbert:
Updates layout & subcollections

Subcollections are now unique to the attached collection.  The upside is that this is less confusing: clicking on a subcollection doesn't mark the checkbox for a bunch of items in the list.  The downside is that this will require more collections on each planet and will likely end up with some duplication.

TODOs:
- Fix subcollection showing id instead of name
- Better UI for subcollection in list of resources/courses

### DIFF
--- a/src/app/courses/courses.component.html
+++ b/src/app/courses/courses.component.html
@@ -100,9 +100,14 @@
             <mat-icon>more_vert</mat-icon>
           </button>
           <mat-chip-list #tagsList class="tags-list">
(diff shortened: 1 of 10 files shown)


paulbert:
@lmmrssa Fixed those todos, and I think this is ready.

I'm not sold on this solution for subcollections in the lists, but my other thought was having the subcollections shown in a dropdown after clicking the collection chip which is worse.

Maybe we could have a future issue to create dividers between the collection/subcollection sets so it's clearer that they are related.

PR #3972, opened 2019-06-05
Add class for rating bar for different color (fixes #3961)

### LINKED ISSUE #3961: Rating bar gets hidden
Row selection and Rating bar for undefined is using same color. So rating bar seems hidden when row is selected.
Sketch (45)
Sketch (46)

samuelchen1213:
Important notes:
1. Added another color to `_variables.scss` called `secondary light`
2. Created new class in `styles.scss` called `secondary-light-color` (Here I removed highlight so it would not effect the highlight color - this was the problem of why they both would have the same color. I created a new class to not interfere with primary-light. Please tell me if I am allowed to do this)
3. Altered rating bar class to `secondary-light-color`
Screen Shot 2019-06-05 at 2 01 52 PM

**Comment any changes neccesary, Thank you!**

### DIFF
--- a/src/app/_variables.scss
+++ b/src/app/_variables.scss
@@ -24,8 +24,8 @@ $primary-text: mat-contrast($planet-primary, $primary-hue);
 $accent-text: mat-contrast($planet-accent, $accent-hue);
 $light-grey: mat-color($mat-grey, 100);
 $grey: mat-color($mat-grey, 400);
-$primary-light: mat-color($planet-app-primary, 100);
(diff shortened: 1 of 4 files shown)


paulbert:
@samuelchen1213 The "primary" part of the class denotes the base color.  This new color should use the variable `$primary-light` and class `.primary-light-color` and the old one should be refactored to `$primary-lighter` and `.primary-lighter-color` so the second word describes the relative hue to the base color.

PR #4000, opened 2019-06-07
Teams: Syncing teams

paulbert:
Very much an experimental branch right now.  Opening this PR so we can have a place for comments


paulbert:
Messages do not sync, but I think this is in a good place for a review & merge if it works.

paulbert:
@lmmrssa Well I implemented messages syncing now, too.  I don't want to introduce these next changes to try to avoid conflicts with #4029, but the next to items on this task are:

- Fix profile images in messages for users not on local planet.  Short term we should make sure the default rather than an image not found icon shows up here, but long term we should think about replicating the user profile images.
- Show the planet code or name next to the name on a message so it's clear that message is from nation or community

PR #4001, opened 2019-06-07
Adds button margins to "add course", "update course", and "add meetups" (Fixes #3985)

### LINKED ISSUE #3985: No button margins on certain pages
The "Add new Course" page and the "Update Course" pages should have margins between the buttons of "Submit", "Cancel", and "Add Step".

Capture

Capture2

sjson421:
Fixes Issue #3985.

Adds 2px button margins to "add course", "update course", and "add meetups".

Capture

### DIFF
--- a/src/app/courses/add-courses/courses-add.scss
+++ b/src/app/courses/add-courses/courses-add.scss
@@ -1,5 +1,4 @@
 @import '../../variables';
-@import '../../../styles.scss';
 
 :host {
   .view-container {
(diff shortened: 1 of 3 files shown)


src/app/courses/add-courses/courses-add.scss
```diff
@@ -1,4 +1,5 @@
 @import '../../variables';
+@import '../../../styles.scss';
```
paulbert: @sjson421 You don't actually need to import the `styles.scss` files because that css is included broadly as a sort of global css.

sjson421: @paulbert Hmm I don't why why my IDE yelled at me there. Fixed!

PR #4029, opened 2019-06-12
News item reply

paulbert:
Still a work in progress, but @lmmrssa wanted to give you a chance to see the direction I was thinking and give feedback.


paulbert:
@lmmrssa OK, I was thinking 2 & 3, and good point on 1.

I think rather than removing the ability to delete, we can update the other messages up the chain of replies.  That way when we get in to having managers/leaders moderate these posts they won't be stuck having to delete other messages to get at one that needs to be removed.

PR #4032, opened 2019-06-12
Development serve for different languages (fixes #2363)

### LINKED ISSUE #2363: Clean up translation tags
The only content within an html tag with the `i18n` tag should be text or an i18n select ([see more here](https://angular.io/guide/i18n#translate-select)).  There should be no angular variable interpolation like the below:
```
<span i18n>Planet: {{planetCode}}</span>
```
Instead of the above we should have the text in its own tag and the interpolation in another:
```
<span i18n>Planet:</span><span> {{planetCode}}</span>
```

We try to avoid this, but it seems there may be instances where this has slipped through.  Someone should go through and ensure that these tags are implemented correctly for translation.

paulbert:
Allows for viewing some of the different language files in the development server with one of the following commands:

```
ng serve --configuration=somali
ng serve --configuration=french
ng serve --configuration=nepali
ng serve --configuration=arabic
```

Made the change to review another PR anyway, so figure I can share so others can use this.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -55,6 +55,30 @@
                   "with": "src/environments/environment.test.ts"
                 }
               ]
+            },
+            "somali": {
+              "aot": true,
+              "i18nFile": "src/i18n/messages.som.xlf",
+              "i18nFormat": "xlf",
+              "i18nLocale": "so"
+            },
+            "french": {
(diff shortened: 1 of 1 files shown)


paulbert:
#2362

PR #4053, opened 2019-06-16
Fixes #3986 [Collections: Disable "Subcollection of" if collection has attached subcollections]

### LINKED ISSUE #3986: Collections: Disable "Subcollection of" if collection has attached subcollections
Making a collection which has other collections attached a subcollection of a different collection causes the other subcollections to disappear from the list & collections.

We should not allow a collection with subcollections to have its value for "Subcollection of" changed.

For a little more detail on the problem, let's say we have two collections, A and B, and B has two subcollections, C and D.  The hierarchy is:
```
A   B
    |--|
    C  D
```
If we then made B a subcollection of A, the hierarchy becomes:
```
A
|
B
|--|
C  D
```
This makes three tiers of collections, but the application will only show two tiers (this is a design decision to keep the collection structure from getting too convoluted).  C and D are no longer present in the application, but still exist in the database.  We can avoid this by not allowing B to become a subcollection until C and D are detached.

coder8102:
Fixes #3986 [Collections: Disable "Subcollection of" if collection has attached subcollections]

issue-3986-1

issue-3986-2


lmmrssa:
@coder8102 It is not what was supposed to be implemented. When collection has child it should not have option to add as subcollection.
Solution is to have none as option of collection has child.
And when you are trying to add anything to subcollection list should show collection of top level only (no sub-collection)

Right now it is filtering collection which has subcollection not shown as option.

coder8102:
> @coder8102 It is not what was supposed to be implemented. When collection has child it should not have option to add as subcollection.
> Solution is to have none as option of collection has child.
> And when you are trying to add anything to subcollection list should show collection of top level only (no sub-collection)
> 
> Right now it is filtering collection which has subcollection not shown as option.

Thanks Laxman. Made changes as suggested.

PR #4054, opened 2019-06-17
Update package.json

lmmrssa:


dogi:
@lmmrssa @leomaxi new version of myplanet is not updated

Issue #3936, opened 2019-05-30
Feedback title for take exam url


singharpita:
right now feedback is submitted for the current page.
in case of take exam from course step, title seems quite long.
![Screenshot from 2019-05-30 21-09-56](https://user-images.githubusercontent.com/27276132/58643742-96e00780-831f-11e9-8b69-8d98236b9188.png)

its following default functionality but  seems bit confusing.

paulbert:
To make this view look a little better, we should convert the parameters (separated by `;` at the end) into an object and display that separately from the rest of the url.

PR #4066, opened 2019-06-18
feedback title truncation (connects #3936)

Liew211:
Screenshot_2019-06-17 Planet Learning

Truncated name, turned params into an object.  I'm not sure of the exact method of display, so I used JSON.stringify for a temporary fix, so please let me know what you'd like best for this.

### DIFF
--- a/src/app/feedback/feedback-view.component.html
+++ b/src/app/feedback/feedback-view.component.html
@@ -40,7 +40,12 @@
     </mat-toolbar-row>
   </mat-toolbar>
   <div class="view-container view-full-height">
-    <div class="feedback-params" *ngFor="let prop of feedback.params | keyvalue">{{prop.key}}: {{prop.value}}</div>
+    <div *ngIf="showParams==='on'">
+      <div class="margin-tb" *ngFor="let prop of feedback.params | keyvalue">
+        <div>{{prop.key}}: {{prop.value}}</div>
+      </div>
+    </div>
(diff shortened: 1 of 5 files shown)


Liew211:
Screenshot_2019-06-18 Planet Learning
This is what the page looks like now.

paulbert:
@Liew211 If there are parameters, it would be better to only show them after clicking a "Show Page Parameters" button

Liew211:
Screenshot_2019-06-24 Planet Learning
Screenshot_2019-06-24 Planet Learning(1)
I partly copied the toggle functionality from the search filter in resources, let me know if you want the button to be above/below the details, and if the size of the button should be smaller.

paulbert:
I think the button should be in the top blue toolbar after the edit button and maybe say "Show URL Details" or "Show Page Details"

Also I think the details should have a different background like with resources/courses so it stands out as different.

Liew211:
@paulbert would a mat-card be ok for the details?  Also, should the button be aligned to the right, or just be next to the edit button?

Liew211:
Screenshot_2019-06-28 Planet Learning
@paulbert how's this?

paulbert:
@Liew211 I like what you have in that screenshot.  I think maybe add a line on top in bold that says **Page Parameters** as a heading and on the show/hide details button add `*planetAuthorizedRoles` to limit this to admins.  I think it would just be confusing for normal members.

Liew211:
@paulbert made the changes, take a look!

PR #4068, opened 2019-06-18
4041 team adding message inside message replies

anishregmee:
solved issue inside teams and ourNews. @lmmrssa @paulbert

PR #4079, opened 2019-06-20
Add margin for "Show main conversation"(Fix #4038)

### LINKED ISSUE #4038: NewsList: Add top margin for reply buttons
When you are viewing a reply for messages in ourNews or a Team, there is not space between the content above and the buttons.

We should add a small margin, `0.5rem`, to the top of these buttons.

image

samuelchen1213:
In news, when replying to reply of messages, a margin of 0.5rem was added to the "Show main conversation" button


samuelchen1213:
@lmmrssa My new solution was to use in line styling for the button. Is this permitted?

singharpita:
@samuelchen1213 we should not use inline styling.

paulbert:
Oops didn't realize this issue would no longer be relevant after #4068.  Closing.

PR #4105, opened 2019-06-24
Remove myLibrary button from library page (fixes 4102)

anishregmee:


src/app/resources/resources.component.html
```diff
@@ -1,10 +1,7 @@
 <mat-toolbar>
   <mat-toolbar-row>
     <button mat-icon-button (click)="goBack()" *ngIf="!isDialog"><mat-icon>arrow_back</mat-icon></button>
-    <span i18n class="margin-lr">Library</span>
-    <button *ngIf="!parent && !isDialog" mat-stroked-button class="margin-lr-3" (click)="toggleMyLibrary()" i18n>
-      {myLibraryFilter, select, on {Community} off {myLibrary}} resources
-    </button>
+    <span *ngIf="!parent && !isDialog" i18n class="margin-lr">{myLibraryFilter, select, on {my} off {}}Library</span>
```
lmmrssa: make the whole word to be in translation, no need to have part of it and also no need for empty `{}`

paulbert: It looks like Library is in the translation within the i18n tag, but I wonder if the below would be more easily translated in CrowdIn?

```
{myLibraryFilter, select, on {myLibrary} off {Library}}
```

lmmrssa: oh there should not be condition of !parent or !dialog as it was for button we do not have button here.

anishregmee: please check the issue now... thank you in advance @lmmrssa @paulbert

PR #4106, opened 2019-06-24
Menu: Add sync button for side menu (fixes 4099)

anishregmee:


src/app/home/home.component.html
```diff
@@ -123,17 +123,25 @@ <h1 i18n>Planet Learning</h1>
...
+            <label i18n>Feedback</label>
+          </a>
+        </li>
         <li>
           <a mat-button routerLink="/surveys" i18n-title title="Surveys">
             <mat-icon svgIcon="surveys"></mat-icon>
             <label i18n>Surveys</label>
           </a>
         </li>
         <li>
```
lmmrssa: li is not required to create if condition does not match
Also you can just add auth directive to li no need for ng-container

lmmrssa: oh it also needs to check if planet is **online**

anishregmee: ok. please check the changes now.. thanx in advance @lmmrssa

PR #4132, opened 2019-06-25
Fix team message replies (fixes # 4040)

diabagatekelly:
I added a background color to the replies.


paulbert:
@diabagatekelly Instead of this let's try some padding & a border around the main message (the one directly under "Viewing replies for:")

You should make a new scss file `news-list.scss` so you can use `@import '../variables.scss'` and access the `$light-grey` or `$grey` color variables for the border color.  Also you can use `border-radius: 4px` to match the curved edges of the `mat-card`.

PR #4137, opened 2019-06-27
Increase protractor timeout

paulbert:
Try to fix the of late common Travis failure in the automated test stage. Increases the test timeout from 11s to 110s

### DIFF
--- a/protractor.conf.js
+++ b/protractor.conf.js
@@ -7,7 +7,7 @@ const { SpecReporter } = require('jasmine-spec-reporter'),
   user = userHandler(timeStamp);
 
 exports.config = {
-  allScriptsTimeout: 11000,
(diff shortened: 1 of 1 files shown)


protractor.conf.js
```diff
@@ -32,7 +32,7 @@ exports.config = {
     });
     jasmine.getEnv().addReporter(new SpecReporter({ spec: { displayStacktrace: true } }));
     browser.params.user = user.get();
-
```
paulbert: Hmm...my IDE automatically removes spaces in empty lines creating this change.  @lmmrssa Go ahead and change this if you want.

lmmrssa: yes this is good we should not have spaces on empty line anyway.

Issue #4120, opened 2019-06-25
Team message reply not syncing


lmmrssa:
If there is reply to any message in team, it does not sync between nation and community. It is happening both ways. Main message and notification sync fine.

PR #4150, opened 2019-06-27
Add sync to news replies (fix #4035)

### LINKED ISSUE #4035: Team Messages: Save draft with PouchDB like courses
If you start writing a message and then close the dialog you will lose your message.

For CoursesAdd we implemented saving a draft locally with PouchDB.  We can do the same thing for team messages so work is not lost.

Note the `db` supplied to PouchDB needs to be unique per team.  Including the team id in that string would probably do it.

samuelchen1213:


samuelchen1213:
#4120 fixes this

Issue #4103, opened 2019-06-24
Make beta dashboard as default dashboard


lmmrssa:
Dashboard should have label of their community such as somalia should have Somalia Planet. And make the Kismayo Community center to be dynamic

lmmrssa:
We need to implement Submissions, mySurvey in this view as well.

paulbert:
@anishregmee This needs a lot more work.  I'll be taking over on it today.  Thanks for starting on it.

PR #4159, opened 2019-06-27
Include community name on top of beta dashboard (connects #4103)

paulbert:
Replaces placeholder with the name property of the planet configuration


paulbert:
@lmmrssa This uses the name in the configuration, so it can be renamed by the admin.

PR #4162, opened 2019-06-28
2675 title for my progress

anishregmee:


src/app/courses/progress-courses/courses-progress-learner.component.html
```diff
@@ -4,7 +4,7 @@
 
 <div class="space-container">
   <mat-toolbar class="primary-color font-size-1">
-    <span i18n>{{headingStart ? headingStart + ' ' : ''}}Course Progress</span>
+    <span i18n>{{'My' + ' '}}Course Progress</span>
```
lmmrssa: it is not required to put {{}} if you are just putting static text

PR #4163, opened 2019-06-28
TeamsView: Add resource

paulbert:
Allows a resource to be linked to a team from the library, or to create a new resource with the option (default) to make it only viewable by the team.

TODOs:

- A private team resource should be editable
- A private team resource should be deleted when removed (with a prompt notifying the user that the resource will be deleted)
- Private resources should be synced (for syncing teams), and maybe the public ones should be as well?  The public ones could cause conflicts, however.  Maybe there could be a button that the nation clicks to flag a public resource as ok to sync?  Or we only sync ones created on the community?

### DIFF
--- a/src/app/courses/add-courses/courses-step.component.ts
+++ b/src/app/courses/add-courses/courses-step.component.ts
@@ -5,7 +5,7 @@ import { MatDialog, MatDialogRef } from '@angular/material';
 import { Subject } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
 import { CoursesService } from '../courses.service';
-import { CoursesAddResourcesComponent } from './courses-add-resources.component';
+import { DialogsAddResourcesComponent } from '../../shared/dialogs/dialogs-add-resources.component';
 
 @Component({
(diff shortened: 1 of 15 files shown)


paulbert:
Also I realize the title for private resources should check for a different uniqueness since we should allow the same name in public or other teams' private resources to exist.

paulbert:
@lmmrssa Please take a look.  The code climate issues are fine for the time being.

Of note: resources created on the community and added to a sync team will sync along with their tags.  Resources added to a community from the nation will be added to `send_items` so the community can opt to fetch later.

As a follow up issue we can handle the case of a resource which doesn't exist locally.

lmmrssa:
@paulbert when testing with resource sync there is error of not having userId that is seen on nation. May be due to some latest changes on other PR caused it.
SharedScreenshot-res-sync

lmmrssa:
> @paulbert when testing with resource sync there is error of not having userId that is seen on nation. May be due to some latest changes on other PR caused it.
> SharedScreenshot-res-sync

Based on your conversation, you can ignore this :)

PR #4198, opened 2019-07-03
Teams: Creator is "Team Leader"

paulbert:
Just starting the process of adding leadership to teams.  The creator of the team is the leader, with no special privileges at this stage.


src/app/teams/teams.service.ts
```diff
@@ -47,7 +47,7 @@ export class TeamsService {
         switchMap((response: any) => response !== undefined ?
           this.updateTeam(
             { limit: 12, status: 'active', createdDate: this.couchService.datePlaceholder, teamPlanetCode: configuration.code,
-              parentCode: configuration.parentCode, ...team, ...response }
+              parentCode: configuration.parentCode, ...team, ...response, createdBy: userId }
```
lmmrssa: we can change **createdBy** to **teamLeader** if we want to have leadership transfer

PR #4221, opened 2019-07-05
Upgrade to Angular 8

paulbert:
The package.json upgrades all of the repositories that need to be updated for Angular 8, and leaves the ones that do not require an upgrade alone.

Right now this is untested because the markdown text editor throws an error.  This should be fixed once https://github.com/Teradata/covalent-text-editor/pull/28 is merged into a release for that library.


paulbert:
Also Angular 8 requires Node 10, so we should upgrade to that first.

paulbert:
Noticed referenced the wrong PR above.  The change is merged for the markdown editor, so once it is in their release I'll update this.

paulbert:
In that last commit I added the **fibers** package to our dev environment.  According to some sources I've read, this should improve the application build time (all mentioned in the sass section)

https://blog.ninja-squad.com/2019/05/29/angular-cli-8.0/
https://www.mindinventory.com/blog/whats-new-angular-8/
https://www.peerbits.com/blog/whats-new-in-angular-8-0-features.html

paulbert:
With fibers the build errored, so I removed that package

PR #4227, opened 2019-07-06
[Fixes# 4170] Team message: should be show more/show less button for long message

coder8102:
[Fixes# 4170] Team message: should be show more/show less button for long message

### DIFF
--- a/src/app/news/news-list-item.component.ts
+++ b/src/app/news/news-list-item.component.ts
@@ -6,7 +6,7 @@ import { UserService } from '../shared/user.service';
   templateUrl: 'news-list-item.component.html',
   styleUrls: [ './news-list-item.scss' ]
 })
-export class NewsListItemComponent implements OnInit {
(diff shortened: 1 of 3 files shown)


src/app/news/news-list-item.component.ts
```diff
@@ -20,6 +20,37 @@ export class NewsListItemComponent {
     private userService: UserService
   ) {}
 
+  ngOnInit() {
+    this.setDisplayMessage(this.item);
+  }
+
+  setDisplayMessage(news) {
+    const lines: string[] = news.message.split('\n\n');
```
paulbert: The message may not have newline characters to denote each line.  A long paragraph could be hundreds of lines because the words wrap around without a newline character.

You should set a max height when the message is in showLess mode and there is overflow.  I'm not sure off the top of my head how to detect for overflow, but maybe you can find that with a little research.

Issue #4210, opened 2019-07-04
Teams: Deleted members still exist on team


sjson421:
Members deleted by the manager still exist on the teams they were in previously, making deletion of said team impossible.

**Steps to Reproduce**:

1. Create a team
2. Have a member join this team
3. Have the manager delete that member
4. The member will still be "on" the team although he or she has been deleted from the system, making deletion impossible since there will always be that deleted member on the team.

paulbert:
Good catch.  When deleting a member we should also delete the membership docs in the team database.

paulbert:
Specifically in the teams database you would delete docs that match: `{ docType: 'membership', userId: user._id }`

paulbert:
Also, use the `bulkDocs` method supplying the `_delete: true` property to delete.

PR #4228, opened 2019-07-07
4210 deleted user team

sjson421:
Issue #4210

Also deletes user from team when a manager deletes that user

### DIFF
--- a/src/app/users/users.component.ts
+++ b/src/app/users/users.component.ts
@@ -223,12 +223,12 @@ export class UsersComponent implements OnInit, OnDestroy, AfterViewInit {
   }
 
   deleteUserFromTeams(user) {
-    return this.couchService.findAll('teams', { selector: { userId: user._id }}).pipe(
+    return this.couchService.findAll('teams', { selector: { userId: user._id } }).pipe(
       switchMap(teams => {
         const docsWithUser = teams.map(doc => ({ ...doc, _deleted: true }));
-        return this.couchService.bulkDocs('teams', docsWithUser)
+        return this.couchService.bulkDocs('teams', docsWithUser);
       })
-    )
+    );
   }
 
   deleteUser(user) {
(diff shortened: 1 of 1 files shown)


src/app/users/users.component.ts
```diff
@@ -220,8 +222,21 @@ export class UsersComponent implements OnInit, OnDestroy, AfterViewInit {
     event.stopPropagation();
   }
 
+  deleteUserFromTeams(user) {
+    // //REPLACE THIS!
+    // this.couchService.findAll('teams').then(teamsDb => {
+    //   const docsWithUser = teamsDb.filter(doc => doc.userId === user._id)
+    //   this.couchService.bulkDocs('teams', docsWithUser.map(doc =>
+    //     this.couchService.delete('teams/' + doc._id + '?rev=' + doc._rev))
```
paulbert: `{ ...doc, _deleted: true }`

src/app/users/users.component.ts
```diff
@@ -220,8 +222,21 @@ export class UsersComponent implements OnInit, OnDestroy, AfterViewInit {
...
+    //   this.couchService.bulkDocs('teams', docsWithUser.map(doc =>
+    //     this.couchService.delete('teams/' + doc._id + '?rev=' + doc._rev))
+    //   );
+    // })
+  }
+
   deleteUser(user) {
     const userId = 'org.couchdb.user:' + user.name;
+
+    this.deleteUserFromTeams(user);
```
paulbert: This line should be run as another element of the array in `forkJoin` on line 243

src/app/users/users.component.ts
```diff
@@ -220,8 +222,21 @@ export class UsersComponent implements OnInit, OnDestroy, AfterViewInit {
     event.stopPropagation();
   }
 
+  deleteUserFromTeams(user) {
+    // //REPLACE THIS!
+    // this.couchService.findAll('teams').then(teamsDb => {
```
paulbert: 1. Return an open observable to be used in the deleteUser function
2. You can use a selector with findAll so you don't have to filter
3. Use a switchMap to run the bulkDocs for deleting
```
return this.couchService.findAll('teams', { selector: { userId: user._id }).pipe(switchMap(teams => {
  return this.couchService.bulkDocs...
})
```

sjson421:
@paulbert Thanks for the very detailed help

PR #4229, opened 2019-07-08
teams leader can delete, not leave (fixes #4211)

### LINKED ISSUE #4211: Teams: Leader can delete, not leave
I think before the team leader leaves a team, he or she should be forced to designate another team leader if there are still other members left in the team. Otherwise, the team is left without a team leader.

anishregmee:
thank you in advance

### DIFF
--- a/src/app/teams/teams.component.html
+++ b/src/app/teams/teams.component.html
@@ -51,7 +51,12 @@
         <mat-cell *matCellDef="let element">
           <ng-container [ngSwitch]="element.userStatus" *ngIf="user.isUserAdmin || user.roles.length">
             <ng-container *ngSwitchCase="'member'">
-              <button mat-raised-button color="primary" class="margin-lr-3" (click)="toggleMembership(element.doc, true, element.membershipDoc)"><mat-icon>remove_circle</mat-icon> Leave</button>
+              <span *ngIf="user._id === element.doc.createdBy; else leaveButton">
+                <button mat-raised-button color="primary" class="margin-lr-3" (click)="toggleDeleteMembership(element.doc)"><mat-icon>delete_forever</mat-icon> Delete</button>
+              </span>
+              <ng-template #leaveButton>
+                <button mat-raised-button color="primary" class="margin-lr-3" (click)="toggleMembership(element.doc, true, element.membershipDoc)"><mat-icon>remove_circle</mat-icon> Leave</button>
+              </ng-template>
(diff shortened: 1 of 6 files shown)


src/app/teams/teams.component.html
```diff
@@ -41,7 +41,12 @@ <h3>
         <mat-cell *matCellDef="let element">
           <ng-container [ngSwitch]="element.userStatus" *ngIf="user.isUserAdmin || user.roles.length">
             <ng-container *ngSwitchCase="'member'">
-              <button mat-raised-button color="primary" class="margin-lr-3" (click)="toggleMembership(element.doc, true, element.membershipDoc)"><mat-icon>remove_circle</mat-icon> Leave</button>
+              <span *ngIf="user._id === element.doc.createdBy; else leaveButton">
+                <button mat-raised-button color="primary" class="margin-lr-3" (click)="toggleDeleteMembership(element.doc)"><mat-icon>delete_forever</mat-icon> Delete</button>
```
lmmrssa: `toggleDeleteMembership` name should be changed as it deletes team not toggledeletion

anishregmee: i have changed the name now.. thanx in advance for your review

src/app/teams/teams.service.ts
```diff
@@ -98,6 +98,13 @@ export class TeamsService {
     );
   }
 
+  deleteTeam(team) {
+    return this.updateTeam({ ...team, status: 'archived' })
+      .pipe(
```
paulbert: This pipe/switchMap is not necessary.  `newTeam` already has all of the properties on `team`

src/app/teams/teams.component.ts
```diff
@@ -110,6 +110,15 @@ export class TeamsComponent implements OnInit, AfterViewInit {
     });
   }
 
+  deleteTeam(team) {
+    this.teamsService.deleteTeam(team).subscribe((newTeam: any) => {
+      this.planetMessageService.showMessage('You have deleted a team.');
+      if (newTeam.status === 'archived') {
```
paulbert: No need for this conditional because the team will always be archived.

src/app/teams/teams.component.ts
```diff
@@ -110,6 +110,15 @@ export class TeamsComponent implements OnInit, AfterViewInit {
...
+        this.teams.data = this.teams.data.filter((t: any) => t.doc._id !== newTeam._id);
```
paulbert: Please refactor to create a new method for this line which is used here and in `toggleMembership`.  Let's call the new method `removeTeamFromTable`

PR #4260, opened 2019-07-09
Conditionally output reboot message (Fix #4081)

### LINKED ISSUE #4081: For myplanet upgrade do not show reboot message.
A "Success!" dialog box appears even when no upgrade has been made, prompting a reload. This could be confusing for first-time users.

Capture

samuelchen1213:
Only render reboot message if the mode is equal to planet

### DIFF
--- a/src/app/upgrade/upgrade.component.html
+++ b/src/app/upgrade/upgrade.component.html
@@ -22,7 +22,7 @@
       </mat-card-header>
       <mat-card-content>
         <p i18n>
-          The upgrade of the system has been done successfully.<br>
+          The upgrade of the system has been done successfully.
         </p>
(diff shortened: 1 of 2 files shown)


src/app/upgrade/upgrade.component.html
```diff
@@ -22,9 +22,9 @@
       </mat-card-header>
       <mat-card-content>
         <p i18n>
-            The upgrade of the system has been done successfully.<br>
-            Please restart your system to apply changes
+          The upgrade of the system has been done successfully.<br>
```
paulbert: Could you remove the `<br>` tag?  Using `<p>` tags is preferred, anyway, so this is a redundant extra space.

PR #4267, opened 2019-07-10
4253_issue_fix

AlekseiKorolev:
Screenshot_2019-07-10 Planet Learning


AlekseiKorolev:
Done

paulbert:
@AlekseiKorolev Thanks.  I forgot to reiterate this also from the issue:

> The text should read **Local** or **Connected with Nation**

As in when `teamType === local` the text should be **Local** and when it is `sync` it should be **Connected with Nation**

PR #4300, opened 2019-07-12
Create tabs in team view (fixes #4233)

### LINKED ISSUE #4233: Teams: UI change suggestion
I think courses should be split into a separate column so that you don't have to scroll so quickly when additional courses are added.

image

**Suggested Solution**:

Create a new column for courses as the second column, with the third column holding member info.

diabagatekelly:
I moved the Requested Member template with the Joined member together so they could be under one tab. Let me know if that’s ok?

### DIFF
--- a/src/app/shared/material.module.ts
+++ b/src/app/shared/material.module.ts
@@ -28,6 +28,7 @@ import {
   MatStepperModule,
   MatProgressSpinnerModule,
   MatTableModule,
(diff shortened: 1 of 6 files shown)


paulbert:
Oh also I made a small change to how the import happens for the new material module so you will have to run `git pull` on this branch before making your new changes

Issue #4298, opened 2019-07-11
Show a messages when no items show up


Liew211:
![image](https://user-images.githubusercontent.com/43839884/61070263-3fcc5880-a3dc-11e9-88d5-891bab301ddc.png)
When there are 0 items in the list, there could be a message saying so, this way the user will know that there isn't a problem with the system

PR #4305, opened 2019-07-12
#4298 Alert added for teams and meetups

sjson421:
#4298 

An alert pops up instead of the dialog box if there are no items to be shown in the dialog box. So far, I've seen the only cases of this being able to happen in teams and meetups. Please tell me if there are more.


sjson421:
@paulbert Since I've been seeing so many of these, how can I track what's causing the errors on Travis? The only thing I can tell is that there's an error for docker, but I can't pinpoint what's going wrong.

paulbert:
To elaborate a little on the review of @lmmrssa, this solution isn't very DRY and requires an extra conditional every time we use `DialogsListComponent`.  We should instead make the change within the component so when we supply an empty list to the dialog it just shows some text (let's do "There are no items to select") instead of the empty table.

PR #4312, opened 2019-07-12
Fix rpi planet travis build

paulbert:


paulbert:
Not necessary anymore because container fixed accepted platforms.  Closing.

PR #4320, opened 2019-07-15
Members: Show learner chip for every active member(fixes #4288)

### LINKED ISSUE #4288: Members: Show learner chip for every active member
Since every member is a learner, we should show the learner chip for every active member

Change should be on the template `users.component.html`

perryc85:
Alt text


perryc85:
Got it. I will start working on those changes. Thank you for the feedback.

perryc85:
@paulbert I made the requested changes to the code. 

image

PR #4355, opened 2019-07-18
count changed to totalSynced (fixes #4328)

### LINKED ISSUE #4328: myPlanet Reports: Clarify "Count" label
The label for the last column is simply "Count" which does not describe what the numbers are actually a count of.  Please change the label to be more descriptive.

anishregmee:
thank you for your review.


paulbert:
@anishregmee Thank you for the update.  This looks good from your end.

@lmmrssa This column does not appear to be the "Number of Times Synced" because in the data there are multiple documents for the same device with a different `time` property but the same `last_synced` property.  Is this page views or logins?  I would like to make sure this column title is not misleading.

PR #4404, opened 2019-07-23
4360_issue_fix

AlekseiKorolev:
Untitled
If I delete ```required``` key in ```addTeamDialogField``` object, the asterisk will disappear, but the field will become required. If I remove attribute ```required``` from component - then works as it should.

### DIFF
--- a/src/app/shared/dialogs/dialogs-form.component.html
+++ b/src/app/shared/dialogs/dialogs-form.component.html
@@ -51,7 +51,7 @@
 
       <!-- Textarea -->
       <mat-form-field *ngIf="field.type === 'markdown'" class="full-width mat-form-field-type-no-underline">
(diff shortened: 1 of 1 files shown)


AlekseiKorolev:
Ok. I have used now ```hideRequiredMarker```

AlekseiKorolev:
@paulbert It did't work. If I switch to a false ```requiered``` key from the ```addTeamDialogFields``` object, field become not requiered but the asterisk will not disappear. The fact that if the ```required``` attribute exists in the component, does not matter the value is true or false, the asterisk appears anyway

paulbert:
@AlekseiKorolev Hmm, I looked again and realize misread the object in `teams.service.ts`.  Something else is wrong if setting that to false does not remove the `*`.  I'll investigate.

AlekseiKorolev:
Ok. I'll try to find more clearly solution

PR #4412, opened 2019-07-24
reset button on collection dialog (fixes #4402)

### LINKED ISSUE #4402: Reset on Collection Dialog
As we have collection and search separated, people will not understand clear also reset collection. So it could be good to put **Reset** inside Collection dialog which clean Collections only.

anishregmee:


anishregmee:
Sorry.. there is still a issue here so please dont review it for now. I will recorrect and push it soon. Thank you

PR #4434, opened 2019-07-26
resources opens in new tab in teamsView (fixes #4429)

### LINKED ISSUE #4429: TeamsView: Resources & Courses should open in new browser tab
clicking of resource within Team loads the resource in same page and after that there is no way to return to Team using back button.
same for course

anishregmee:
thank you for your review.


anishregmee:
@paulbert yes thank you paul, i understood this issue. i think it would be better to create new issue for that to work.. please suggest me further.

Issue #4448, opened 2019-07-30
Teams: Change the "Remove" button to an "X" in the top corner of the card


paulbert:
We have a request to change the remove button in the cards for teams (members & resources) to be more like the dashboard: an "X" in the upper right corner.

Current TeamsView members:
![image](https://user-images.githubusercontent.com/9203229/62097293-56840380-b254-11e9-86f4-0d28cae08d22.png)

Dashboard examples:
![image](https://user-images.githubusercontent.com/9203229/62097310-64d21f80-b254-11e9-823c-06d0d9ac8b63.png)

PR #4456, opened 2019-07-30
[#4448] Change remove button to icon on team

singharpita:
@paulbert made changes on Team Member and Resources. Shall we make changes on Requested Members as well?


paulbert:
@singharpita Thanks, I'll take a look at the changes. For now let's leave the requests as is. I think it's clearer to have the full "Accept"/"Reject" words there.

PR #4459, opened 2019-07-31
Vagrant: Fix member creation error in development (fixes #2470)

### LINKED ISSUE #2470: Cannot create the new user : user becoming member
screenshot 2018-11-07 12 06 25
screenshot 2018-11-07 12 06 10

paulbert:
Updates the Vagrantfile and setup script to

1. Only run the `docker start` in the "always" script if it is not already running
2. In addition run the CouchDB setup script after the container starts with proxy _admin authentication

@dogi I think we can add the new configuration setting to our template instead of here, correct?

In addition in case this problem comes up again added a small script `v-couchdb-setup.sh` to simplify the fix and referenced this in the **Troubleshooting** section of the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,14 @@ sudo npm install --unsafe-perm
 
 The trailing `/*` will remove all files & sub-directories of node_modules.  You won't be able to remove node_modules because of the link between the vagrant VM and your host.
 
+### Cannot create new members in development environment
+
+Sometimes our custom setup for the `_users` database is overwritten by the default.  If that is the case, run the following command to run our database setup script again:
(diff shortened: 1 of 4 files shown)


paulbert:
@lmmrssa Dogi took a look, but if you are able to test also that would be great.

To test: destroy then up dev.  Setup configuration & try to create a new member.  Then halt & up.  Try to create a new member again.

I noticed creating a new member after the initial `vagrant up` didn't work once out of the 5-10 times I tried it.  As long as creating a new member after the halt/up sequence works I think this is good to merge.

PR #4460, opened 2019-08-01
Make entire team table row clickable (fixes #4427)

### LINKED ISSUE #4427: Teams: Click anywhere on row to open team
Currently to open a team you have to click directly on the team name.

It would make things a little smoother if the members could click anywhere on the row to open the team page.

diabagatekelly:


diabagatekelly:
> clicking button also opens team view which should not happen

By button, do you mean the team name?

diabagatekelly:
> clicking button also opens team view which should not happen

Never mind I see what you mean.

src/app/teams/teams.component.ts
```diff
@@ -106,7 +109,8 @@ export class TeamsComponent implements OnInit, AfterViewInit {
     });
   }
 
-  addTeam(team?) {
+  addTeam(team?, e) {
```
paulbert: This is causing a build error because required function parameters must come before optional ones.

lmmrssa:
There is error when trying to add team due to stoppropagation function

PR #4461, opened 2019-08-01
4420_issue_fix

AlekseiKorolev:
I did this through the ```ngOnInit``` function, because we need to check the condition and use the same part of code when the page init, when we return from the add meetings dialog and when we toggle.

### DIFF
--- a/src/app/meetups/meetups.component.html
+++ b/src/app/meetups/meetups.component.html
@@ -18,6 +18,7 @@
         <mat-icon>add</mat-icon>
       </button>
       <span class="toolbar-fill"></span>
+      <mat-slide-toggle (click)="toggleMeetups()" checked="checked">Show outdated meetups</mat-slide-toggle>
       <button mat-button [disabled]="selectedNotJoined === 0" (click)="meetupsToggle(selection.selected, 'add')">
         <mat-icon aria-hidden="true" class="margin-lr-3">library_add</mat-icon><span i18n>Join Selected</span>
         <span *ngIf="selectedNotJoined > 0"> ({{selectedNotJoined}})</span>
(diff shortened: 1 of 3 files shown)


src/app/meetups/meetups.component.ts
```diff
@@ -48,6 +48,7 @@ export class MeetupsComponent implements OnInit, AfterViewInit, OnDestroy {
   selectedJoined = 0;
   isAuthorized = false;
   dateNow: any;
+  showOutdated = true;
```
paulbert: Let's make the default `false`

src/app/meetups/meetups.component.html
```diff
@@ -18,6 +18,7 @@
         <mat-icon>add</mat-icon>
       </button>
       <span class="toolbar-fill"></span>
+      <mat-slide-toggle (click)="toggleMeetups()" checked="checked">Show outdated meetups</mat-slide-toggle>
```
paulbert: Instead of here let's have the button before the search in the upper toolbar

src/app/meetups/meetups.component.ts
```diff
@@ -64,12 +65,24 @@ export class MeetupsComponent implements OnInit, AfterViewInit, OnDestroy {
     this.couchService.currentTime().subscribe((date) => this.dateNow = date);
   }
 
+  toggleMeetups() {
+    this.showOutdated = !this.showOutdated;
+    this.ngOnInit();
```
paulbert: We shouldn't rerun `ngOnInit` because it creates more subscriptions to the `meetupUpdated$` observable.

Instead you should add a function to the `composeFilterFunctions` with one argument which will be the meetup in the array to test.  You can make sure the filter runs by setting `this.meetups.filter = this.meetups.filter || ' '`.

AlekseiKorolev:
I apologize for working with this issue for so long. The reason is that I really could not understand why the search result does not appear. The logic is that the search result appears if all functions in ```composeFilterFunctions``` return true. This means that if we search for something by a word, the result appears only if the title and description contain this word.

AlekseiKorolev:
The last commit is incorrect, I'm preparing another

Issue #4468, opened 2019-08-01
Resources/Courses: Maintain collection selection after returning from view


paulbert:
### Current Behavior

After selecting a collection then clicking to view a resource or course, the back button will return to a full list of the resources courses

### Desired Behavior

In the last step, the back button should go back to the same collection selection that was initially selected in the first step.

There are many ways to do this, but we implement something like this with the search in `UsersComponent` (this could be improved upon).  Some combination of a service or additional logic in `TagsService` & the [optional route parameters](https://angular.io/guide/router#optional-route-parameters) is a good place to start.

Video showing the current behavior:
https://www.loom.com/share/2c12927685f2463d978f25933085114f

PR #4542, opened 2019-08-11
#4468 maintain collection

sjson421:
#4468 

Created on a new branch because angular 8 changes and tag/subtag changes required a new start.


sjson421:
@paulbert Filtering by collections from the button at the top works correctly, but this particular branch has some really funky behavior with clicking the blue tag filter buttons not showing up that makes it difficult to test new methods on top of them. I suggest that once that issue is fixed, a new issue can be created, which I can take back up.

src/app/shared/forms/planet-tag-input-dialog.component.ts
```diff
@@ -52,6 +52,7 @@ export class PlanetTagInputDialogComponent {
   get okClickValue() {
     return { wasOkClicked: true, indeterminate: this.indeterminate ? mapToArray(this.indeterminate, true) : [] };
   }
+  selectedTags = [];
```
paulbert: This variable is never used, and can be removed

src/app/shared/forms/planet-tag-input-dialog.component.ts
```diff
@@ -98,7 +99,7 @@ export class PlanetTagInputDialogComponent {
     this.selected.clear();
     this.data.reset(this._selectMany);
   }
-
+
```
paulbert: We shouldn't have a line with only spaces.  If you're using VS Code you can set the editor to remove these extra spaces automatically.

src/app/shared/forms/planet-tag-input.component.ts
```diff
@@ -211,15 +213,19 @@ export class PlanetTagInputComponent implements ControlValueAccessor, OnInit, On
     });
   }
 
-  dialogTagUpdate(tag, isSelected, tagOne = false) {
+  dialogTagUpdate(tag, isSelected, tagOne = false, selectedTags) {
```
paulbert: `selectedTags` is not used as a function parameter, so you should declare it as a variable with `let` instead of as a function parameter.

paulbert: Reading this a little more I see `selectedTags` is only modified, it doesn't actually do anything.  Did you mean to declare `selectedTags` as a property of this component instead of **planetTagInputDialogComponent**?

src/app/shared/forms/tags.service.ts
```diff
@@ -133,4 +136,18 @@ export class TagsService {
     return tag.attachedTo === undefined || tag.attachedTo.length === 0;
   }
 
+  /**
+   * Reroutes to new URL on filtering so that the back button of a particular result go to the previous filtered results
+   */
+  filterReroute(tags) {
```
paulbert: This method is never used, and can be removed.

src/app/resources/resources.component.ts
```diff
@@ -113,6 +115,11 @@ export class ResourcesComponent implements OnInit, AfterViewInit, OnDestroy {
       this.emptyData = !this.resources.data.length;
       this.resources.paginator = this.paginator;
       this.dialogsLoadingService.stop();
+
+      const urlSelectedTags = this.route.snapshot.paramMap.get('tags').split('`');
```
paulbert: This will cause an error when the `tags` parameter doesn't exist (a case that should happen regularly)

src/app/shared/forms/planet-tag-input.component.ts
paulbert: This is also causing the funky behavior.  It is never defined, so `selectedTags.push` throws an error

sjson421: @paulbert Looks like I accidentally deleted some of my code while merging. I'll keep you posted. The funky behavior is with the blue buttons, not the issue at hand.

src/app/resources/resources.component.ts
sjson421: It actually does not.

sjson421:
@paulbert To be specific, the funky behavior is with the blue filter buttons appearing sporadically

src/app/resources/resources.component.ts
paulbert: When loading the resources page this line throws an error.

image

sjson421:
@paulbert Some refactoring was done in order to accommodate for the changes in the Course collection. The collection filter is now correctly stored in the URL but only works for the updated collection dialog box of Courses and not for Resources. This was under the assumption that Resources will eventually use the same dialog box. Filtering also works with the blue buttons.

In order to add the same functionality to Resources once its own collection dialog box has been updated, just copy over the changes in courses.component.ts (in case this change isn't completed by the time I finish my internship). TagsService has already been imported into the resources component.

PR #4576, opened 2019-08-14
TeamsView: Refactor DialogsPromptComponent (fixes #4396)

### LINKED ISSUE #4396: TeamsView: Refactor DialogsPromptComponent for code duplication
`DialogsPromptComponent` is called in three different ways, currently with some duplication across the different methods.  We should be able to refactor this to reduce the duplication in **TeamsViewComponent**.

paulbert:
#4396


paulbert:
@singharpita @lmmrssa Please review.  Added in a fix for a regression in #4621 that slipped through so this should be merged if possible.

Issue #4540, opened 2019-08-09
Teams: Add option for "Open Team"


paulbert:
We have a request to add an option for a team to be "open".  The difference between this and the current teams is that when a person requests to join these teams they should automatically become a member without the need for an approval from a team member.

This should be the default when creating a team, but all current teams should remain "Closed"

PR #4577, opened 2019-08-14
[#4540] Open team

singharpita:


lmmrssa:
There is now Public team so we can update this to reflect public team

Issue #4509, opened 2019-08-06
ResourcesView: Bounce to library if resource does not exist


singharpita:
Steps to replicate:
- Create a team
- create new resources with private resource checkbox 
- open that resource and copy the url
- - delete that team.
- now press the browser back button
- It will redirect to the deleted private resources and also allowed to submit it
- paste the above url and hit enter
- deleted private resource shown

can deleted private resource be still accessible through link or browser back button ?

paulbert:
Editing title since we should just route to the library from ResourcesView link to a resource that does not exist on the database

sjson421:
@paulbert I get the feeling that your comment is something completely different from the issue that arpita's presented. I will work on your issue.

PR #4607, opened 2019-08-19
#4509 Redirects to resources if resource nonexistent

sjson421:
#4509


sjson421:
I didn't realize that Paul fixed this issue about an hour before I submitted my fix (they were pretty much the same, with paul's having the additional alert box) and deleted that old branch to start over from a clean branch.

Issue #4599, opened 2019-08-15
Tasks: Sync with nation when part of connected team


paulbert:
Right now tasks do not sync, so there is no way to send tasks back and forth between nation & community.

Need to add the tasks database to the list of databases to push & pull in the `SyncDirective`

PR #4620, opened 2019-08-20
[#4599] Team task sync

singharpita:


singharpita:
@paulbert can you check if the approach I did is correct? Or do you have any other better suggestion?

PR #4626, opened 2019-08-21
Health: Add database with at rest encryption

paulbert:
Uses views to encrypt documents at rest.  For example:

Adding a doc
```
curl -X PUT http://localhost:2200/health/_design/health/_update/encrypt -d '{"_id":"1","key":[1,2,3,4,5,6,7,8,9,0,1,2,3,4,5,6,7,8,9,0,1,2,3,4, 5,6,7,8,9,0,1,2],"cat":"zuzu"}' -H "Content-Type:application/json" 
```

Getting a doc
```
curl -X POST http://localhost:2200/health/_design/health/_show/decrypt/1 -d '{"key":[1,2,3,4,5,6,7,8,9,0,1,2,3,4,5,6,7,8,9,0,1,2,3,4,5,6,7,8,9 ,0,1,2]}' -H "Content-Type:application/json" -H "Accept:application/json"
```

The key should be an array with 16 or 32 integers 0 to 255.  Not sure the best method to generate these for the different users right now.

I think at this point we can add this and work on incorporating it into the application, but prioritize reviewing other PRs.


paulbert:
Oh, after creating a doc you can check to see the encrypted data here:
```
http://localhost:2200/health/1
```

dogi:
@paulbert :D

PR #4627, opened 2019-08-21
treehouses buster 0.10.1 & couchdb 2.3.1

dogi:


dogi:
@paulbert wonder if we should do already https://github.com/open-learning-exchange/planet/blob/treehouses-buster-0.10.1/docker/planet.yml#L5

PR #4680, opened 2019-08-27
Add delete collection button in courses (fixes #4591)

### LINKED ISSUE #4591: Courses: Can't delete course collection
Resource collections can be deleted but the ones on course no longer have an option to delete them.

Liew211:


src/app/shared/forms/planet-tag-input-dialog.component.html
```diff
@@ -58,6 +58,7 @@
         <planet-tag-input-toggle-icon *ngIf="tag.subTags.length > 0" (click)="toggleSubcollection($event,tag._id)" [isOpen]="subcollectionIsOpen.get(tag._id)"></planet-tag-input-toggle-icon>
         <span class="toolbar-fill"></span>
         <button mat-stroked-button *ngIf="isUserAdmin" (click)="editTagClick($event,tag)">Edit</button>
+        <button mat-stroked-button *ngIf="isUserAdmin" [disabled]="tag.subTags.length > 0" (click)="deleteTag($event,tag)">Delete</button>
```
paulbert: The `span` with the tooltip should be included so when the button is disabled there is an indication to the user why it is disabled.

Liew211:
@paulbert Misunderstood your comment at first, but I think it should be good now?

PR #4682, opened 2019-08-27
Improve travis build time

paulbert:


paulbert:
@lmmrssa @dogi You should test `planet-tags:0.8.43-test-build-time-eae5acf9` to see if the app works & all languages are present.

Next commits will make changes so only English builds when the build is triggered by anything other than a tag.

paulbert:
Tag `0.8.43-test-build-time-3609fbf8` should be tested for the English only version

paulbert:
At this point the test would be to see if the planet container has a working app for each of the languages it built.  I didn't add the condition to only build all languages with a tag until one of the last commits, so we can test multi language with one of the earlier builds.  Here's what you can use for install.yml in the docker-compose for each build:

install.yml to test for all languages
```
services:
  db-init:
    image: treehouses/planet-tags:db-init-0.8.43-test-build-time-eae5acf9
  planet:
    image: treehouses/planet-tags:0.8.43-test-build-time-eae5acf9
version: "2"
```
install.yml to test for english only build
```
services:
  db-init:
    image: treehouses/planet-tags:db-init-0.8.43-test-build-time-3609fbf8
  planet:
    image: treehouses/planet-tags:0.8.43-test-build-time-3609fbf8
version: "2"
```

PR #4716, opened 2019-08-29
Feedback: Updating title error (fixes #4648)

### LINKED ISSUE #4648: Feedback: updating feedback title gives error
Screenshot from 2019-08-22 21-37-13

perryc85:


src/app/feedback/feedback-view.component.ts
```diff
@@ -111,15 +111,16 @@ export class FeedbackViewComponent implements OnInit, OnDestroy {
   }
 
   editTitle(mode) {
-    this.editTitleMode = mode;
+    this.couchService.put(this.dbName + '/' + this.feedback._rev, this.feedback).subscribe(
```
paulbert: The name of this function could be a little clearer, but this simply switches from just showing the title text or showing an input field to edit the title text.

This function should remain as is.

src/app/feedback/feedback-view.component.ts
```diff
@@ -111,15 +111,16 @@ export class FeedbackViewComponent implements OnInit, OnDestroy {
...
+      this.editTitleMode = mode);
   }
 
   setTitle() {
     this.couchService.put(this.dbName + '/' + this.feedback._id, this.feedback).subscribe(
       () => {
         this.editTitleMode = false;
```
paulbert: The change should happen here.  I know the function does not list a parameter, but there is one (we can call it `res`) which will include a `rev` property that should overwrite `this.feedback._rev`

perryc85: Awesome. Thank you for the feedback.

src/app/feedback/feedback-view.component.ts
```diff
@@ -22,7 +22,7 @@ export class FeedbackViewComponent implements OnInit, OnDestroy {
   user: any = {};
   newMessage = '';
   isActive = true;
-  editTitleMode = false;
+  toggleTitleMode = false;
```
paulbert: Sorry to ask for a reversion here, but I think I just missed the clear way to name these.  This variable should remain `editTitleMode` and the method should change to `toggleEditTitleMode`

paulbert: Actually with a little more thought I realize your suggestion `changeEditTitleMode` would be better since we supply a value to the function.

src/app/feedback/feedback-view.component.ts
```diff
@@ -110,16 +110,19 @@ export class FeedbackViewComponent implements OnInit, OnDestroy {
     );
   }
 
-  editTitle(mode) {
-    this.editTitleMode = mode;
+  toggleTitle(mode) {
```
paulbert: See other comment, change to `changeEditTitleMode`

src/app/feedback/feedback-view.component.ts
```diff
@@ -110,16 +110,19 @@ export class FeedbackViewComponent implements OnInit, OnDestroy {
...
+    this.toggleTitleMode = mode;
   }
 
   setTitle() {
     this.couchService.put(this.dbName + '/' + this.feedback._id, this.feedback).subscribe(
-      () => {
-        this.editTitleMode = false;
+      (res) => {
+        this.toggleTitleMode = false;
+        this.couchService.put(this.dbName + '/' + res._rev, this.feedback).subscribe(
```
paulbert: At this point we're done changing values on the database, but we need to modify `this.feedback` so it has the updated value for `_rev`.

Confusingly CouchDB returns that as `res.rev` without the underscore.

src/app/feedback/feedback-view.component.ts
```diff
@@ -110,16 +110,19 @@ export class FeedbackViewComponent implements OnInit, OnDestroy {
...
+    this.changeEditTitleMode = mode;
   }
 
   setTitle() {
     this.couchService.put(this.dbName + '/' + this.feedback._id, this.feedback).subscribe(
-      () => {
-        this.editTitleMode = false;
+      (res) => {
+        this.changeEditTitleMode = false;
+        this.couchService.put(this.dbName + '/' + res.rev, this.feedback).subscribe(
```
paulbert: There shouldn't be another request to CouchDB here.  You need to update `this.feedback._rev` to match `res.rev`

perryc85: Okay. I will fix this too. Thank you for the feedback here as well.

perryc85: @paulbert - I made the latest changes you requested.

PR #4719, opened 2019-08-29
Add matbadge on tasks

### LINKED ISSUE #4640: Team: Add task count on tab
There should be a count of tasks on the tab, just like there are on the right side of the Members and Join Request tabs.

Capture

Liew211:
Fixes #4640 

Syntax still needs fix

### DIFF
--- a/src/app/teams/teams-view.component.html
+++ b/src/app/teams/teams-view.component.html
@@ -100,7 +100,7 @@
         </mat-tab>
         <mat-tab #taskTab>
           <ng-template mat-tab-label>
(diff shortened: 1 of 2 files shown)


paulbert:
You can combine the `ng-container` and `span` into one element to reduce the nesting.  Just make sure to put the `i18n` attribute on the span.

Overall this is a good start, but the number should not be the full number of tasks.  It should instead be based on the currently logged in member:
- **Leader**: The total number of unfinished tasks for the team
- **Other members**: The total number of unfinished tasks assigned to the logged in member

You can make another property that you set after the tasks are fetched.

Issue #4706, opened 2019-08-29
Last login missing


Liew211:
Time of last login is blank

Issue #4707, opened 2019-08-29
Resource views and ratings disappear on refresh in manager page


Liew211:
Resources remain in the report detail, but on the manager page, the resources views in the past 30 days and total ratings in the past 30 days become 0 after the page is refreshed.

PR #4723, opened 2019-09-02
[#4707] Resource rating on manager dashboard

singharpita:


singharpita:
This also fixes #4706

PR #4750, opened 2019-09-04
added positive number validator(Fixed #4692)

### LINKED ISSUE #4692: Enterprise budget amount validation
It should not allow to have negative amount.

wanglihe2010:
fixed #4692


wanglihe2010:
I am wondering why the input has both CustomValidator.Required and required:true. I thought they are the same.

PR #4769, opened 2019-09-06
Tasks: Notify a user when they are assigned a task (fixes #4634)

### LINKED ISSUE #4634: notification to assignee user when task is assigned
it would be better if notification goes to that user when a task is assigned to that specific user.

perryc85:


src/app/tasks/tasks.component.ts
```diff
@@ -55,6 +59,31 @@ export class TasksComponent implements OnInit {
     this.tasksService.addTask({ ...task, assignee }).subscribe((res) => {
       this.tasksService.getTasks();
     });
+    this.sendNotifications();
+  }
+
+  sendNotifications(assignee: any = '') {
+    const link = '/teams/view';
```
paulbert: This link won't show anything.  It needs the team id at the end `/teams/view/${teamID}`

src/app/tasks/tasks.component.ts
```diff
@@ -55,6 +59,31 @@ export class TasksComponent implements OnInit {
...
+      'priority': 1,
+      'status': 'unread',
+      'time': this.couchService.datePlaceholder,
+      userPlanetCode
+    });
+
+    if (assignee !== '' && assignee.userDoc) {
+      return this.couchService.findAll('notifications', findDocuments({ link, type: 'newTask', status: 'unread' })).pipe(
+        switchMap(() => {
+          const newNotifications = this.tasks.filter(assignee)
```
paulbert: You should filter on the result of the notifications, not the current list of tasks.

Also, `Array.filter()` takes a function as its first parameter.  The function returns `true` or `false` for each element of the array, and the result will be all elements in the array which return `true`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

perryc85: Thank you for the feedback. Will make these changes.

perryc85: Hey Paul, I pushed another revision to see if I am on a better track to solving this issue. Again, thank you for the feedback and sorry for the time it has taken.

PR #4779, opened 2019-09-08
4763 tasks completion time

anishregmee:
thank you for your review


src/app/tasks/tasks.component.html
```diff
@@ -15,6 +15,7 @@
     <mat-checkbox [checked]="task.completed"></mat-checkbox>
     <p matLine>{{task.title}}</p>
     <p matLine><ng-container i18n>Deadline:</ng-container> {{task.deadline | date}}</p>
+    <p matLine *ngIf="task.completed"><ng-container i18n>Completed Time: </ng-container> {{task.completedTime | date:'MMM d, y, h:mm a'}}</p>
```
paulbert: Let's remove this, but also for future reference please use generic date pipes (the default, or `longDate`, `medium`, etc) so the exact format is set by the locale.

src/app/tasks/tasks.component.ts
```diff
@@ -56,7 +61,7 @@ export class TasksComponent implements OnInit {
   }
 
   toggleTaskComplete(task) {
-    this.tasksService.addTask({ ...task, completed: !task.completed }).subscribe((res) => {
+    this.tasksService.addTask({ ...task, completed: !task.completed, completedTime: this.now }).subscribe((res) => {
```
paulbert: This is not how we add the current time to a document.  We use `couchService.datePlaceholder` which then is replaced when `couchService.updateDocument` is run.

In this case I think it would be best to append `completedTime` in **TasksService** if the task is completed.

anishregmee: thank you for the review @paulbert

Issue #4735, opened 2019-09-03
Tasks: Avatar not most up to date


paulbert:
**To Replicate**
- Create a team & a task on that team
- Assign a task to a team member
- Change the profile image of the assigned member
- In tasks the member image is not updated (this part is shown in the below video)

**Desired Behavior**
In the TasksComponent we should show the most up to date avatar for the assigned member of the task.

https://www.loom.com/share/fc59e266e7394bab8f470dd01424e138

RheuX:
![image](https://github.com/user-attachments/assets/fb1c8fcc-ffee-4600-b0dd-f68061576518)

![image](https://github.com/user-attachments/assets/b2ae9a12-ac1c-45a9-a444-f348f6714102)

PR #4815, opened 2019-09-12
4735 tasks vatar

anishregmee:
#4735


jessewashburn:
If I understand this correctly, it wants to add avatars to the assignees for tasks. We have that. 
image

PR #4816, opened 2019-09-12
4613 issue

AlekseiKorolev:
I add a rule to check if the user is expecting a response to a request

### DIFF
--- a/src/app/teams/teams-view.component.ts
+++ b/src/app/teams/teams-view.component.ts
@@ -282,10 +282,20 @@ export class TeamsViewComponent implements OnInit, OnDestroy {
   }
 
   addMembers(selected: any[]) {
-    this.dialogsLoadingService.start();
     const newMembershipDocs = selected.map(
       user => this.teamsService.membershipProps(this.team, { userId: user._id, userPlanetCode: user.planetCode }, 'membership')
     );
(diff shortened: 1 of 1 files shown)


src/app/teams/teams-view.component.ts
```diff
@@ -277,10 +277,20 @@ export class TeamsViewComponent implements OnInit, OnDestroy {
   }
 
   addMembers(selected: any[]) {
-    this.dialogsLoadingService.start();
     const newMembershipDocs = selected.map(
       user => this.teamsService.membershipProps(this.team, { userId: user._id, userPlanetCode: user.planetCode }, 'membership')
     );
+    const checkActiveRequest = this.requests.some((req: any) => {
+      if (req.userId === newMembershipDocs[0].userId) {
```
paulbert: This only affects the first member selected to be added to the team.  All other request docs will remain and only the first member selected will be added.

src/app/teams/teams-view.component.ts
```diff
@@ -277,10 +277,20 @@ export class TeamsViewComponent implements OnInit, OnDestroy {
...
+        this.changeMembership('added', req);
```
paulbert: In general, you should avoid side effects in array functions.  This PR needs a lot of changes so no need to fix this exact problem, but wanted to note this for future reference.

PR #4844, opened 2019-09-17
4839_issue_fix

AlekseiKorolev:


src/app/meetups/add-meetups/meetups-add.component.ts
```diff
@@ -85,11 +85,7 @@ export class MeetupsAddComponent implements OnInit {
 
   createForm() {
     this.meetupForm = this.fb.group({
-      title: [
-        '',
-        CustomValidators.required,
```
paulbert: We should keep `CustomValidators.required` for the title

src/app/meetups/add-meetups/meetups-add.component.html
```diff
@@ -10,9 +10,6 @@
     <form class="form-spacing" [formGroup]="meetupForm" novalidate>
       <mat-form-field>
         <input matInput i18n-placeholder placeholder="Title" formControlName="title" required>
-        <mat-error>
```
paulbert: Because there will still be a validator, you should put the `mat-error` element back in.

PR #4858, opened 2019-09-17
added time to input date(Fixed 3276)

### LINKED ISSUE #3276: cannot create a meetup for current day
When creating a current meetup, we are unable to select the current day as the start day. a red error message states "Cannot be before current date". either change so that the current day can be selected or change the error message to say something like "Cannot begin on or before today" or something like that.

wanglihe2010:
Fix #3276. I did not format the couchservice current time output and found a better way to do the validation


wanglihe2010:
@paulbert As for the time zone inaccuracy, do you mean `ac.value` is the local time and `couchService.currentTime()` is the GMT time?

PR #4896, opened 2019-09-24
4894_fix

AlekseiKorolev:


src/app/meetups/add-meetups/meetups-add.component.ts
```diff
@@ -103,7 +103,7 @@ export class MeetupsAddComponent implements OnInit {
       meetupLocation: '',
       createdBy: this.userService.get().name,
       createdDate: this.couchService.datePlaceholder,
-      recurringNumber: [ 10, Validators.min(2) ]
+      recurringNumber: [ 10, Validators.min(2) && CustomValidators.integerValidator ]
```
paulbert: This is the incorrect syntax.  To have two synchronous validators, you combine them into one array.  With this the minimum validator is no longer working.

See https://github.com/open-learning-exchange/planet/blob/master/src/app/users/users-update/users-update.component.ts#L93 for an example

paulbert:
Travis links to incorrect build above because a PR with the same title was created.  Correct Travis build is here: https://travis-ci.org/open-learning-exchange/planet/builds/589035017

PR #4897, opened 2019-09-25
4784_fix

AlekseiKorolev:
I understood this issue that it was necessary to remove the asterisk from each radio button and add to the title of each group of radio buttons without css rule edit


paulbert:
@AlekseiKorolev Sorry for the misunderstanding.  The issue is that the HTML or the CSS needs tweaking so the asterisk only shows up on the title/main heading based on the required attribute and does not show up for the label of each button.

The text should not have the asterisk manually added.

AlekseiKorolev:
This time I used HTML

paulbert:
To add to that, I think there are ways to do this by only modifying the HTML, only modifying the CSS, or modifying both.  The only requirement is that after the modifications the * for required fields applies generally to all `mat-radio-groups`

PR #4904, opened 2019-09-26
ExamsView: Fix grading endless spinner (fixes #4643)

### LINKED ISSUE #4643: Infinite spinner when grading tests
When text-answer test questions are submitted, and the admin goes to grade the answers, an infinite spinner comes up after Submit Answer is clicked.
image

paulbert:


paulbert:
@lmmrssa If you check the database the grade is recorded.  It looks like the grading form doesn't initialize correctly after that.  We can handle that in another issue.

PR #4905, opened 2019-09-26
CoursesProgress: Filter course progress by planet

paulbert:
On nation where there is course progress from many planets, add a dropdown for the progress view to filter by planet.


paulbert:
Closing in favor of #6078

Issue #4875, opened 2019-09-19
Reports: Select date range


paulbert:
It would be nice to add a feature to select a date range for the data export.

We could change the user flow to be as follows:

1. Click "Export"
2. Dialog pops up with a dropdown for selecting the report type and a couple of radio buttons for choosing "All Data" or select a specific date range.  If date range is selected, a couple of MatDatePickers can open with start date and end date.
3. Click OK to download the report.

PR #4906, opened 2019-09-26
[#4875] Add date filter on report

singharpita:


paulbert:
@singharpita This is useful, but not what the issue is referring to. The issue is for adding date selection for the data export to CSV.

I’ll take a closer look at this, but please open a different PR for date range for the CSV export.

PR #4924, opened 2019-09-27
show the tasks in the calendar(fixes 4877)

ScottHuangNYU:


ScottHuangNYU:
@paulbert I added tasks to the `events` array and show them on the calendar. But it still have some problem about the attribute names in task is not the same as meetups' name. So the when I click an d open the task event on the calendar, the information of task event will be missed because there are something wrong when `extendedProps` is task.

Issue #4871, opened 2019-09-19
Team: join request notification refers to join request tab


wanglihe2010:
Same as task assigned notification, we can also set join request notification refers to `join request` tab, instead of `plan` tab

PR #4940, opened 2019-10-01
change request notification link( Fixed 4871)

wanglihe2010:
#4871 
I added `activeTab` to manage required tab in one place, so I also changed code in task link. Please let me know if there is a better way to do it.

### DIFF
--- a/src/app/teams/teams-view.component.ts
+++ b/src/app/teams/teams-view.component.ts
@@ -54,7 +54,7 @@ export class TeamsViewComponent implements OnInit, AfterViewChecked, OnDestroy {
   leaderDialog: any;
   finances: any[];
   tabSelectedIndex = 0;
-  initTab = '';
+  initTab;
   taskCount = 0;
 
   constructor(
@@ -92,8 +92,10 @@ export class TeamsViewComponent implements OnInit, AfterViewChecked, OnDestroy {
   ngAfterViewChecked() {
     const activeTab: MatTab = this.getActiveTab(this.initTab);
     if (activeTab && activeTab.position !== 0) {
-      setTimeout(() => this.tabSelectedIndex = this.tabSelectedIndex + activeTab.position, 0);
-      this.initTab = '';
+      setTimeout(() => {
+        this.tabSelectedIndex = this.tabSelectedIndex + activeTab.position;
+        this.initTab = activeTab.position === 0 ? '' : this.initTab;
+      }, 0);
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/app/teams/teams.service.ts
```diff
@@ -208,11 +208,14 @@ export class TeamsService {
   }
 
   teamNotification(message, user, { team, url }) {
+    const link = url.split(';')[0];
     const userId = user.userId || user._id;
+    const linkParams = message.includes('requested') ? { activeTab: 'applicantTab' } : {};
```
paulbert: Instead of doing a string search on the `message`, we can pass in `type` from `sendNotifications` to this function and use that to test against.

src/app/teams/teams-view.component.ts
```diff
@@ -92,8 +92,10 @@ export class TeamsViewComponent implements OnInit, AfterViewChecked, OnDestroy {
   ngAfterViewChecked() {
     const activeTab: MatTab = this.getActiveTab(this.initTab);
     if (activeTab && activeTab.position !== 0) {
-      setTimeout(() => this.tabSelectedIndex = this.tabSelectedIndex + activeTab.position, 0);
-      this.initTab = '';
+      setTimeout(() => {
+        this.tabSelectedIndex = this.tabSelectedIndex + activeTab.position;
+        this.initTab = activeTab.position === 0 ? '' : this.initTab;
```
wanglihe2010: Hi, Paul, the `activeTab.position` will not never be 0 here since we have if condition above

paulbert: `tabSelectedIndex` modifies the `activeTab.position`.  For whatever reason the first time modifying this the `activeTab` is not selected, so adding that condition means it changes the `tabSelectedIndex` on the next change cycle also.

wanglihe2010: Got it, thanks, so it will check whether the tab successfully changed to the desire tab. What will happen if not successfully changed, will it trigger some functions? Why do you make `initTab` to be `''` instead of `undefined`?

paulbert: `ngAfterViewChecked` runs every change cycle.  The `setTimeout` is a way of making the function within wait until the cycle completes, then changing `tabSelectedIndex` will start another.  Once the tab has successfully been selected, we change it to `''` so it only selects the tab when the component loads.  We run `setStatus` on initialization and after any changes the user makes to the team.  The condition on line 180 checks for `initTab === undefined` so it only gets set to a different value when the component is initializing.

PR #4951, opened 2019-10-03
MeetupsAdd: Fix time validation (fixes #4928)

### LINKED ISSUE #4928: MeetupsAdd: End time validation regression

paulbert:
#4928

### DIFF
--- a/src/app/validators/custom-validators.ts
+++ b/src/app/validators/custom-validators.ts
@@ -49,18 +49,6 @@ export class CustomValidators {
     };
   }
 
-  static timeValidator(ac: AbstractControl): ValidationErrors {
-
-    if (!ac.value) {
-      return null;
-    }
-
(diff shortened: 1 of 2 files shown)


lmmrssa:
When trying to make multi day to single day event, it is not allowing to make end date empty. End time is validating fine.

src/app/validators/custom-validators.ts
```diff
@@ -112,7 +100,7 @@ export class CustomValidators {
       }
 
       // if start date has not been given a value yet return back
-      if (!startDate) {
+      if (!startDate || !endDate.value) {
```
paulbert: This should do to allow the end date to be blank after it has been given a value.

paulbert: @lmmrssa

PR #4964, opened 2019-10-04
fixed spinner( Fixed 4954)

### LINKED ISSUE #4954: TeamsView: When removing member spinner opens under dialog
There should not be a spinner opening below the dialog.  A spinner is only needed after **OK** is clicked to remove the member.

https://www.loom.com/share/da7ce366da904dcc8ebb3d6e88a6eb4a

wanglihe2010:
fixed #4954 

I moved the spinner in `changeMembershipRequest` function to `changeMembership` function. 

Adding `leader` is not related to this issue, but it should be an error.

### DIFF
--- a/src/app/teams/teams-view.component.ts
+++ b/src/app/teams/teams-view.component.ts
@@ -258,9 +258,10 @@ export class TeamsViewComponent implements OnInit, AfterViewChecked, OnDestroy {
 
   changeMembership(type, memberDoc?) {
     this.dialogsLoadingService.start();
-    this.changeMembershipRequest(type, memberDoc)().subscribe((message) => {
+    this.changeMembershipRequest(type, memberDoc)().pipe(
+      finalize(() => this.dialogsLoadingService.stop())
+    ).subscribe((message) => {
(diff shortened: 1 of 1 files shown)


src/app/teams/teams-view.component.ts
```diff
@@ -236,20 +236,20 @@ export class TeamsViewComponent implements OnInit, AfterViewChecked, OnDestroy {
   changeMembershipRequest(type, memberDoc?) {
     const changeObject = this.changeObject(type, memberDoc);
     return () => {
-      this.dialogsLoadingService.start();
       return changeObject.obs.pipe(
         switchMap(() => type === 'added' ? this.teamsService.removeFromRequests(this.team, memberDoc) : of({})),
         switchMap(() => this.getMembers()),
         switchMap(() => this.sendNotifications('added')),
         map(() => changeObject.message),
-        finalize(() => this.dialogsLoadingService.stop())
```
paulbert: Please keep this in the `finalize` operator.  This will ensure the dialog closes even if there is an error.

PR #4975, opened 2019-10-08
Health: Switch encryption to aes cbc

paulbert:
Still need to come up with key storage, but moving over to a method that matches some Java implementations we've seen.


paulbert:
@leomaxi What I tested for the above matches this:

http://tpcg.io/MDmZGK

paulbert:
Oh the above link is messed up for me.  You won't be able to see it working live, but you can see the implementation here:

https://gist.github.com/paulbert/05fb8794ef216746aeac4cc406f62b8c

paulbert:
Similar to the last time we can test with:

Adding a doc
```
curl -X PUT http://localhost:2200/health/_design/health/_update/encrypt -d '{"_id":"1","key":[1,2,3,4,5,6,7,8,9,0,1,2,3,4,5,6,7,8,9,0,1,2,3,4, 5,6,7,8,9,0,1,2],"cat":"zuzu"}' -H "Content-Type:application/json" 
```

Getting a doc
```
curl -X POST http://localhost:2200/health/_design/health/_show/decrypt/1 -d '{"key":[1,2,3,4,5,6,7,8,9,0,1,2,3,4,5,6,7,8,9,0,1,2,3,4,5,6,7,8,9 ,0,1,2]}' -H "Content-Type:application/json" -H "Accept:application/json"
```

Issue #4998, opened 2019-10-10
Console Error when jump to the community


ScottHuangNYU:
![image](https://user-images.githubusercontent.com/31773567/66583405-df6ac580-eb48-11e9-8de9-05f14aa2cd4c.png)

paulbert:
This could be related to #4990

PR #5002, opened 2019-10-11
Community: Fix message replies & edit (fixes #4990)

### LINKED ISSUE #4990: Community: message reply does not show up
community(only) message does not have `show replies` after adding some replies. 
image

paulbert:
#4990


paulbert:
Also will fix #4998

PR #5019, opened 2019-10-15
Added Office to the title for Enterprises(fixes #4932)

### LINKED ISSUE #4932: Enterprises: Add "Office" to title
We have a request to add "Office" at the end of the name for enterprises in `TeamsViewComponent`.

For example, in the below instead of **paul0.8.38 Enterprise 3** it should read **paul0.8.38 Enterprise 3 Office**

image

smitshrestha101:
Fixes #4932 
Added office to the title of all/any enterprises.

### DIFF
--- a/src/app/teams/teams-view.component.html
+++ b/src/app/teams/teams-view.component.html
@@ -4,12 +4,10 @@
 </mat-toolbar>
 <div class="space-container">
   <mat-toolbar class="primary-color font-size-1">
-      <h3 class="margin-lr-3 ellipsis-title" >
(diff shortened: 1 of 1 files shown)


smitshrestha101:
@paulbert Is this the change you are looking for?

paulbert:
@smitshrestha101 No.  You don't need the else block because `{{team.name}}` is already included in the if block.  The conditional should only affect **Office**.

smitshrestha101:
Thank you @paulbert , somehow I couldn't think that.

PR #5027, opened 2019-10-17
Change time input to 24h with ngx-mask(Fixed 4969)

### LINKED ISSUE #4969: MeetupsAdd: Input 24 hour time
We should change the below time inputs to use 24 hour rather than 12 hour time, if possible.

image

wanglihe2010:
fixed #4969


paulbert:
Pushing some changes to clean this branch, but the library is actually pretty bad for 24 hour.  It's not really doing a good job flagging invalid input & forces you to type in a leading 0.

I think switching this to a normal input that we can write our own mask for would be the faster way to get to a working 24 hour input.

paulbert:
@wanglihe2010 I have had a moment to think this over a little.  Could you try writing a [custom directive](https://angular.io/guide/attribute-directives) that will control the input & format the value for this field instead of the mask library?

Requirements:
1. It should ignore non-number key presses
2. There should be max 4 digits in the string
3. If any key press would make an invalid time (i.e. **2**5:55, 40:0**0**, or 1:**7**0) then that input should be ignored.
4. The colon should be added in automatically after three digits have been entered.

It might be easier to start with a new branch & PR, but feel free to work off this one if you want.

wanglihe2010:
> @wanglihe2010 I have had a moment to think this over a little. Could you try writing a [custom directive](https://angular.io/guide/attribute-directives) that will control the input & format the value for this field instead of the mask library?
> 
> Requirements:
> 
> 1. It should ignore non-number key presses
> 2. There should be max 4 digits in the string
> 3. If any key press would make an invalid time (i.e. **2**5:55, 40:0**0**, or 1:**7**0) then that input should be ignored.
> 4. The colon should be added in automatically after three digits have been entered.
> 
> It might be easier to start with a new branch & PR, but feel free to work off this one if you want.

@paulbert  Sure, I will do it

paulbert:
Closing in favor of #5173

PR #5066, opened 2019-10-24
Add a role Health Provider and save it as health to database(fixes 5051)

ScottHuangNYU:


ScottHuangNYU:
@paulbert you means change the element type of `roleList`  from string to object?

paulbert:
@ScottHuangNYU Change role list to an array of objects with { text, value } properties instead of an array of strings.  Then change the template and the `roleSubmit` function to use those properties.

PR #5081, opened 2019-10-26
Updated width of examination columns to 100px(fixes #5042)

### LINKED ISSUE #5042: Health: Thinner columns
Columns for the **Examinations** table should be thinner.  Let's make them `100px`.

image

smitshrestha101:
Fixes #5042 

Changed the min-width of the columns in Examination table to 100px.

### DIFF
--- a/src/app/health/health.scss
+++ b/src/app/health/health.scss
@@ -21,5 +21,5 @@ mat-table {
   width: max-content;
 }
 .table-column {
-  min-width: $exam-table-col-width;
+  min-width: 100px;
 }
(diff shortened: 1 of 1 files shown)


src/app/health/health.scss
```diff
@@ -21,5 +21,5 @@ mat-table {
   width: max-content;
 }
 .table-column {
-  min-width: $exam-table-col-width;
+  min-width: 100px;
```
paulbert: Instead of changing the css property, you should change the scss variable at the top of the file.  That variable is used in a couple of places to set the width of the columns.

PR #5086, opened 2019-10-28
myHealth : disable the submit button when form is empty(fixed 5046)

ScottHuangNYU:


src/app/health/health-event.component.ts
```diff
@@ -46,6 +46,17 @@ export class HealthEventComponent implements OnInit {
     });
   }
 
+  isEmptyForm() {
+    return this.healthForm.controls.temperature.value === '' && this.healthForm.controls.pulse.value === '' &&
```
paulbert: @ScottHuangNYU Please change to loop over the controls rather than explicitly checking each field.  This would make it more difficult to maintain if we add or remove form fields.

ScottHuangNYU: @paulbert I changed to loop over the controls. But I wonder why I tried to return false in the ` (this.healthForm.controls[key].value !== '') ` if statement instead of using a `notAllEmpty` to make the `isEmptyForm` function to return false after I made any input can not work?

src/app/health/health-event.component.ts
```diff
@@ -46,6 +46,20 @@ export class HealthEventComponent implements OnInit {
     });
   }
 
+  isEmptyForm() {
+    let notAllEmpty: Boolean;
+    Object.keys(this.healthForm.controls).forEach((key: string) => {
```
paulbert: Two things you can use here to improve:

1. Use [Object.values()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values) to access the values instead of the keys
2. Use [Array.some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) to return a true/false in one line

ScottHuangNYU: @paulbert I found the value from Object.values() is not the value I want. It still a AbstractControl Object. Let say the element from Object.values() is `v`, The value used to make a verification should be the `v.value`.

Issue #4878, opened 2019-09-19
DatePicker: Use Somali week


paulbert:
In Somalia the week starts on Saturday, but we only have this implemented for the calendar not the MatDatepickers.

We need to find the best way to set that:
https://material.angular.io/components/datepicker/overview

wanglihe2010:
@paulbert  
I tried to add a custom `DateAdapter` to set first day of week as showed in the this [example](https://stackblitz.com/edit/angular-dateadapter-example-stackoverflow1-oarzsz?file=app%2Fdatepicker-overview-example.ts), but it does not work. Do you know the reason? Please check this [commit](https://github.com/open-learning-exchange/planet/commit/aaaaaa0934c4beec666fffb65b18c0cd3cbbd9cd)

PR #5101, opened 2019-10-29
Change week first day to Saturday(Fixed 4878)

wanglihe2010:
#4878


wanglihe2010:
@paulbert  I was thinking to change the name of `custom-date-adapter.ts` but not sure what should be appropriate. And should I move the `custom-date-adapter.ts` from `add-meetup` folder to `shared` folder?

PR #5108, opened 2019-10-30
Couch per user

paulbert:
Sets the CouchDB into couch per user mode with the "databases" named like
```
userdb-<configurationcode>-<hexofusername>
```

Creates the setting in two places
1. When the configuration is initially created
2. In `couchdb-setup.sh` if the Planet configuration has been set.  This is used to upgrade current deployments and fix in case the CouchDB setting is changed.

### DIFF
--- a/couchdb-setup.sh
+++ b/couchdb-setup.sh
@@ -34,6 +34,16 @@ insert_dbs() {
   done
 }
 
+set_couch_per_user() {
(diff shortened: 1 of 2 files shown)


paulbert:
Modified this so it will only change if beta mode is enabled (either on a per user or planet wide basis).

Note turning beta mode off will not undo the setting.

PR #5114, opened 2019-10-31
HealthUpdate: Add email validator (fixes #5054)

### LINKED ISSUE #5054: myHealth : when the emergency contact type is email, if the email is invalid format. It shouldn't be allowed.

paulbert:
#5054


paulbert:
Noticed that the form validation doesn't block submission.  That can be fixed in another issue #5115

Issue #4913, opened 2019-09-26
Calendar: No of occurance exceed limit should show warning error message


singharpita:
right now it allows to enter any range and result in spinner loading
![Screenshot from 2019-09-26 20-34-31](https://user-images.githubusercontent.com/27276132/65699101-4a25f680-e09d-11e9-828b-769ccd0f5b1c.png)

paulbert:
We can refactor `CalendarComponent.weeklyEvents` to remove the recursion.  I thought the limit for modern browsers had increased significantly, but 800 is not enough.

PR #5137, opened 2019-11-05
[#4913] Limit recurring of meetup

singharpita:


paulbert:
@singharpita We shouldn't limit the number of recurrences if possible.  Let's try to refactor to remove the recursion in `weeklyEvents` and use a loop/iterable instead.

PR #5148, opened 2019-11-06
Dashboard: Dialog for unfinished surveys (fixes #5147)

### LINKED ISSUE #5147: Dashboard: Popup on login for unfinished surveys

paulbert:
Adds a dialog after a member logs in if they have any surveys to complete.


paulbert:
@singharpita Ok looks like it was working because I was setting a breakpoint while the component was loading.  Changed it so it works without the debugger.

PR #5160, opened 2019-11-07
Users: Show "Name" instead of username (fixes #5093)(#5160)

### LINKED ISSUE #5093: Users: Show "Name" instead of username
**Current Behavior**

In the table of **UsersComponent** we show the username as one of the columns

**Desired Behavior**

We should change the title of this column to "Name" and if `firstName` or `lastName` exists for the user show it as `{{firstName}} {{middleName}} {{lastName}}` and if not continue to show `{{name}}`

image

perryc85:


src/app/users/users.component.ts
```diff
@@ -122,7 +122,7 @@ export class UsersComponent implements OnInit, OnDestroy, AfterViewInit {
       ),
       filterDropdowns(this.filter),
       filterFieldExists([ 'doc.requestId' ], this.filterType === 'associated'),
-      filterSpecificFields([ 'doc.name' ])
+      filterSpecificFields([ 'doc.firstName', 'doc.lastName' ])
```
paulbert: @perryc85 In reviewing this, I realize that having this split in the filter means that if you search for the **first name** and **last name** together the name will not show up.  The solution is to make a change in `initializeData()` to add a property to the `userInfo` object called `fullName` with the same logic in the template.

That way the search can simply use `fullName` and will correctly return first & last name searches.

PR #5172, opened 2019-11-11
Health: Provider access to health docs

paulbert:


paulbert:
As I was implementing this I realized that to give this the best chance of a smooth transition in our current deployments we should update the database security on login rather than later in the process.

PR #5220, opened 2019-11-19
Fix newline character in database dates (fixes #5195)

### LINKED ISSUE #5195: Newline character in dates on database
This is a strange one that I just noticed.  All dates added to the database have a newline character at the end of them.  This may be specific to the `prod` vagrant, so we should also test on the servers and RPi.

image

image

paulbert:


paulbert:
Should be tested in production with
```
treehouses/planet-tags:0.9.27-5195-newline-date-a4d73cec
```

The issue should not exist in development because that does not retrieve the current time from the server.

PR #5227, opened 2019-11-20
Users: Refactored name and search (fixes #5216)

### LINKED ISSUE #5216: Users: Refactor name & search
There are a few changes that need to be made to the **UsersComponent**

1. In the HTML template the **Name** column should simply display `element.fullName` instead of the `*ngIfElse` blocks.
2. The `fullName` property of `userInfo` should be `` `${user.firstName} ${user.middleName} ${user.lastName}` `` if `user.firstName` or `user.lastName` exists, otherwise it should be `user.name`.

perryc85:


src/app/users/users.component.html
```diff
@@ -86,12 +86,9 @@
       <ng-container matColumnDef="name">
         <mat-header-cell *matHeaderCellDef mat-sort-header i18n>Name</mat-header-cell>
         <mat-cell *matCellDef="let element">
-          <span *ngIf="element.doc.firstName || element.doc.lastName; else origName">
-            {{element.doc.firstName}} {{element.doc.middleName}} {{element.doc.lastName}}
+          <span>
+            {{ element.doc.firstName || element.doc.lastName ? element.fullName : element.doc.name }}
```
paulbert: This conditional should be the value for `fullName` so the search works when `element.doc.name` is used.

perryc85: Hey Paul, I have a question. Do you mean something like this ... `{{ element.fullName ? element.fullName : element.doc.name }}` ?

Issue #5232, opened 2019-11-21
myHealth: Update Profile not visible & title incorrect


paulbert:
In **HealthComponent** there is a button to **Update Profile** in the blue bar to the left of **Add Examination**.

**Current Behavior**

It does not show up at all and the title of the page is always **Health Care**.

**Desired Behavior**

When viewing your own member health profile by clicking **myHealth** on the dashboard:
- The **Update Profile** button should be visible
- The title of the page should be **myHealth**

When viewing a different member health profile:
- The **Update Profile** button should _not_ be visible
- The title of the page should be **Health Care**

**Notes**

These view changes are supposed to be controlled by a boolean variable `isOwnUser`, but it appears the value of that variable is always false.

PR #5246, opened 2019-11-25
Health username is shown (fixes #5233)

### LINKED ISSUE #5233: Health: username should be shown in Health Care Profile when no other details is present
Its kind of confusing to know, which user information we are updating in case User information is missing.
Screenshot from 2019-11-21 20-31-15
Screenshot from 2019-11-21 20-32-20

Can username should be displayed at top similar member profile?
Screenshot from 2019-11-21 20-34-52

robinGiri:
issue #5233 

### DIFF
--- a/src/app/health/health.component.html
+++ b/src/app/health/health.component.html
@@ -5,9 +5,9 @@
 </mat-toolbar>
 
 <div class="space-container">
-  <mat-toolbar class="primary-color font-size-1">
-    <a mat-raised-button color="accent" class="margin-lr-3" i18n [routerLink]="['update']" *ngIf="isOwnUser">Update Details</a>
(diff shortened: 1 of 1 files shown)


src/app/health/health.component.html
```diff
@@ -8,6 +8,7 @@
   <mat-toolbar class="primary-color font-size-1">
     <a mat-raised-button color="accent" class="margin-lr-3" i18n [routerLink]="['update']" *ngIf="isOwnUser">Update Details</a>
     <a mat-raised-button color="accent" class="margin-lr-3" i18n [routerLink]="['event', { id: userDetail._id }]">Add Examination</a>
+    <span *ngIf="userDetail.firstName === undefined" >Member login: {{userDetail.name}}</span>
```
lmmrssa: needs space before > removed.
also based on #5232 for own health record **myHealth** will be title so **Member login** is not required

PR #5264, opened 2019-11-27
[#5215] Health record export

singharpita:


src/app/health/health.component.ts
```diff
@@ -75,4 +79,16 @@ export class HealthComponent implements OnInit, AfterViewChecked {
     this.displayedColumns = Object.keys(this.eventTable.data[0]);
   }
 
+  exportCSV() {
+    const data = this.eventTable.data.map(h =>
+      Object.entries(h).reduce((object, [ key, value ]: [ any, any ]) => {
```
paulbert: Please fix the indentation in this block

src/app/health/health.component.html
```diff
@@ -7,6 +7,8 @@
   <mat-toolbar class="primary-color font-size-1">
     <a mat-raised-button color="accent" class="margin-lr-3" i18n [routerLink]="['update']">Update Details</a>
     <a mat-raised-button color="accent" class="margin-lr-3" i18n [routerLink]="['event', { id: userDetail._id }]">Add Examination</a>
+    <span class="toolbar-fill"></span>
+    <a mat-raised-button color="primary" class="margin-lr-3" i18n (click)="exportCSV()" >Export Examination</a>
```
paulbert: This should be the accent color or a mat-stroked-button so it does not blend into the toolbar

src/app/health/health.component.ts
```diff
@@ -75,4 +79,16 @@ export class HealthComponent implements OnInit, AfterViewChecked {
...
+        key = isNaN(key) ? key : this.datePipe.transform(key, 'shortDate');
```
paulbert: A better solution (and a solution to the only recent date shown) would be to convert the date from a UNIX timestamp to the Excel number.  You could put that in **CsvService**.

```
= event.date / 86400 + 25569
```

From:
https://excelribbon.tips.net/T010849_Converting_UNIX_Date_Time_Stamps.html

paulbert (changes requested):
This is a good start, but the profile information & other notes from each examination should also be exported.

Also if there are two examinations on the same day, only the latest one will be exported.

PR #5282, opened 2019-12-02
Fix travis error

lmmrssa:


paulbert:
@lmmrssa Ok that fix is not working even on the correct stage.  Still seeing the error with `npm.config.loaded` sometimes.

https://travis-ci.org/open-learning-exchange/planet/jobs/620727684#L299

PR #5302, opened 2019-12-04
TeamsViewFinance: total after filter (fixes #5287)

### LINKED ISSUE #5287: TeamsViewFinances: Show total after date filter
**Current Behavior**

When the date range is filtered in **TeamsViewFinancesComponent**, the _Total_ column is removed.

**Desired behavior**

When the date range is filtered the _Total_ column should remain and be updated to reflect the date range being viewed.

image

robinGiri:
issue #5287 


paulbert:
Looks like the total Credits & Debits are gone:

image

This should be the total of the column & should update based on the date range selected.

PR #5384, opened 2019-12-17
Change production English locale (fixes #2794)

### LINKED ISSUE #2794: Change locale to en-GB?
Since we're not deploying in the US, we should consider changing our app locale to `en-GB` so the dates are in an internationally more common format.

@dogi @lmmrssa Thoughts?

paulbert:
Trying out `en-GB`, could use `en-EU` instead.

Change is so the app uses 24 hour time and euro style date formats.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ jobs:
     - stage: docker-creation
       name: "English"
       script:
-        - bash ./.travis/deploy_single_language.sh 'eng' 'en'
+        - bash ./.travis/deploy_single_language.sh 'eng' 'en-GB'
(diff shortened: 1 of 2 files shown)


paulbert:
Doing my own tests, it looks like this causes problems with the datepicker forms where the date and month switch after typing.  For example typing `5/12/2019` will immediately change to `12/5/2019` when inputting the form or moving to another field.

I think we need to incorporate Moment.js for this to work since the default `DataAdapter` only supports en-US (see https://material.angular.io/components/datepicker/overview#choosing-a-date-implementation-and-date-format-settings)

PR #5395, opened 2019-12-19
Certification: Add & remove courses

paulbert:
Note this is still in a hidden route `manager/certifications`

Adds the functionality to add & remove courses from a certification.


lmmrssa:
Adding certificate and courses working.
Some improvement points are
* Make certificate unique
* Enable remove course only if some course are selected
* Fix issue of Save Return not working after clicking Save

PR #5411, opened 2019-12-20
TeamsViewFinance: validation message (fixes #5406)

### LINKED ISSUE #5406: TeamsViewFinance: validation message
steps to replicate issue:
1. Go to community finance
2. Add new transaction 

While adding  **amount** field if you write letters then validation message says **This field is required** 
finance

robinGiri:
issue: #5406 

### DIFF
--- a/src/app/validators/custom-validators.ts
+++ b/src/app/validators/custom-validators.ts
@@ -13,7 +13,7 @@ export class CustomValidators {
     if (!ac.value) {
       return null;
     }
-    return (ac.value > 0) ? null : (ac.value < 0) ?  { invalidPositive : true } : { invalidInt : true };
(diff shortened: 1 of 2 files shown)


src/app/validators/custom-validators.ts
```diff
@@ -13,7 +13,7 @@ export class CustomValidators {
     if (!ac.value) {
       return null;
     }
-    return (ac.value > 0) ? null : { invalidPositive : true };
```
paulbert: This changes the definition of this function in an unexpected way.  Let's try using the `integerValidator` instead.

robinGiri: @paulbert  **integerValidator** is not working as we have thought in this issue. So can you suggest me another validator which works in this condition.

PR #5424, opened 2019-12-24
Users: Notify users when their role changes (fixes #5288) 

### LINKED ISSUE #5288: Users: Notify users when their role changes?
Should we notify a user when they are promoted, demoted, or assigned an additional role other than their default learner role?

perryc85:


src/app/users/users.service.ts
```diff
@@ -190,9 +193,30 @@ export class UsersService {
...
+      link,
+      linkParams: { activeTab: 'taskTab' },
+      'type': 'newRole',
+      'priority': 1,
+      'status': 'unread',
+      'time': this.couchService.datePlaceholder,
+      userPlanetCode: user.userPlanetCode
+    };
+    return this.couchService.findAll(
+      'notifications',
```
paulbert: This needs a specific query since this should be checking to see if a role change notification already exists for this user.  Right now this `_find` query is only checking to see if any notifications exist for the entire system.

src/app/users/users.service.ts
```diff
@@ -190,9 +193,30 @@ export class UsersService {
...
+    ).pipe(
+      switchMap((res: any[]) => res.length > 0 ? this.couchService.updateDocument('notifications', notificationDoc) : of({}))
+    ).subscribe((users: any) => users);
```
paulbert: Ideally this would not include the `.subscribe` and `sendNotifications` would be called as part of a `switchMap` after the role change was completed on the database.

src/app/users/users.service.ts
```diff
@@ -212,11 +212,11 @@ export class UsersService {
       'time': this.couchService.datePlaceholder,
       userPlanetCode: user.userPlanetCode
     };
-    return this.couchService.findAll(
-      'notifications',
-    ).pipe(
-      switchMap((res: any[]) => res.length > 0 ? this.couchService.updateDocument('notifications', notificationDoc) : of({}))
```
paulbert: The documents should still be added to the `notifications` database.

src/app/users/users.service.ts
```diff
@@ -140,7 +143,9 @@ export class UsersService {
       roles: roles,
       oldRoles: [ ...user.roles ] || [ 'learner' ],
     };
-    return this.couchService.put('_users/org.couchdb.user:' + tempUser.name, tempUser);
+    return this.couchService.put('_users/org.couchdb.user:' + tempUser.name, tempUser).pipe(
+      switchMap(this.sendNotifications(user))
```
paulbert:
```
switchMap(() => this.sendNotifications(user))
```

src/app/users/users.service.ts
```diff
@@ -189,10 +194,30 @@ export class UsersService {
...
+      'message': 'You were assigned a new role',
+      link,
+      linkParams: { activeTab: 'taskTab' },
+      'type': 'newRole',
+      'priority': 1,
+      'status': 'unread',
+      'time': this.couchService.datePlaceholder,
+      userPlanetCode: user.userPlanetCode
+    };
+    return this.couchService.post('newRole/_find',
```
paulbert:
```
return this.couchService.findAll('notifications/_find',
```

src/app/users/users.service.ts
```diff
@@ -189,10 +194,30 @@ export class UsersService {
...
       return observers;
     }, [])).pipe(map((responses) => this.stateService.requestData(this.dbName, 'local')));
   }
 
+  sendNotifications(user) {
+    const link = '/manager/users';
+    const notificationDoc = {
+      user: user._id,
+      'message': 'You were assigned a new role',
+      link,
```
paulbert: Let's just have `link: '/'`

Let's also remove `linkParams`

PR #5435, opened 2019-12-27
Community: Notify for reply news(fixes #5379)

### LINKED ISSUE #5379: Community: Notify community when a new news alert is posted
Should we alert community members every time a new news story is posted?

robinGiri:
issue: #5379 
Thanks for your review in advance


src/app/news/news-list.component.ts
```diff
@@ -82,6 +87,25 @@ export class NewsListComponent implements OnChanges {
...
+      link,
+      'type': 'replyMessage',
+      'priority': 1,
+      'status': 'unread',
+      'time': this.couchService.datePlaceholder,
+      userPlanetCode: oldNews.messagePlanetCode
+    };
+    return this.couchService.findAll( 'notifications', ).pipe(
+      switchMap((data: any[]) => data.length > 0 ? this.couchService.updateDocument('notifications', notificationDoc) : of({}))
+    ).subscribe((news: any) => news);
```
paulbert: `(news: any) => news` doesn't do anything, and should be removed.

src/app/news/news-list.component.ts
```diff
@@ -82,6 +87,25 @@ export class NewsListComponent implements OnChanges {
       this.dialogsFormService.closeDialogsForm();
       this.dialogsLoadingService.stop();
     });
+    this.sendNewsNotifications(oldNews);
+  }
+
+  sendNewsNotifications(oldNews) {
+    const userId = this.userService.get()._id;
+    const link = '/community';
```
paulbert: **NewsListComponent** is not specific to the community news, so the link could be for a team or enterprise.  You'll need to construct this from `this.viewableBy` and `this.viewableId`.

src/app/news/news-list.component.ts
```diff
@@ -82,6 +87,25 @@ export class NewsListComponent implements OnChanges {
       this.dialogsFormService.closeDialogsForm();
       this.dialogsLoadingService.stop();
     });
+    this.sendNewsNotifications(oldNews);
+  }
+
+  sendNewsNotifications(oldNews) {
+    const userId = this.userService.get()._id;
```
paulbert: This is not the correct user to send the notification to.  It should be sent to the author of the message that was replied to, not the currently logged in user.

src/app/news/news-list.component.ts
```diff
@@ -82,6 +87,25 @@ export class NewsListComponent implements OnChanges {
...
+    const link = '/community';
+    const notificationDoc = {
```
paulbert: We should also add `item: oldNews.replyTo` to this object so we can identify which message this is related to.

src/app/news/news-list.component.ts
```diff
@@ -82,6 +87,25 @@ export class NewsListComponent implements OnChanges {
...
+      user: userId,
+      'message': 'Someone replyed to your community message.',
+      link,
+      'type': 'replyMessage',
+      'priority': 1,
+      'status': 'unread',
+      'time': this.couchService.datePlaceholder,
+      userPlanetCode: oldNews.messagePlanetCode
+    };
+    return this.couchService.findAll( 'notifications', ).pipe(
```
paulbert: The find needs a query, otherwise the notification will not be sent once there is any notifications in the system.

PR #5436, opened 2019-12-30
Health: Display special needs & notes with markdown (fixes #5181)

### LINKED ISSUE #5181: Health: Show markdown for Special Needs and Notes
The **Special Needs** and **Notes** fields are inputted with a markdown editor and need to be displayed in the rich markdown rather than plaintext.  The `td-markdown` component can be used to turn the text into markdown.

image

paulbert:


paulbert:
#5181

PR #5438, opened 2019-12-30
ExamsView: Improved overflow (fixes #5437)

### LINKED ISSUE #5437: ExamsView: Overflow not handled correctly
This creates two different problems:

1. In the normal view, the question could overflow in a way that forces the user to scroll the menu out of view.  This is not breaking, but not ideal.
2. In the preview dialog view, the answer input & next question button are off the screen without any way to scroll to them.  This is breaking that functionality.

paulbert:


paulbert:
#5437

PR #5439, opened 2019-12-30
Resources: limiting lenght of name(fixes #4756)

### LINKED ISSUE #4756: Resources : limiting the length of name of new collection
when creating new collection in resources,
when we add a very long name to collection,
edit and delete button disappears.

robinGiri:
issue : #4756 


paulbert:
Putting on hold.  I think we need to come up with a better way of handling this since 30 characters could be too long for smaller screens but isn't enough for descriptive names.  30 is smaller than the longest collection name in production.

PR #5472, opened 2020-01-03
PDF: Export Survey submission data

paulbert:
Adds the option to export survey submissions from the **SurveysComponent** (pictured below).

Note you will have to run an `npm install` from `/vagrant` in the vagrant to add the new packages.  We've noticed some problems with `npm install` in windows.  If there is an error you can try the `--no-bin-links` flag which may fix the issue.

image


paulbert:
Note I had to update many of the packages to their latest versions to fix a build error in Travis (wasn't showing up locally so there wasn't much I could do to debug).

lmmrssa:
Seeing issue with Markdown editor after update in packages.
Screen Shot 2020-01-06 at 3 01 42 PM

Issue #5591, opened 2020-01-21
Surveys: Option to only export answers in PDF


paulbert:
When exporting survey responses we should have a dialog pop up which allows the user to select if the export will include the questions or the answers.  If only questions are selected, then we should only export one copy of the questions of the survey (not one for every submission).

In the field we have some very long questions in some surveys which make reading the responses difficult.

In this case, being able to print out one copy with just questions for reference and then print out all of the responses would be useful.

paulbert:
Dialog should have two checkboxes labeled **Include Questions** and **Include Answers** which both default to checked.  If neither are checked, the OK button to export should be disabled.

PR #5604, opened 2020-01-22
[#5591] Export survey pdf options

singharpita:


singharpita:
@paulbert  Haven't put the validation instead implemented check on exportPDF itself for now. If you suggest on form validation i will try to look into it.

PR #5625, opened 2020-01-24
5029 share between community

lmmrssa:


paulbert (commented):
A couple of things:

1. I think we should limit this to resources created on the community.  If we expand it then there's a large risk of this generating conflicts all over the system.
2. I'm not sure about the way this gets sent to the nation.  I understand that the resource needs to be sent to the nation to allow the second community to replicate it, but this seems like a backdoor route for people to put whatever they want into the nation library.  If we do it this way I would suggest we have a "pending resources" on the Nation first.
3. Related to the above, I thought we were going to limit this to hubs sending to their children?  Doing it that way would make the points above less necessary (but I still think we should work towards solutions for them).

@dogi I'm curious about your thoughts about this as well.

dogi:
think we should activate this feature only on HUBs ;)

oops started to write before reading point 3
@paulbert would say I agree with all points

lmmrssa:
@paulbert yes it send copy of resource to nation for replication. It is same as we are sending resources from community to nation right now.
Also it will be limited to hub as they will only see community under hub when they try to send it.
Regarding Pending Resource you mean resource which needs approval before showing upon library?

paulbert:
@lmmrssa This still is not limited to hubs.  Any community can send any resource or course to any other community.

We need to limit this so only hubs can send resources or courses to their specific communities.

PR #5666, opened 2020-01-30
Community: Show editing for local admins (fixes #5661)

### LINKED ISSUE #5661: Community: Service & Finances editing not available
Add Transaction button disappear after adding transaction.
Screen Shot 2020-01-30 at 8 58 32 PM

paulbert:
#5661 

Fixed the condition for showing the editing buttons for Services and Finances.  They were not showing up because the `planetCode` is based on the route parameters, and for `/` that is `null`.


paulbert:
We should not worry about the case where the user could type in `/community/{local community code}` because the local community information will not be in the `communityregistrationrequests` database  and that page will not load any of the community information.

PR #5674, opened 2020-02-03
TeamsView: User member name message(fixes #5659)

### LINKED ISSUE #5659: Teams: Use member name messages
**Current Behavior**

In the confirmation dialog and the snackbar message for changing team membership status the username is used instead of the member's real name.

**Desired Behavior**

If there is a `firstName`, we should use the full name (`firstName + middleName + lastName` with appropriate spaces).  If not we can fall back on the username (`name` property).

image

image

image

robinGiri:
Issue: #5659 


paulbert:
`openDialogPrompt` is a function used for more than just member changes.

Throws this error with resources:
image

I think this needs a little bit of a larger refactor, so will try to implement now for testing tonight.

PR #5699, opened 2020-02-05
UsersTable: Fix filters for user planet (fixes #5697)

### LINKED ISSUE #5697: Users: Filter to view child users not working

paulbert:
#5697 

First noticed that none of the `child_users` were going through to the table and made some adjustments to improve `StateService` and `UsersService` which cleared that up.

Next I found that when **Local** was selected it was also showing users from `child_users` so I fixed that filter also.

Note I think there may be improvements that we can make later, but first we need to improve the load time of this screen because it is difficult to work on when a refresh can take 1-3 minutes.


paulbert:
This will conflict with #5702.  Prioritize that one first.

PR #5701, opened 2020-02-05
issue5095 fix

AlekseiKorolev:
+ add a badge in the upper right corner
+ associated it with a variable that depends on all messages

### DIFF
--- a/src/app/teams/teams-view.component.html
+++ b/src/app/teams/teams-view.component.html
@@ -37,7 +37,10 @@
   </mat-toolbar>
   <div class="view-container view-full-height">
     <mat-tab-group [(selectedIndex)]="tabSelectedIndex">
-      <mat-tab i18n-label label="Chat" *ngIf="team?.public === true || userStatus === 'member'">
+      <mat-tab *ngIf="team?.public === true || userStatus === 'member'">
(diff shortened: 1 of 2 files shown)


AlekseiKorolev:
I add logic for select only root message and it works (201-203 lines)
```javascript
this.messageCount = this.members[0].messagesList
  ? this.news.filter(message => this.members[0].messagesList.find(message._id) === -1 && !message.replyTo).length
  : this.news.filter(message => !message.replyTo).length;
```
I add function to store and update the list of read messages (171-175)
This doesn't work and doesn't produce any error. If I halt and up dev again I'll get "Cannot get" error
```javascript
updateReadMessages() {
  const messagesList = this.news.map(message => message._id);
  const member = this.members.filter(item => item.userId === this.user._id);
  this.couchService.updateDocument(`${this.dbName}/${member._id}`, { messagesList });
}
```
I tried add logic to set the messages as read if the user selected the "chat" tab (105-107)
But this doesn't work, because 0 (the index if the "chat" tab) already used as the initial value
```javascript
if (activeTab && this.tabSelectedIndex === 0) {
  this.messageCount = this.updateReadMessages();
}
```

AlekseiKorolev:
open anoter

PR #5723, opened 2020-02-07
Teams: Task header and list inconsistance(fixes #5686)

### LINKED ISSUE #5686: Teams: task header UI inconsistency
header "**task**" is not consistent when comparing with team leader and others members.
also top margin would be better in case of other members.
Screenshot from 2020-02-04 08-49-43

robinGiri:
issue: #5686 


paulbert:
See my comment on #5686.  Closing.

PR #5706, opened 2020-02-06
5672 edit header and message is now fixed (fixes #5672)

gyawaliamit7:
Screen Shot 2020-02-06 at 12 27 06 AM

Screen Shot 2020-02-06 at 12 27 10 AM

PR #5727, opened 2020-02-08
5672 editing header

gyawaliamit7:


paulbert:
Closing.  This appears to be just the first two commits from #5706

Issue #5467, opened 2020-01-02
Validation: Uniqueness with leading & trailing spaces


singharpita:
1. Add a certification of value "new"
2. now try to add again same vallue "new". it shows value already exist validation message
3. now insert space such that now it becomes " new"
4. Name is not validated this time and it allows to save the value
![Screenshot from 2020-01-02 20-54-30](https://user-images.githubusercontent.com/27276132/71674002-4099e200-2da2-11ea-9860-837178d745b2.png)

paulbert:
Updating title since this is an issue with the uniqueness validation in general.

This is low priority because the production deployments should be able to manage this if needed.  We can help with the leading and trailing spaces by updating the regex for the `isUnique$` to check for any number of leading & trailing spaces on the database and trim those from `ac.value`.

The main thing is that we should try to fix this with only changes to `isUnique$` in `ValidatorService` to keep the fix maintainable.

PR #5732, opened 2020-02-10
[#5467] Unique whitespaces

singharpita:


singharpita:
@paulbert regex was added to check if there was already some certificate with whitespaces in title. Added trim to clean current input.

Issue #5660, opened 2020-01-30
Tasks: Not showing member profile image


paulbert:
Found an issue where the member profile image is not found for a task.  It will need replication to find the exact cause, but the issue is related to the fact that when changing profile images the file name is set to either `img` or `img_` to force the overwrite of the file.

The member in this example has an image in the `attachments` database saved as `img_`, but the app is looking for `img`.

![image](https://user-images.githubusercontent.com/9203229/73462199-64245e00-4349-11ea-81b5-7f0e5023bf04.png)

jessewashburn:
Related to #4735

PR #5759, opened 2020-02-12
[#5660] Use attachments database for image

singharpita:


singharpita:
@paulbert found that attachments database update was not triggered and it was affecting community and team pages, so added that as fix in this PR. Will look into task specific image issue as well.

PR #5760, opened 2020-02-12
Travis fail fullcalendar package

lmmrssa:


paulbert:
Fullcalendar fixed their issue today, so builds are working now.

PR #5763, opened 2020-02-12
5741 selected collection list name is now declared correctly (fixes #5741)

### LINKED ISSUE #5741: Selection collection list name is not shown in Courses/Resources
1. select any collection title from collection.
2. nothing is shown
Screenshot from 2020-02-11 08-35-28
Screenshot from 2020-02-11 08-34-54

gyawaliamit7:
Screen Shot 2020-02-12 at 1 20 09 PM

### DIFF
--- a/src/app/resources/resources.component.html
+++ b/src/app/resources/resources.component.html
@@ -24,7 +24,7 @@
       <button *ngIf="!parent && myView !== 'myPersonals'" mat-mini-fab routerLink="add"><mat-icon>add</mat-icon></button>
       <div class="column margin-lr-5">
         <planet-filtered-amount [table]="resources"></planet-filtered-amount>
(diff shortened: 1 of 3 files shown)


gyawaliamit7:
> This is showing the tag id instead of the text of the tag. It's important to show the text here.

the problem was alltag shows the list of array initially but later the list in null. In the down figure  down of the line of "let's see what is printed is 1" is an object of alltag which is loaded successfully but again in the same runtime there is "let's see what is printed is 0" where object of alltag is set to null. As the null value is being passed that is the reason default tagname is being shown. Should i try to access all tag directly from couchDB in services rather that passing it from resources.component.html or there is another better approach to solve the problem? 

Screen Shot 2020-02-13 at 3 24 11 AM

paulbert:
@gyawaliamit7 Closing.  Please do not open a new PR for changes on a PR.  Simply add commits to the existing branch.

Issue #5073, opened 2019-10-24
Notifications: New resource notification


paulbert:
It would be a nice addition to have a notification for when new resources are added to the community.  When a new resource is added, a notification for each member of the planet should be created if there isn't one with `status === 'unread'`:

> There are new resources in the Library.  Click to see them!

The link should go to `/resources;sort=createdDate` and the sort parameter should set the sorting on the table to initialize with the **Created On** column in descending order.

EDIT: Changing link parameter

paulbert:
Note the syntax for going to that link is:

```
[ 'resources', { sort: 'createdDate' } ]
```

paulbert:
Next step for this would be to fix the `sort` parameter being dropped & add the functionality for the initial sorting.

PR #5777, opened 2020-02-13
New resource notification feature added (connects #5073)

sjkadali:
Issue #5073 

@paulbert , @lmmrssa please review. Thanks in advance.

### DIFF
--- a/src/app/resources/resources-add.component.ts
+++ b/src/app/resources/resources-add.component.ts
@@ -195,7 +195,6 @@ export class ResourcesAddComponent implements OnInit {
           (resourceRes) => this.afterResourceUpdate(message, resourceRes),
           (err) => this.planetMessageService.showAlert('There was an error with this resource')
         );
-        this.resourcesService.sendResourceNotification(newResource);
       } else {
         this.afterResourceUpdate(message);
       }
@@ -222,7 +221,8 @@ export class ResourcesAddComponent implements OnInit {
       file,
       { newTags: this.tags.value, existingTags: this.existingResource.tags }
     ).pipe(
-      switchMap(([ res ]) => this.couchService.get(`resources/${res.id}`))
+      switchMap(([ res ]) => this.couchService.get(`resources/${res.id}`)),
+      switchMap(( resourceRes ) => this.resourcesService.sendResourceNotification(resourceRes))
     );
   }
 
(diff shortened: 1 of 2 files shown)


src/app/resources/resources.service.ts
```diff
@@ -128,4 +131,38 @@ export class ResourcesService {
...
+    const resourceNotification = forkJoin([
+      this.usersService.getAllUsers(),
+      this.couchService.findAll('notifications', findDocuments({ link: 'resources', type: 'newResource', status: 'unread' }))
+    ]).pipe(
+      switchMap(([ users, notifications ]: [ any[], any[] ]) => {
+        users.filter(user => {
+          return (currentUser.name !== user.name &&
+            user.name !== 'satellite' &&
+            notifications.every(notification => notification.user !== user._id));
+        }).map(user => resourceDocs.push(this.addResourceDoc(user, createdDate)));
```
paulbert: `Array.map` and `Array.filter` return a modified array.  You should use that value rather than performing a side effect to build a different array.

src/app/resources/resources.service.ts
```diff
@@ -128,4 +131,38 @@ export class ResourcesService {
     return mediaTypes.find((type) => mimeType.indexOf(type) > -1) || 'other';
   }
 
+  sendResourceNotification(resource) {
+    const createdDate = resource.createdDate;
+    const currentUser = this.userService.get();
+    const resourceDocs = [];
+    const resourceNotification = forkJoin([
```
paulbert: It would be preferred to return the `forkJoin` without any subscription so if we add calls to this function in different components the other components can control what happens in success and error cases.

As a note, you don't need to assign this to a value and then immediately call it.  In the future if you do that you should remove the assignment and just call the function.

src/app/resources/resources-add.component.ts
```diff
@@ -195,6 +195,7 @@ export class ResourcesAddComponent implements OnInit {
           (resourceRes) => this.afterResourceUpdate(message, resourceRes),
           (err) => this.planetMessageService.showAlert('There was an error with this resource')
         );
+        this.resourcesService.sendResourceNotification(newResource);
```
paulbert: After changing the function to `return forkJoin...`, you should add this in another `switchMap` in the pipe for `updateResource`

sjkadali:
@paulbert thanks for the review. Have made the requested changes. Please review. Thanks!

PR #5780, opened 2020-02-13
Certifications: Add "View" page

paulbert:


lmmrssa:
We can remove selection checkbox on course list on certification view.
Disable Remove courses if none of course selected on Certification Add/Update

Issue #5779, opened 2020-02-13
Users: Remove manager promotion


paulbert:
We would like to remove the ability to promote members to the **Manager** role.  Any existing members that already have that role should keep it, but the **Promote** button in **UsersComponent** should only promote to the **Admin** role.  The menu that opens when that button is clicked can be removed.

PR #5784, opened 2020-02-14
[#5779] Disable Manager Promote

singharpita:


singharpita:
Demote function is still there as we might have manager roles that needs to be demoted.

PR #5812, opened 2020-02-20
CoursesProgress: Fix duplicate docs (fixes #5293)

### LINKED ISSUE #5293: CoursesStepView: Duplicate docs created in courses_progress
When viewing a step for the first time, two documents are created in the `courses_progress` database when there should only be one.

paulbert:
Docs in the `courses_progress` database should now be unique by `userId`, `courseId`, and `stepNum`


lmmrssa:
Code complexity issue but accepting it to keep it moving we can fix code complexity later.

PR #5875, opened 2020-03-01
5845 myplanet record now shows "no record found" 

gyawaliamit7:


src/app/logs-myplanet/logs-myplanet.component.html
```diff
@@ -24,5 +24,8 @@
         <planet-myplanet-table dataType="logs" [data]="planet.children"></planet-myplanet-table>
       </mat-expansion-panel>
     </ng-container>
+    <ng-container *ngIf="checkEmptyLogs()">
```
paulbert: @gyawaliamit7 This is another example of a function call in a template, something that we should avoid.  Angular is not able to perform change detection on this, so it will call the function and rerender the template far more than necessary.

Please make the condition based on a property.

src/app/logs-myplanet/logs-myplanet.component.ts
```diff
@@ -62,4 +62,13 @@ export class LogsMyPlanetComponent implements OnInit {
     }, (error) => this.planetMessageService.showAlert('There was a problem getting ' + this.childType));
   }
 
+  checkEmptyLogs() {
+    this.apklogs.forEach(element => {
```
paulbert: This doesn't return what you intend.  The `return false` is in the scope of the function parameter for `forEach` and does not affect `checkEmptyLogs`.  `checkEmptyLogs` always returns `true`.

Generally, you should not return a value in a `forEach` function since that always returns `undefined`.  Returning a value makes the code harder to read because at first glance it looks like it is doing something it is not.

PR #5919, opened 2020-03-06
Changed Task-username to link and direct to profile (fixes #5913)

### LINKED ISSUE #5913: Task: username should be made link and redirect to profile
Screenshot from 2020-03-05 21-42-52

sjkadali:
fixes #5913

### DIFF
--- a/src/app/meetups/view-meetups/meetups-view.component.ts
+++ b/src/app/meetups/view-meetups/meetups-view.component.ts
@@ -167,15 +167,8 @@ export class MeetupsViewComponent implements OnInit, OnDestroy {
     this.meetupService.openDeleteDialog(this.meetupDetail, callback);
   }
 
-  routeToCreator(createdBy: string) {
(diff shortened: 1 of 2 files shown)


sjkadali:
@paulbert , @lmmrssa please review the PR. Thanks!

src/app/meetups/view-meetups/meetups-view.component.ts
```diff
@@ -174,4 +174,11 @@ export class MeetupsViewComponent implements OnInit, OnDestroy {
     }
   }
 
+  routeToAssignee(assignedTo: string) {
```
paulbert: It looks like this is the same as the `routeToCreator` function on this component.  Instead of copying the function you can just use the other one.

For better naming let's call it `routeToUser` and call the variable `username`.

sjkadali:
@paulbert thanks for the review. Have made the requested changes. Please review. Thanks!

PR #6009, opened 2020-03-23
CourseView: Take/Retake/Continue in course step (fixes #5958)

### LINKED ISSUE #5958: CoursesView: Take Test button text changes on hide/show of course step detail
- go to any course that has atleast two steps with each having their own test.
- submit the first test without clicking on "Finish test". That will result in "Continue test" status
- now go to CoursesViewComponent.
- toggle the First step with "Continue test" status. leave it as it is
- toggle the second step.
- Now both step shows status "Take test"
1edca592-fc96-419e-8b3a-41fc866aa3ad

robinGiri:


paulbert:
Did a quick debug and realized we only need a very small change.  Rather than reverting the changes here opening a new PR.

PR #6030, opened 2020-03-26
ExamsView: Prefer exams database to embedded course (fixes #5974)

### LINKED ISSUE #5974: Courses: Use test/survey from exams database not course document
### Current Behavior

When loading a test or survey in a course step in **ExamsViewComponent**, we use the data embedded in the course document instead of the document in the `exams` database.

### Desired Behavior

Let's instead use the document in the `exams` database, and fallback on the data embedded in the course document only if that does not exist (possible if there was a replication error).

paulbert:


paulbert:
#6030 

To test you should manually delete course test or survey from the `exams` database

PR #6075, opened 2020-04-03
Community Leader: Add/Change title (fixes #6073)

### LINKED ISSUE #6073: Community Leader: Title when not added should be labelled as "Add title"
right now at the time of add or update , label always showing as "Change Title"

it could be made similar to Team's member, Add role/Change Role
Screenshot (17)
Screenshot (18)

robinGiri:

### DIFF
--- a/src/app/teams/teams-member.component.html
+++ b/src/app/teams/teams-member.component.html
@@ -27,7 +27,7 @@
     <button *ngIf="actionMenu.indexOf('title') > -1"
       (click)="openDialog({ member: member, change: 'title' })" mat-menu-item>
       <mat-icon>edit</mat-icon>
-      <span i18n>{ memberType, select, community {Change Title} other {{ hasRole, select, true {Change Role} false {Add Role}}} }</span>
+      <span i18n>{ memberType, select, community {{ leadershipTitleType, select, Add {Add Title} Change {Change Title}}} other {{ hasRole, select, true {Change Role} false {Add Role}}} }</span>
     </button>
   </mat-menu>
 </mat-card-header>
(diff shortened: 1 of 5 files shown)


src/app/teams/teams-member.component.html
```diff
@@ -27,7 +27,7 @@
     <button *ngIf="actionMenu.indexOf('title') > -1"
       (click)="openDialog({ member: member, change: 'title' })" mat-menu-item>
       <mat-icon>edit</mat-icon>
-      <span i18n>{ memberType, select, community {Change Title} other {{ hasRole, select, true {Change Role} false {Add Role}}} }</span>
+      <span i18n>{ memberType, select, community {{ leadershipTitleType, select, Add {Add Title} Change {Change Title}}} other {{ hasRole, select, true {Change Role} false {Add Role}}} }</span>
```
paulbert, line 30: The syntax for `i18n` is always single brackets so `{{ leadershipTitleType` should be `{ leadershipTitleType`, for example.

src/app/teams/teams-member.component.ts
```diff
@@ -34,6 +35,7 @@ export class TeamsMemberComponent implements OnInit {
 
   ngOnInit() {
     this.memberType = this.member.teamId === undefined ? 'community' : 'other';
+    this.leadershipTitleType = this.leadershipTitleType === undefined ? 'Add' : this.leadershipTitleType === '' ? 'Add' : 'Change';
```
paulbert, line 38: This only will set the text when the component initializes.  You should use `ngOnChanges` so the text updates after a title is added.

Also, it is cleaner to use a different property local to the component for the `'Add'`/`'Change'` value.  If another one of the input properties changes, the text would change from **Add** to **Change** because the value of `leadershipTitleType` is now `'Add'`.

PR #6080, opened 2020-04-06
Update myPlanet version

leomaxi:


paulbert:
Adding changes to the Planet version PR rather than merging separately.

PR #6095, opened 2020-04-08
5957 issue fix

AlekseiKorolev:
I added code that doesn't change the original "Take Test" behavior.
**Lines 188-190.** I changed ```isComplete```, because in "Preview Test" it never changes. It activates the "Finish Test" button
**Lines 258-267 and line 274.** I changed the behavior of ```nextQuestion``` after sending the penultimate answer. Now, this is the same as "Take Test" behavior.
I tried to introduce some logic to make both of these behaviors look alike at the final stage, but I couldn't find a clear and simple way.

### DIFF
--- a/src/app/exams/exams-preview.component.ts
+++ b/src/app/exams/exams-preview.component.ts
@@ -1,9 +1,9 @@
 import { Component, Inject, ViewChild } from '@angular/core';
-import { MAT_DIALOG_DATA } from '@angular/material';
+import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material';
 
 @Component({
   template: `
-    <planet-exams-view [isDialog]="true" [questionNum]="1" [exam]="data.exam"></planet-exams-view>
+    <planet-exams-view [isDialog]="true" [questionNum]="1" [exam]="data.exam" (closePreview)="closeDialog($event)"></planet-exams-view>
     <mat-dialog-actions>
       <button color="primary" mat-raised-button mat-dialog-close i18n>Close Preview</button>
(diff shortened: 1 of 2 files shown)


AlekseiKorolev:
The codeclimate issue is my 21st function when only 20 is allowed
I will try to find a way out

singharpita:
completion message is shown when clicking on finish test.
Screenshot_2020-04-09 Planet Learning

 BUT if we click on Submit answer then both button get disabled as well answer also disappear.
Screenshot_2020-04-09 Planet Learning(1)

Not sure is that how it works?

AlekseiKorolev:
image

src/app/exams/exams-preview.component.ts
```diff
@@ -1,9 +1,9 @@
 import { Component, Inject, ViewChild } from '@angular/core';
-import { MAT_DIALOG_DATA } from '@angular/material';
+import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material';
 
 @Component({
   template: `
-    <planet-exams-view [isDialog]="true" [questionNum]="1" [exam]="data.exam"></planet-exams-view>
+    <planet-exams-view [isDialog]="true" [questionNum]="1" [exam]="data.exam" (closePreview)="closeDialog($event)"></planet-exams-view>
```
paulbert, line 6: @AlekseiKorolev You need to remove `$event` for this to build correctly.  The `closeDialog` method does not take any arguments.

paulbert (changes requested):
See my comment

PR #6097, opened 2020-04-09
ReportsDetail: Full name in user name (fixes #6088)

### LINKED ISSUE #6088: ReportsDetail: Use member full name not username
### Current Behavior

Member usernames are used for the member visits table in reports.

### Desired Behavior

The full name of the member should be used (if available).  This is more human readable than the username.

image

robinGiri:


src/app/manager-dashboard/reports/reports-detail.component.ts
```diff
@@ -93,11 +95,17 @@ export class ReportsDetailComponent implements OnInit, OnDestroy {
...
-    .subscribe((loginActivities: any) => {
+    forkJoin([
+      this.usersService.getAllUsers(),
+      this.activityService.getAllActivities('login_activities', activityParams(this.planetCode, this.filter))
+    ]).subscribe(([ users, loginActivities ]: [ any[], any ]) => {
       this.loginActivities = loginActivities;
       const { byUser, byMonth } = this.activityService.groupLoginActivities(loginActivities);
       this.reports.totalMemberVisits = byUser.reduce((total, resource: any) => total + resource.count, 0);
+      byUser.map((activity) => {
+        const userDoc = users.filter((user) => user.name === activity.user);
```
paulbert, line 106: You should use `find` here.  `filter` will return an array, so `find` is better when you are looking for a single document.

Also check to make sure `user.planetCode === this.planetCode` because users on different planets can have the same username.

Issue #6055, opened 2020-03-31
ReportsCourses: New tab in report detail


paulbert:
Let's add another tab for reporting with a table for courses.

Maybe we can start with:

1. Title
2. Views
3. Enrollments
4. Completions
5. Average Rating

for columns in the table

PR #6112, opened 2020-04-13
[#6055] Course progress report

singharpita:


src/app/manager-dashboard/reports/reports-detail-activities.component.ts
```diff
@@ -23,17 +24,27 @@ export class ReportsDetailActivitiesComponent implements OnInit, OnChanges, Afte
...
       sortNumberOrString(this.sortingObject(item, property), property);
   }
 
   ngOnChanges() {
-    this.activities.data = this.activitiesByDoc.map(
-      activity => ({
+    this.activities.data = this.activitiesByDoc.map(activity => {
+      const progressCount = { enrolled: 0, passed: 0 };
+      /* if (this.activityType === 'courses') {
+        const progress = (this.progresses.find(p => p.doc._id === activity.courseId) || { progress: {} }).progress;
```
paulbert, line 36: 1. Use `progresses.filter` to get only the progress documents that are for that course
2. `enrolled` should be the number of documents for step 1 (steps should be unique per user)
3. `completed` (use that instead of passed) should be for the number who passed all steps.  You can use `CertificationsService.isCourseCompleted` for that

src/app/manager-dashboard/reports/reports.service.ts
```diff
@@ -219,4 +219,22 @@ export class ReportsService {
     });
   }
 
+  getCourseProgress({ planetCode, tillDate, fromMyPlanet, filterAdmin }: ActivityRequestObject = {}) {
+    return forkJoin([
```
paulbert, line 223: Use the methods from `CoursesService` instead, similar to `CertificationsViewComponent`

singharpita:
@paulbert I have tried to use courseservice listener and usersservice but not sure about the pass count. I tired to group by user and if one of step failed then mark as failed. Could you check if it ok and do needful.

lmmrssa:
needs to resolve conflict

PR #6188, opened 2020-04-23
myPlanetReports: Device online log (fixes #5572)

### LINKED ISSUE #5572: myplanet report separate device online log
Right not there is log of myplanet (tablets) which is recording how long device was online. We should separate or filter that report
Screen Shot 2020-01-18 at 12 32 41 AM

robinGiri:
issue: #5572 


src/app/manager-dashboard/reports/myplanet-table.component.ts
```diff
@@ -39,6 +42,17 @@ export class MyPlanetTableComponent implements OnInit, OnChanges, AfterViewInit
 
   ngOnChanges() {
     this.myPlanets.data = this.data;
+    if (this.usage !== undefined) {
+      let total;
+      this.usage.map(data => {
```
paulbert, line 47: This is modifying the `data.max.totalUsedTime` variable as a side effect.  This should be avoided.

`Array.map` returns an array of the same size with values based on the function provided.  You should return in the function `{ ...data, max: { ...data.max, totalUsedTime: total } }`

src/app/manager-dashboard/reports/myplanet-table.component.ts
```diff
@@ -39,6 +42,17 @@ export class MyPlanetTableComponent implements OnInit, OnChanges, AfterViewInit
...
+        data.max.usages.map((time) => {
```
paulbert, line 48: This is not a good use for `Array.map`.  You are totaling up values based on the array provided, and `Array.reduce` is much better for that.  Then you can do `const total = data.max.usages.reduce(...)`.

robinGiri:
@paulbert can you review changes once.

Issue #6194, opened 2020-04-23
Notification for new user does not show user if user profile updated


lmmrssa:
If notification has been created without user profile updated and later on updated profile, then when you click on notification for user created it will show empty record as user name has already been updated and name might be different than username

paulbert:
@lmmrssa This is an issue where an image would be helpful.  The notification for a new member links to the members list with the username in the search field (pictured below).

I think linking directly to the user profile page would be better because that link is based on the username which does not change. That route would be `/manager/users/profile/{username}`.

![image](https://user-images.githubusercontent.com/9203229/80173264-46b80680-85bd-11ea-88b3-ea3679de0041.png)

PR #6198, opened 2020-04-25
[#6194] updated user notification url

singharpita:


src/app/login/login-form.component.ts
```diff
@@ -186,8 +186,8 @@ export class LoginFormComponent {
     const data = {
       'user': 'org.couchdb.user:' + userName,
       'message': `New member <b>${addedMember}</b> has joined.`,
-      'link': '/manager/users/',
-      'linkParams': { 'search': addedMember },
+      'link': '/manager/users/profile/' + addedMember,
+      //'linkParams': { 'search': addedMember },
```
paulbert, line 190: @singharpita This line should be removed.

It also fails the lint test.  Is there an issue with `npm run install-hooks`?

Issue #6214, opened 2020-04-28
Reports: CSV does not include current day


paulbert:
Right now the CSV export for courses at least, and maybe resources, does not include the current day.

We need to fix that so the end date selected includes all activities that day.

PR #6217, opened 2020-04-29
[#6214] Fix csv export not showing last date

singharpita:


paulbert:
@singharpita Thanks for the quick PR. Please make sure to assign yourself to an issue when you make a PR so we can keep track of what's being worked on.

PR #6220, opened 2020-04-30
Health: Userdb access script (fixes #6211)

### LINKED ISSUE #6211: Health: Planet script to add health role to userdbs

paulbert:
#6211 

The script will run & get the database list, but it doesn't run anything in the `if [[ $DB =~ ^userdb- ]]; then` clause.  Will look into it later.  I have to call it a night.

### DIFF
--- a/docker/planet/default.conf
+++ b/docker/planet/default.conf
@@ -149,11 +149,11 @@ server {
   location = /healthaccess {
     include fastcgi_params;
 
-    if ($args ~ "u=(.{1,})$") {
+    if ($args ~ "p=(.{1,})$") {
       set $credentials $1;
     }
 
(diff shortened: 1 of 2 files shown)


paulbert:
Working on my raspberry pi to make a request to `/healthaccess` with `?p={pin number}`.

This is the brute force approach, and it's taking 17sec to check the 55ish user databases on my raspberry pi and update the security on the databases.

I think we could improve that by modifying the endpoint to a POST which takes the pin number and a list of users so it doesn't have to check all of their user databases.  Unfortunately the security needs to be checked on every database individually, so I think the only way to speed up is to limit the users to the ones with information on the tablet.

PR #6271, opened 2020-05-07
Fix share button (fixes #6089)

### LINKED ISSUE #6089: Teams: Share main chat post
### Current Behavior

The **Share with Community** button does not show up for the main post (outlined in gray) when viewing replies for that post.

### Desired Behavior

If it meets the other criteria (not already shared, the post above it in the chain has been shared), then the button should show.

image

Haoran-Hu:
Fixes #6089 

Share button now can be seen after showing the replies if the main post has not been shared before.

### DIFF
--- a/src/app/news/news-list.component.html
+++ b/src/app/news/news-list.component.html
@@ -9,6 +9,7 @@
     [replyObject]="replyObject"
     [showRepliesButton]="false"
     [editable]="editable"
+    [shareTarget]="shareTarget"
     (changeReplyViewing)="showReplies($event)"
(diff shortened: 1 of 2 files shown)


Haoran-Hu:
@paulbert I'm on it.

Haoran-Hu:
@paulbert Do you mean after showing the reply of a reply, the reply 1 should have a share button?

Haoran-Hu:
@paulbert Still got stuck on the way of finding out if the parent post of a post is shared.

Issue #5989, opened 2020-03-19
Create Team: remove vertical scroll bar


singharpita:
![Screenshot from 2020-03-19 19-38-03](https://user-images.githubusercontent.com/27276132/77074634-3b4f4980-6a19-11ea-95ea-5167b3d51b9c.png)

PR #6275, opened 2020-05-08
[#5989] remove vertical scrollbar

singharpita:
- vertical scroll bar is removed from all those where 'dialog-form-component' is being used.
Team,enterprise and so on


singharpita:
@paulbert 
added style based on screen size. Please review it once.

PR #6279, opened 2020-05-11
Health: Fix number validation (fixes #6263)

### LINKED ISSUE #6263: myHealth: validation should be shown for various fields
Screenshot_2020-05-07 Planet Learning
-Temperature **integer only**
- Pressure -**integer only**
- height **integer only**
-weight **integer only**

paulbert:
#6263 

In researching I found that you can also enter something just `+` in Chrome which should trigger a validation error without a number after.

Also found that this is a longstanding issue with Angular.  Added a little workaround that will automatically be in place for any `<input type="number">` fields.


paulbert:
The form validation is fixed in #6274

PR #6291, opened 2020-05-13
Reviews: Give rating by clicking in between stars(fixes #6269)

### LINKED ISSUE #6269: Area between rating stars
IMG_20200507_203157
There is area between stars of rating which is very small. But even when clicking on this area popup rating and saves rating. So it is possible to end up submitting rating as 0
Shall we remove that spacing or not trigger rating in that space

robinGiri:
Issue: #6269 


paulbert:
Reducing the margin messes up the view: the yellow bleeds into stars beyond the rating and the alignment is now off.  I reverted that change and made it so if the value is 0 nothing happens.

image

Issue #5821, opened 2020-02-20
Certification: link pointer is missing


singharpita:
which confuses weather Certification name is just label or link.
![Screenshot from 2020-02-20 20-46-18](https://user-images.githubusercontent.com/27276132/74946940-271c2a80-5422-11ea-83e2-933fd694a56a.png)

paulbert:
You can use the `cursor-pointer` style for this

PR #6292, opened 2020-05-13
Add span element with class for styling. Issue #5821

chtelmko:
Issue #5821 

Added a span element with cursor-pointer style class.

Hope this is a correct way to do a pull request. Sorry if I messed it up.


chtelmko:
Will do, thank you.

PR #6295, opened 2020-05-13
Update input validation (fixes #6284)

### LINKED ISSUE #6284: myHealth: allowing 0 in Vitals fields
since we are restricting number to be greater or equal to 0, it doesn't make sense to have value 0 in weight, height, pulse and pressure in it.

Screenshot_2020-05-12 Planet Learning(2)

jsschf:
Issue #6284 

Added validation to prevent temperature, blood pressure, height and weight from being 0.

Annotation 2020-05-13 002510

### DIFF
--- a/src/app/health/health-event.component.html
+++ b/src/app/health/health-event.component.html
@@ -10,11 +10,11 @@
       <form class="form-spacing" [formGroup]="healthForm">
         <h4 class="full-width primary-text-color">Vitals</h4>
         <mat-form-field>
-          <input matInput i18n-placeholder placeholder="Temperature in °C" formControlName="temperature" type="number" min="0">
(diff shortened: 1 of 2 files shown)


jsschf:
I'm unsure if it would make more sense to just replace the custom validator _positiveNumberValidator_ with the built in validator, so I kept both in.

jsschf:
@paulbert Will do! Do you want me to remove _postiveNumberValidator_ from _custom-validators.ts_ as well? I think that _health-event.component.ts_ was the only place it was used.

paulbert:
@jsschf It's used in a few other forms, so you can leave it.  Thanks!

jsschf:
@paulbert Is there anything else I need to do with this PR?

Issue #4873, opened 2019-09-19
Calendar event: close button is missing in View event dialog box


singharpita:
![Screenshot from 2019-09-19 21-20-17](https://user-images.githubusercontent.com/27276132/65258859-610b9c80-db23-11e9-88df-450106c8e06a.png)

paulbert:
We can add a close button in the bottom right for this view.

PR #6298, opened 2020-05-14
[#4873] calendar close button

singharpita:


paulbert:
@singharpita When there is an issue created that is a duplicate of one that you have submitted a PR for, like #6312, please notify us so we can close that new issue.

Closing this since it is fixed in #6323

Issue #6317, opened 2020-05-14
Certifications: The i18n attribute is missing


chtelmko:
Problem:
On the "add certification" page. The i18n attribute is missing on a few tags.
Line 21,22 and 44.

Steps to reproduce the problem:
Going to add Certification page

Screenshots:
![missing i18 tag](https://user-images.githubusercontent.com/40531136/81983984-30242e80-9602-11ea-9852-a6ab00127715.png)

proposed solution:

Add the attributes

PR #6318, opened 2020-05-15
Add i18n attribute. Issue #6317

chtelmko:
Added i18n attribute to the tags that did not have one.

### DIFF
--- a/src/app/manager-dashboard/certifications/certifications.component.html
+++ b/src/app/manager-dashboard/certifications/certifications.component.html
@@ -18,8 +18,8 @@
   <div *ngIf="!emptyData; else notFoundMessage" class="view-container view-full-height view-table">
     <mat-table #table [dataSource]="certifications" matSort [matSortDisableClear]="true">
       <ng-container matColumnDef="name">
-        <mat-header-cell *matHeaderCellDef mat-sort-header>Certification Name</mat-header-cell>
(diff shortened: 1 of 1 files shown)


src/app/manager-dashboard/certifications/certifications.component.html
```diff
@@ -18,8 +18,8 @@
   <div *ngIf="!emptyData; else notFoundMessage" class="view-container view-full-height view-table">
     <mat-table #table [dataSource]="certifications" matSort [matSortDisableClear]="true">
       <ng-container matColumnDef="name">
-        <mat-header-cell *matHeaderCellDef mat-sort-header>Certification Name</mat-header-cell>
-        <mat-cell *matCellDef="let element">{{element.name}}</mat-cell>
+        <mat-header-cell i18n *matHeaderCellDef mat-sort-header>Certification Name</mat-header-cell>
+        <mat-cell i18n *matCellDef="let element">{{element.name}}</mat-cell>
```
paulbert, line 22: `i18n` translation will not work for interpolation.  You should remove the one on this line.

chtelmko: Will do, should I keep it in the line 21?

paulbert: Line 21 is good.  It's just the `{{`/`}}` where the value of a component property is used for text that cannot be translated with i18n

PR #6328, opened 2020-05-16
Indicate deleted members (fixes #5945)

### LINKED ISSUE #5945: Messages/News posted by deleted members of the Community are shown on Teams/chat and on Community/News.
## Current Behavior
Messages/News posted by deleted members of the Community are shown on Teams/chat and on Community/News with reply options.

Screenshot from 2020-03-11 11-27-13

Screenshot from 2020-03-11 11-26-46

## Desired Behavior
Messages/News posted by deleted members should not be shown or it should show members as deleted and with no reply options.

Haoran-Hu:
fixes #5945 

Try to indicate deleted members on Community

WORK IN PROGRESS


src/app/news/news-list.component.ts
```diff
@@ -52,6 +55,8 @@ export class NewsListComponent implements OnChanges {
     if (this.replyViewing._id !== 'root') {
       this.replyViewing = this.items.find(item => item._id === this.replyViewing._id);
     }
+    this.usersService.requestUsers();
+    this.activeMembers = this.usersService.data;
```
paulbert, line 59: Two things
- `requestUsers` starts a request to the database which will not complete when line 59 runs.  You need to subscribe to `usersListener` to get the updated data based on the database request.
- You should use the array of users.  You don't need `loginActivities` or `childUsers`.

src/app/news/news-list-item.component.html
```diff
@@ -7,6 +7,7 @@
           item.doc.user.firstName + ((' ' + item.doc.user.middleName) || '') + ' ' + item.doc.user.lastName :
           item.doc.user.name}}
       </a>
+      <span *ngIf=!isActiveMember>(no longer member)</span>
```
paulbert, line 10: You should change this to `*ngIf="notMember"`.  It reads a little better than having `!` and having to mentally flip the boolean.

src/app/news/news-list-item.component.ts
```diff
@@ -47,6 +49,7 @@ export class NewsListItemComponent implements OnChanges, AfterViewChecked {
   ngOnChanges() {
     this.targetLocalPlanet = this.shareTarget === this.stateService.configuration.planetType;
     this.showShare = this.shouldShowShare();
+    this.isActiveMember = !this.shouldShowIndication();
```
paulbert, line 52: No need to make this a `!` of the function, especially considering it is then reversed again in the template (see my other comment, you should change that as well).

Generally if you are reversing a boolean you should take a minute to think of ways you could just flip your condition.  If not that's fine, but in this case since the function and the value are only used in one place you should be able to.

src/app/news/news-list-item.component.ts
```diff
@@ -19,6 +19,7 @@ export class NewsListItemComponent implements OnChanges, AfterViewChecked {
   @Input() isMainPostShared = true;
   @Input() showRepliesButton = true;
   @Input() editable = true;
+  @Input() activeMembers: any = {};
```
paulbert, line 22: `activeMembers` should be an array of users, not the `usersService.data`

src/app/news/news-list-item.component.ts
```diff
@@ -131,4 +134,12 @@ export class NewsListItemComponent implements OnChanges, AfterViewChecked {
       (!this.targetLocalPlanet || (!this.newsService.postSharedWithCommunity(this.item) && this.isMainPostShared));
   }
 
+  shouldShowIndication() {
+    if (this.shareTarget === this.planetCode) {
```
paulbert, line 138: Two things:
- Rather than `shareTarget` you should compare `this.planetCode` to `this.item.doc.user.planetCode`.
- The condition in the if can be combined with the `return` statement inside to make this a one line function that reads a little better.

src/app/news/news-list-item.component.ts
```diff
@@ -131,4 +134,12 @@ export class NewsListItemComponent implements OnChanges, AfterViewChecked {
...
+      return this.activeMembers.user.findIndex(this.item.doc.user._id) !== -1 &&
```
paulbert, line 139: This isn't quite it.  You don't need to look at the `loginActivities`, those are records of each login for the members of the community.

You need to check each user and see if the `_id` and the `joinDate` match the user on the news document.

paulbert (changes requested):
Good start, but see my comments for completing this.  Thanks!

src/app/news/news-list.component.ts
Haoran-Hu: @paulbert still unsure where to call `requestUsers`

Issue #6210, opened 2020-04-28
submission search


lmmrssa:
make it possible to search by person
search name of courses/survey similar to resource by words

lmmrssa:
Lets just search by words for now as there is another issue for search by user

PR #6335, opened 2020-05-18
Added FilterSpecificFieldsByWord function to setupTable to make it possible to search for separate words

chtelmko:
…. Issue #6210

Added two buttons to the toolbar header to allow to switch between searching by person(user) or by title(name).

planet with user and name search

This might not be what is needed or correct, but this is the only way I could think of making it work. Because, the Resource search uses data from the db and the submissions search has a mat-table-data source.

This is not the final commit that should be merged. I just wanted to push my changes, so that others would be able to look them over.

### DIFF
--- a/src/app/submissions/submissions.component.ts
+++ b/src/app/submissions/submissions.component.ts
@@ -75,25 +75,25 @@ export class SubmissionsComponent implements OnInit, AfterViewChecked, OnDestroy
     }
     this.coursesService.requestCourses();
     zip(this.submissionsService.submissionsUpdated$, this.coursesService.coursesListener$()).pipe(takeUntil(this.onDestroy$))
-      .subscribe(([ submissions, courses ]) => {
-        submissions = submissions.filter(data => data.user && data.type !== 'photo' && data.parent).reduce((sList, s1) => {
-          const sIndex = sList.findIndex(s => (s.parentId === s1.parentId && s.user._id === s1.user._id && s1.type === 'survey'));
-          if (!s1.user._id || sIndex === -1) {
-            sList.push(s1);
-          } else if ((s1.parent.updatedDate || 0) > (sList[sIndex].parent.updatedDate || 0)) {
-            sList[sIndex] = s1;
-          }
-          return sList;
-        }, []).map(submission => this.appendCourseInfo(submission, courses));
-        // Sort in descending lastUpdateTime order, so the recent submission can be shown on the top
-        submissions.sort((a, b) => b.lastUpdateTime - a.lastUpdateTime);
-        this.submissions.data = submissions.map(submission => ({
-          ...submission, submittedBy: this.submissionsService.submissionName(submission.user)
-        }));
-        this.dialogsLoadingService.stop();
-        this.applyFilter('');
-        this.emptyData = !this.submissions.filteredData.length;
-      });
+    .subscribe(([ submissions, courses ]) => {
+      submissions = submissions.filter(data => data.user && data.type !== 'photo' && data.parent).reduce((sList, s1) => {
+        const sIndex = sList.findIndex(s => (s.parentId === s1.parentId && s.user._id === s1.user._id && s1.type === 'survey'));
+        if (!s1.user._id || sIndex === -1) {
+          sList.push(s1);
+        } else if ((s1.parent.updatedDate || 0) > (sList[sIndex].parent.updatedDate || 0)) {
+          sList[sIndex] = s1;
+        }
+        return sList;
+      }, []).map(submission => this.appendCourseInfo(submission, courses));
+      // Sort in descending lastUpdateTime order, so the recent submission can be shown on the top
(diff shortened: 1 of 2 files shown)


lmmrssa:
Also do not make changes to lines which are not required to change

chtelmko:
I added the space in those lines because it was giving me errors and didn't want to push my branch otherwise. It was saying tslint failed.

chtelmko:
What should I use as an example for the search by word

chtelmko:
Could you help me with where to look for an example for the word search. I looked over the resources in the library and the courses. To me, they behave the same the only difference is that they use the db for their data instead of mat table data source.

chtelmko:
@lmmrssa: Could you clarify what you mean by word search please?

Currently the search runs via a key-up event as soon as the user types one letter. 

Which of the following solutions should be implemented?

1. A debounce so the search doesn't occur until a certain time elapses between user inputs?
2. A length requirement so that the search doesn't happen until the user types in a certain number of characters? (This may be confusing to users without specifying the length requirement).
3. A dropdown with a concrete list of options (if so, where do those options come from?)
4. Another solution I haven't listed?


Thank you

src/app/submissions/submissions.component.ts
```diff
@@ -126,7 +129,11 @@ export class SubmissionsComponent implements OnInit, AfterViewChecked, OnDestroy
   }
 
   setupTable() {
-    this.submissions.filterPredicate = composeFilterFunctions([ filterDropdowns(this.filter), filterSpecificFields([ 'parent.name' ]) ]);
+    this.submissions.filterPredicate = composeFilterFunctions(
```
paulbert, line 132: Indentation here should look like the following:
```
compostFilterFunctions([
  item1,
  item2
]);
```

src/app/submissions/submissions.component.ts
```diff
@@ -1,6 +1,8 @@
 import { Component, OnInit, ViewChild, AfterViewChecked, OnDestroy, Input, Output, EventEmitter } from '@angular/core';
 import { MatPaginator, MatTableDataSource, MatSort, MatDialog } from '@angular/material';
-import { filterSpecificFields, composeFilterFunctions, filterDropdowns, dropdownsFill } from '../shared/table-helpers';
+import { filterSpecificFields, composeFilterFunctions, filterDropdowns, dropdownsFill, filterTags, filterAdvancedSearch,
+        filterShelf, filterSpecificFieldsByWord } from '../shared/table-helpers';
+
```
paulbert, line 5: Please remove the extra line

src/app/submissions/submissions.component.ts
```diff
@@ -10,6 +12,7 @@ import { findDocuments } from '../shared/mangoQueries';
 import { DialogsLoadingService } from '../shared/dialogs/dialogs-loading.service';
 import { CoursesService } from '../courses/courses.service';
 
+
```
paulbert, line 15: Please remove the extra line

src/app/submissions/submissions.component.ts
```diff
@@ -126,7 +129,11 @@ export class SubmissionsComponent implements OnInit, AfterViewChecked, OnDestroy
   }
 
   setupTable() {
-    this.submissions.filterPredicate = composeFilterFunctions([ filterDropdowns(this.filter), filterSpecificFields([ 'parent.name' ]) ]);
+    this.submissions.filterPredicate = composeFilterFunctions(
+      [
+        filterSpecificFieldsByWord([ 'parent.name' ]),
+        filterDropdowns(this.filter), filterSpecificFields([ 'parent.name' ])
```
paulbert, line 135: You need to remove `filterSpecificFields`.  It's conflicting with `filterSpecificFieldsByWord`.

src/app/submissions/submissions.component.ts
```diff
@@ -203,4 +210,5 @@ export class SubmissionsComponent implements OnInit, AfterViewChecked, OnDestroy
     return { ...submission, courseTitle: submissionCourse.doc.courseTitle, stepNum };
   }
 
+
```
paulbert, line 213: Please remove the extra line

src/app/submissions/submissions.component.ts
```diff
@@ -1,6 +1,8 @@
 import { Component, OnInit, ViewChild, AfterViewChecked, OnDestroy, Input, Output, EventEmitter } from '@angular/core';
 import { MatPaginator, MatTableDataSource, MatSort, MatDialog } from '@angular/material';
-import { filterSpecificFields, composeFilterFunctions, filterDropdowns, dropdownsFill } from '../shared/table-helpers';
+import { filterSpecificFields, composeFilterFunctions, filterDropdowns, dropdownsFill, filterTags, filterAdvancedSearch,
```
paulbert, line 3: Our preferred multiline `import` style is:
```
import {
  thing1, thing2, thing3,
  thing4, thing5
} from 'things';
```

PR #6352, opened 2020-05-20
Survey: Notification on update (fixes #6191)

### LINKED ISSUE #6191: Survey Submission: on update of survey, it should show updated notification message
- create a survey and send that survey to any member as admin
- fill up the survey by that member.
- update the survey by that member again.
- login to admin and check notification
- It shows "[user] has completed the survey" even though member user has updated the survey.

Since once the survey is filled up, it can only be modified so notification should be also shown as
"[user] has updated the survey".
Screenshot_2020-04-23 Planet Learning(3)

robinGiri:
Issue: #6191 

### DIFF
--- a/src/app/submissions/submissions.service.ts
+++ b/src/app/submissions/submissions.service.ts
@@ -239,7 +239,7 @@ export class SubmissionsService {
   }
 
   sendSubmissionNotification(isRecorded: boolean) {
-    const isUpdated = this.userService.get().isUserAdmin;
+    const isUpdated = this.submission.status === "complete";
     const data = {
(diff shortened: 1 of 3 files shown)


src/app/submissions/submissions.service.ts
```diff
@@ -236,9 +236,10 @@ export class SubmissionsService {
   }
 
   sendSubmissionNotification(isRecorded: boolean) {
+    const isUpdated = this.userService.get().isUserAdmin;
```
lmmrssa, line 239: isupdated is not controlled by user being admin

PR #6356, opened 2020-05-21
Test input validation (fixes #6334)

### LINKED ISSUE #6334: HealthEvent: Allow decimals
I think the built in Angular number validator maybe defaults to 0 decimal places.  If it allows a value for precision, let's go with 2.

jsschf:
According to the [Angular Docs](https://angular.io/api/core/ElementRef), the `elementRef.nativeElement` used in _planet-number-validator.directive.ts_ is a potential security risk. 

My interpretation of the purpose of the directive was ensure that any input field with "type=number" was valid, and as a workaround due to the strange behavior between the TS _number_ type and JS _float_ type.

My current fix is to use the `pattern` HTML input attribute and a regular expression to allow up to two decimal places. If input that doesn't match the pattern, then the `ng-invalid` and `aria-invalid=true` attributes are active and prevents the form from submitting.

#### Questions/Thoughts
1. To allow the HTML pattern to work, I had to comment out the entire _planet-number-validator.directive.ts_ file and remove it from _planet-forms.module.ts_. Is method OK or do I need to find a fix while using the directive?
1. I only added the attribute to the `temperature` input field as a proof-of-concept. If you are OK with my implementation I can add the attribute to the rest of the input fields.
1. Another implementation that does the same thing is add the `Validators.pattern()` method in the Form Control of the TS file. However, that wasn't passing TSLint when I tried to push my branch.
1. The current error message for failed pattern validation says `Invalid input. Hover for more info`, and the tooltip that pops up says `Letters, numbers _ . - allowed.` To make it more specific, maybe we could change the placeholder in the input fields in _health-event.component.ts_ to show what the expected input is. Or we could add another custom error message in _form-error-messages.component.ts_
1. My implementation works with the expected range of values added in the latest version.
1. A potential problem is that this `pattern` attribute will have to be manually added to each input field that needs the two decimal precision, rather than have it automatically active on every `input type=number` field because of the _planet-number-validator.directive.ts_

### DIFF
--- a/src/app/health/health-event.component.html
+++ b/src/app/health/health-event.component.html
@@ -10,7 +10,7 @@
       <form class="form-spacing" [formGroup]="healthForm">
         <h4 class="full-width primary-text-color">Vitals</h4>
         <mat-form-field>
-          <input matInput i18n-placeholder placeholder="Temperature in °C" formControlName="temperature" type="number" min="1">
+          <input matInput i18n-placeholder placeholder="Temperature in °C" formControlName="temperature" type="number" pattern="[0-9]+\.?[0-9]{0,2}?" min="1">
           <mat-error><planet-form-error-messages [control]="healthForm.controls.temperature"></planet-form-error-messages></mat-error>
         </mat-form-field>
         <mat-form-field>
--- a/src/app/health/health-event.component.html
+++ b/src/app/health/health-event.component.html
@@ -10,7 +10,7 @@
       <form class="form-spacing" [formGroup]="healthForm">
         <h4 class="full-width primary-text-color">Vitals</h4>
         <mat-form-field>
-          <input matInput i18n-placeholder placeholder="Temperature in °C" formControlName="temperature" type="number" pattern="[0-9]+\.?[0-9]{0,2}?" min="1">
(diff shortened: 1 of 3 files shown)


jsschf:
I apologize for my lengthy write up and questions, I wasn't sure if my thoughts on the fix were the same as what was intended

paulbert:
@jsschf The security risk from Angular is valid, but not in this case.  The worry is if you allow a user to write to the DOM you open yourself up to XSS, but we are only reading from it.

The fix in **planet-number-validator.directive.ts** is because any non numerical value, including an empty value, becomes `null` for `input[type="number"]` in Angular.  This makes it difficult to discern an empty value from an invalid value.

The issue that we were seeing is that the native HTML `input[type="number"]` validation would say a decimal value is invalid.  Looks like adding a [step](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#step) attribute would solve that.

Looking at the detail let's set step values to:

Temperature: `0.1`
Pulse: nothing, no need for decimal places
Height: `0.1`
Weight: `0.1`

jsschf:
@paulbert Ok so I don't need to worry about the regex or _planet-number-validator.directive.ts_, just adding a `step` attribute to the HTML input fields is enough?

jsschf:
@paulbert I undid everything I did last commit and added the step attribute to the three inputs.

Issue #6363, opened 2020-05-21
ExamsAdd: Test preview multiple choice retains value


lmmrssa:
On test preview all answers are cleaned once page is navigated between questions. But checkbox remains and requires to uncheck and check again for Submit Answer button to be enabled

lmmrssa:
This is just happening when you open preview test while creating test. Looks to be ok when you are on course view and click preview test.

PR #6384, opened 2020-05-25
[#6363] test preview multiple choice value fixes

singharpita:
issue #6363


lmmrssa:
@singharpita changes looks working but it will be good to change setexampreview function and reset there checkboxstate instead of change you made

singharpita:
@lmmrssa made changes based on above comments.
Please review it once.

PR #6386, opened 2020-05-25
Remove manager completely

lmmrssa:


lmmrssa:
Requires all old managers to be demoted first

PR #6403, opened 2020-05-26
disable show all time button (fixes #6396)

### LINKED ISSUE #6396: ReportsDetail: Show All Time active on initialization
### Current Behavior

When the **ReportsDetailComponent** initializes the **Show All Time** button is active, despite the fact the date range is at the minimum start and maximum end date.

If you click the button with or without changing the dates it becomes disabled.

### Desired Behavior

The **Show All Time** button should be disabled when this page loads.

https://www.loom.com/share/2d24616cb3ed42c9aa818e7ca884cb10

Haoran-Hu:
fixes #6396 

The **Show All Time** button is disabled when the page loads.


Haoran-Hu:
@paulbert `this.today` is of type **Date**, `setHours` would return type of **number**, so it might cause some new issues when doing assignment on line 86. Instead, I use `setHours` on `value.endDate` on line 129 since it is of type any. I think it may be a good approach.

Issue #6395, opened 2020-05-26
ReportsDetail: Export dialog range same as report selection


paulbert:
### Current Behavior

The form to select dates for exporting data always initializes to the lifetime of the Planet

### Desired Behavior

The form should initialize to the same value as what is selected for the reports.

https://www.loom.com/share/fadc76b4bfb245fdae090cc3ce7e0d86

PR #6440, opened 2020-05-29
 Start date value in reports. Issue #6395

chtelmko:
Issue #6395

Changed where the start date value is coming from, for exporting reports.

### DIFF
--- a/src/app/manager-dashboard/reports/reports-detail.component.ts
+++ b/src/app/manager-dashboard/reports/reports-detail.component.ts
@@ -271,8 +271,6 @@ export class ReportsDetailComponent implements OnInit, OnDestroy {
 
   openExportDialog(reportType: 'logins' | 'resourceViews' | 'courseViews' | 'summary') {
     const minDate = new Date(this.activityService.minTime(this.loginActivities.data, 'loginTime')).setHours(0, 0, 0, 0);
-    const startDateFromForm: Date = this.dateFilterForm.controls.startDate.value;
-    const endDateFromForm: Date = this.dateFilterForm.controls.endDate.value;
     const commonProps = { 'type': 'date', 'required': true, 'min': new Date(minDate), 'max': new Date(this.today) };
     const fields = [
       {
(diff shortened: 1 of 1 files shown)


src/app/manager-dashboard/reports/reports-detail.component.ts
```diff
@@ -270,7 +270,13 @@ export class ReportsDetailComponent implements OnInit, OnDestroy {
   }
 
   openExportDialog(reportType: 'logins' | 'resourceViews' | 'courseViews' | 'summary') {
-    const minDate = new Date(this.activityService.minTime(this.loginActivities.data, 'loginTime')).setHours(0, 0, 0, 0);
+    let minDate = new Date(this.activityService.minTime(this.loginActivities.data, 'loginTime')).setHours(0, 0, 0, 0);
+
+    const startDateFromForm: Date = new Date(this.dateFilterForm.controls.startDate.value);
+    if (startDateFromForm !== this.minDate) {
```
paulbert, line 276: This statement will always be `true`.  An Object in JavaScript is always a reference, and when you create an object (with `new` or `const object = { ... }`) it creates a new reference.  When you compare an object with `===`, `!==`, `==`, or `!=` only the references are compared, not the values.

Regardless, you do not need to check if the form value is the same as minDate.  Either way you should set the values to the form value.

chtelmko: Thank you, for your review. I have made the changes.

src/app/manager-dashboard/reports/reports-detail.component.ts
```diff
@@ -285,8 +287,8 @@ export class ReportsDetailComponent implements OnInit, OnDestroy {
       }
     ];
     const formGroup = {
-      startDate: new Date(minDate),
-      endDate: [ new Date(this.today), CustomValidators.endDateValidator() ]
+      startDate: new Date(startDateFromForm.setHours(0, 0, 0, 0)),
```
paulbert, line 290: You should remove `setHours` and you do not need to create a `new Date` here.  This also applies to `endDate`

PR #6443, opened 2020-06-01
Version 0.11.8

paulbert:


docker/planet/scripts/create_version_json.sh
```diff
@@ -3,7 +3,7 @@ LATEST_APK_VERSION="v0.7.88"
 LATEST_APK_VER_CODE="788"
 MIN_APK_VERSION="v0.7.82"
 MIN_APK_VER_CODE="782"
-APK_PATH="https://github.com/open-learning-exchange/myplanet/releases/download/v0.7.82/myPlanet.apk"
+APK_PATH="https://github.com/open-learning-exchange/myplanet/releases/download/v0.7.88/myPlanet.apk"
```
dogi, line 6: hmm ... thought this needs to get changed to max every time ...

paulbert: @dogi Yes it was not changed in the last version, so I changed it to make sure even if there isn't a new myPlanet version it is correct now.

PR #6445, opened 2020-06-01
Pass the examType Reference to the the setExamPreview function. (Fixes #5955)

### LINKED ISSUE #5955: Preview survey shows "Finish test" button
it should be "Finish Survey"
Screenshot from 2020-03-12 19-39-18

Mutugiii:
Fixes #5955 

Pass the examType Reference to the the setExamPreview function.

### DIFF
--- a/src/app/exams/exams-add.component.ts
+++ b/src/app/exams/exams-add.component.ts
@@ -216,7 +216,7 @@ export class ExamsAddComponent implements OnInit {
       this.showErrorMessage();
       return;
     }
-    this.examType = params
+    this.examType = params;
     this.dialog.open(ExamsPreviewComponent, { data: { exam: this.examForm.value, examType: this.examType }, minWidth: '75vw' });
   }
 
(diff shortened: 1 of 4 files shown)


src/app/exams/exams-add.component.ts
```diff
@@ -216,7 +216,8 @@ export class ExamsAddComponent implements OnInit {
       this.showErrorMessage();
       return;
     }
-    this.dialog.open(ExamsPreviewComponent, { data: { exam: this.examForm.value }, minWidth: '75vw' });
+    this.examType = params;
```
paulbert, line 219: Rather than pass a parameter to the function and modify the component property, simply pass the component property to the dialog.

This confuses things.  The only time `showPreviewDialog` is called `this.examType` is the parameter so this assignment is redundant.

Mutugiii: Thanks for that! I hadn't looked at it in that way, I believe it is ok now

PR #6463, opened 2020-06-02
Change default sorting (fixes #6441)

### LINKED ISSUE #6441: Reports: Health dates sort descending
In the table by week we should sort that data oldest first by default.

Haoran-Hu:
fixes #6441 

Change the default sorting to descending for week dates.

### DIFF
--- a/src/app/manager-dashboard/reports/reports-detail-activities.component.html
+++ b/src/app/manager-dashboard/reports/reports-detail-activities.component.html
@@ -1,10 +1,10 @@
-<mat-table #table [dataSource]="activities" [matSortDisableClear]="true" matSort>
+<mat-table #table [dataSource]="activities" [matSortDisableClear]="true" matSort [matSortActive]="matSortActive" matSortDirection="desc">
   <ng-container matColumnDef="title">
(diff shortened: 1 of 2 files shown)


Haoran-Hu:
@paulbert Do not have enough data (data not longer than 2-week span)  to test the solution right now. Feedbacks would be appreciated and I can make corresponding changes.

paulbert:
@Haoran-Hu `matSortActive` on the `mat-table` sets the initial sorting column.

Also since this table is used in a few different places, you should set this with an optional `@Input` variable (see `users-table.component.html` and `users-table.component.ts` for an example).

Haoran-Hu:
@paulbert So the default input would be the week dates?

Haoran-Hu:
@paulbert In `user-table.component.ts`, there isn't really a default value for the input maybe because an initial sorting column is not required. But for this issue, the initial sorting column is required.

paulbert:
The initial sorting is required for this use of **ReportsDetailActivitiesComponent**, but other components use it that do not require sorting.

I realize you don't need to supply any new `@Input` variables.  If the `activityType` is `health`, then the `matSortActive` should be `weekOf`, otherwise it should be undefined.

PR #6464, opened 2020-06-03
Change member table UI (fixes #6447)

### LINKED ISSUE #6447: DialogsAddTable: Member table should not include "Action"
### Current Behavior

There are three UI changes for this dialog:
1. The dialog for adding a member includes the "Action" column with one button per row to "View Profile".  The button fortunately does not perform an action, but it should be removed regardless.
2. The profile image has a pointer icon when the cursor hovers over which should not be there
3. A user can remove roles from another member from this view.

### Desired Behavior

1. The "Action" column should not be there in the dialog.
2. The cursor should remain default on the image.
3. No roles should be changed in this view.

Note that dialog is **DialogsAddTableComponent** with `mode` set to `'users'`.  That has **UsersComponent** as a child which in turn has **UsersTableComponent** as a child.

image

Haoran-Hu:
fixes #6447

1. The "Action" column should not be there in the dialog.
2. The cursor should remain default on the image.
3. No roles should be changed in this view.

### DIFF
--- a/src/app/users/users-table.component.html
+++ b/src/app/users/users-table.component.html
@@ -18,10 +18,10 @@
       <mat-header-cell *matHeaderCellDef i18n>Profile Image</mat-header-cell>
       <mat-cell *matCellDef="let element">
         <div *ngIf="element.imageSrc; else noImage">
-          <img class="profile-image cursor-pointer" [src]="element.imageSrc">
+          <img class="profile-image" [src]="element.imageSrc">
         </div>
         <ng-template #noImage>
-          <img class="profile-image cursor-pointer" src="assets/avatar.png">
+          <img class="profile-image" src="assets/avatar.png">
         </ng-template>
       </mat-cell>
     </ng-container>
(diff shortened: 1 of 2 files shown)


Haoran-Hu:
@paulbert **Health Provider** does not show up after adding a role to a member.

src/app/users/users-table.component.html
```diff
@@ -88,45 +88,15 @@
         <mat-chip-list>
           <mat-chip *ngIf="element.doc.roles.length > 0 || element.doc.isUserAdmin" i18n>Learner</mat-chip>
           <ng-container *ngFor="let role of element.doc.roles; index as i">
-            <mat-chip *ngIf="role !== 'learner'" [removable]="!element.doc.isUserAdmin" (removed)="removeRole(element.doc,i)">
+            <mat-chip *ngIf="role !== 'learner'">
```
lmmrssa, line 91: just need to add condition for `removable` no need to change `removed`

src/app/users/users-table.component.html
```diff
@@ -88,45 +88,15 @@
...
               <planet-role [role]="role"></planet-role>
-              <mat-icon matChipRemove *ngIf="isUserAdmin && !element.doc.isUserAdmin">cancel</mat-icon>
```
lmmrssa, line 93: just add condition on if

src/app/users/users-table.component.html
```diff
@@ -88,45 +88,15 @@
...
             </mat-chip>
           </ng-container>
           <mat-chip *ngIf="element.doc.isUserAdmin && element.doc.roles.length === 0">Admin</mat-chip>
         </mat-chip-list>
         <span *ngIf="!element.doc.isUserAdmin && element.doc.roles.length === 0" i18n>Inactive</span>
       </mat-cell>
     </ng-container>
-    <ng-container matColumnDef="action">
```
lmmrssa, line 101: column should still be there but should be controlled by displayedcolumns property from component

src/app/users/users-table.component.ts
```diff
@@ -41,7 +41,7 @@ export class TableState {
 })
 export class UsersTableComponent implements OnInit, OnDestroy, AfterViewInit, OnChanges {
 
-  @Input() displayedColumns = [ 'select', 'profile', 'name', 'visitCount', 'joinDate', 'lastLogin', 'roles', 'action' ];
+  @Input() displayedColumns = [ 'select', 'profile', 'name', 'visitCount', 'joinDate', 'lastLogin', 'roles' ];
```
lmmrssa, line 44: should be based on condition

Haoran-Hu: @lmmrssa 'action' is an element in `displayedColumns`, how to exclude 'action' based on condition. Is that `{action:isDialog}`

Haoran-Hu: @lmmrssa I switch the status of `displayedColumns` in `user-table.component.ts`. I think it is working.

PR #6467, opened 2020-06-03
add openssl package

lmmrssa:


docker/planet/Dockerfile
```diff
@@ -3,15 +3,16 @@ FROM nginx:1.16.0-alpine
 RUN rm -rf /usr/share/nginx/html/*
 RUN apk add --no-cache \
       fcgi=2.4.0-r8 \
-    fcgiwrap=1.1.0-r3 \
-    spawn-fcgi=1.6.4-r3 \
+      fcgiwrap=1.1.0-r3 \
```
paulbert, line 6: @dogi I believe the indentation here is for a purpose, correct?

dogi: @paulbert yes ... packages we need and there dependencies ...

PR #6470, opened 2020-06-03
[#6465] hash pin

lmmrssa:


paulbert:
@lmmrssa @dogi Do we want to allow anyone to access the PIN?

A client supplies the key, and then gets the PIN back encrypted with that key.  If I'm reading this correctly, anyone can make that request and get the PIN.  Please let me know if I'm missing something.

Also let's be more careful with the term "hash" in the code because that is one-way always.  This is a two-way encryption so seeing a variable `HASH` makes this confusing.

PR #6471, opened 2020-06-04
ReportsHealth: Add trend chart (fixes #6469)

### LINKED ISSUE #6469: Reports: Health trend chart

paulbert:
#6469 

Starts with simply COVID-19 by week.


lmmrssa:
@paulbert not being able to see all health data, it only showed graph for one health status. Merging it to test if it will show up on instance with more health record or not

PR #6482, opened 2020-06-04
myHealth: NaN is allowed in form submit (fixes #6457)

### LINKED ISSUE #6457: myHealth: NaN allows form submit
In the Height or Weight fields if you enter I think over 20 digits the warning message pops up saying "NaN", but still allows you to submit. Then on past examinations the field is empty.
Annotation 2020-06-01 211324735
Annotation 2020-06-01 214819

robinGiri:
issue: #6457
I have been experiencing few problem in this issue.
1. When I enter 17th place number digit it auto rounds.
2. When I enter 22th place number it converts into scientific notation number and I am unable to convert it into normal number.


paulbert:
If we have a use case where we need to store big numbers accurately we can look into options, but for this use case we just need to have the number not switch to something else that will cause the form to submit with an unexpected value.

PR #6527, opened 2020-06-09
Survey: Append copy (fixes #5091)

### LINKED ISSUE #5091: Surveys: Once survey taken do not allow edit only copy
EDIT: You can ignore this comment, it's not really relevant to the direction we decided on.

We will need to have a better way to handle survey versions for the CSV export.  Right now the export uses the latest version of the survey on the local database for the questions, and matches answers from the submissions based on `question.body`.  Any answers for other questions that may have been deleted or any changes to the order of the questions are ignored in the export.

One alternative would be to only export survey information that matches if the survey revision matches the one on the submission.  This is the easiest, but makes it harder to compare data if a small change was made to a survey that was resent.

Another alternative would be to allow the user to choose the revision of the survey submissions to export.

The other more complex alternative would be to export in the following format:

 User | Date | Version | Survey Title | Question | Answer 
 ------ | ----- | ----- | ----- | ----- | -----
 as is | as is | The digits of _rev before - | Title of the survey | Question Text | Answer Text

This will leave it up to the users to aggregate to see the survey answers side by side in columns.

jsschf:
Fixes #5091 

So the query works, however I'm having a strange issue where planet updates the amount of times a survey has been taken, but the `submissions` database doesn't reflect that.
Annotation 2020-06-09 150912
Annotation 2020-06-09 151011

This doesn't make sense to me because that data in **SurveysComponent** is coming from the `submissions` database. @paulbert @lmmrssa could you shed some light on what's happening here or tell me if I'm misunderstanding?

### DIFF
--- a/src/app/exams/exams-add.component.ts
+++ b/src/app/exams/exams-add.component.ts
@@ -102,9 +102,7 @@ export class ExamsAddComponent implements OnInit {
       this.examForm.controls.name.setAsyncValidators(this.nameValidator(data.name));
       this.examForm.patchValue(data);
       this.initializeQuestions(data.questions);
-    }, (error) => {
-      console.log(error);
-    });
+    }, error => console.log(error));
     if (this.examType !== 'survey') {
       return;
     }
@@ -115,9 +113,7 @@ export class ExamsAddComponent implements OnInit {
         this.examForm.patchValue({ name: this.examForm.value.name += ' - COPY' });
         this.examForm.controls.name.setAsyncValidators(this.nameValidator());
       }
-    }, (error) => {
-        console.log(error);
-    });
+    }, error => console.log(error));
     this.courseName = this.coursesService.course.form ? this.coursesService.course.form.courseTitle : '';
   }
 
--- a/src/app/exams/exams-add.component.ts
+++ b/src/app/exams/exams-add.component.ts
@@ -22,6 +22,7 @@ import { StateService } from '../shared/state.service';
 import { markdownToPlainText } from '../shared/utils';
 import { SubmissionsService } from './../submissions/submissions.service';
 import { findDocuments } from '../shared/mangoQueries';
+import { forkJoin, of } from 'rxjs';
 
 const showdown = require('showdown');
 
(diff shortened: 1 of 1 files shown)


paulbert:
I think the `taken` field is making it into the submissions document by mistake.  That's a calculated field based on the number of documents in the submissions database for that test/survey, and it's not updated for all documents when the value changes.

Generally we try to make sure calculated fields like that do not make it into the database because it is misleading, but it looks like this one made it in.  Sorry for the confusion!

jsschf:
@paulbert Ok thanks that makes sense.

jsschf:
@paulbert @lmmrssa I'm having trouble pushing my code because tslint is failing, and the debug log doesn't give specific information. Is there a way to find which line or which rule I'm breaking, or setting up an auto-format on save using the tslint.json rules?

jsschf:
Hi @paulbert @lmmrssa two things:
+ So yesterday I was working on this issue and it wasn't saving the updated name with `- COPY` appended. While I was trying to figure it out, I tried it again without changing anything and it worked. Today it doesn't work again and I'm not sure why. 
+ The copied surveys don't save the survey information, only the name and I'm not sure why either.
Annotation 2020-06-11 083806


I would appreciate guidance on this issue.

paulbert:
@jsschf When I tested I saw ** - COPY** added correctly after saving, but it was not added in the form. I also saw all of the question information was saved correctly.

I see that if you edit the name and then change it back to the previous one it will still save.  You'll need to reset the validator to no exceptions (line 101 but without the parameter).

src/app/exams/exams-add.component.ts
```diff
@@ -92,15 +95,27 @@ export class ExamsAddComponent implements OnInit {
...
+        this.submissionsService.getSubmissions(findDocuments({ 'parent._id': this.route.snapshot.paramMap.get('id') }))
+        .subscribe((data) => {
+          if (data.length !== 0) {
+            this.pageType = 'Add';
+            this.documentInfo = {};
+            this.examForm.value.name += ' - COPY';
+          }
+        }, (error) => {
+          console.log(error);
+       });
```
paulbert, line 117: Small thing, but there should be one more space in this indentation

jsschf:
@paulbert Two of the Code Climate issues are that the constructor in **ExamsAddComponent** looks similar to the constructor in **CoursesComponent**, which I don't think necessitates a fix. The other two issues are that `ngOnInit()` exceeded both the Cognitive Complexity and lines of code allowances. I'm not sure how to refactor to resolve those issues. I would appreciate guidance.

Issue #6517, opened 2020-06-09
Reports: On nation link to "View Community" if community is online


paulbert:
We have a request to add a feature on the **ReportsComponent** to add a button which opens the community in a new tab.

The button should be disabled unless a request to the community URL is successful.

![image](https://user-images.githubusercontent.com/9203229/84101435-f31e4280-a9db-11ea-8b64-e1775e499780.png)

PR #6529, opened 2020-06-10
[#6517] Community view list

singharpita:
@paulbert I have used getUrl function to check if url of planet can be accessed. Could not make sure how to trap different errors with url request. Could you review current implementation and suggest what need to be done.

### DIFF
--- a/src/app/manager-dashboard/reports/reports.component.ts
+++ b/src/app/manager-dashboard/reports/reports.component.ts
@@ -7,7 +7,7 @@ import { ManagerService } from '../manager.service';
 import { arrangePlanetsIntoHubs, attachNamesToPlanets, getDomainParams } from './reports.utils';
 import { StateService } from '../../shared/state.service';
 import { ActivatedRoute, Router } from '@angular/router';
-import { takeUntil, switchMap, catchError, of } from 'rxjs/operators';
+import { takeUntil, switchMap, catchError } from 'rxjs/operators';
 
 @Component({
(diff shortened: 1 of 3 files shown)


src/app/manager-dashboard/reports/reports.component.ts
```diff
@@ -61,12 +61,12 @@ export class ReportsComponent implements OnInit, OnDestroy {
...
         this.arrangePlanetData(planets, hubs);
-        return forkJoin([
-          this.activityService.getActivities('resource_activities', 'byPlanet', domain),
-          this.activityService.getActivities('login_activities', 'byPlanet', domain),
-          this.activityService.getAdminActivities({ domain })
-        ]);
-      })
+        return forkJoin(planets.filter((p: any) => p.docType !== 'parentName').map((p: any) =>
+          this.couchService.getUrl('db', { domain: p.localDomain, protocol: 'http', usePort: true })
+          .pipe(catchError(err => of({ [p.code] : false })), switchMap(() => of({ [p.code]: true })))
```
paulbert, line 66: This is a problem.  `catchError` will continue along the operator chain, so this means every planet is shown as online because the `switchMap` will still run after the `catchError`.

paulbert (changes requested):
The specific error isn't important.  If there was an error, the button should stay disabled.  It should only be enabled if the request is successful.

I also think you should just use `couchService.get` rather than modifying `getUrl`.

Issue #6503, opened 2020-06-08
Proper warnings needed for Finances


Haoran-Hu:
In **Community**, **Finances**, when adding transactions, there should be some warnings when the balance is negative.

### Screenshots
![Screen Shot 2020-06-08 at 6 24 23 PM](https://user-images.githubusercontent.com/57075986/84086282-855f2000-a9b5-11ea-9b90-2b9dd9f35f16.png)

paulbert:
We could add a message at the top if the balance is negative.  Should only do this when viewing all transactions.

paulbert:
Let's have the message be placed above the table, start with a `warning` icon and have the text `The current balance is negative`.

PR #6530, opened 2020-06-10
Add warning message if balance is negative. Issue #6503

chtelmko:
Issue #6503

Created a method to check for negative balance and to make sure that the user is viewing all transactions.
Added div with the warning message.


design/resources/resources-design.json
```diff
@@ -1 +1,5 @@
-{"views":{"count_tags":{"map":"function (doc) {\n        for (var i = 0; i < doc.tags.length; i++) {\n          emit(doc.tags[i], 1);\n        }\n      }","reduce":"_count"},"titles":{"map":"function (doc) {\n        emit(doc.title.toLowerCase().trim(), doc.privateFor);\n      }"}}}
\ No newline at end of file
+<<<<<<< HEAD
```
paulbert, line 1: @chtelmko Why did you change this file at all?

I don't think this should have been changed, and now there are merge conflict markers which should always be handled before committing & pushing.

I will revert this file's changes.

chtelmko: I didn't think there were anything changed on that file. I made that commit in order to create my pull request. I wasn't able to push my push my branch,It was saying I had uncommitted changes on the file. Thank you for reverting the changes.

PR #6544, opened 2020-06-11
Reports: Show/Hide empty diagnosis charts (Fixes #6507)

### LINKED ISSUE #6507: Health Report: Diagnosis trend showing chart/graph for empty value
- select a data range where no examination has been taken during the week or else Diagnosis is empty
- see the Diagnosis Trend.
- It still show graph even though there is no value for respective Diagnosis
Screenshot_2020-06-09 Planet Learning

It should be empty in case no value for Diagnosis.

Mutugiii:
Fixes #6507

### DIFF
--- a/src/app/manager-dashboard/reports/reports-health.component.html
+++ b/src/app/manager-dashboard/reports/reports-health.component.html
@@ -20,7 +20,7 @@
       </mat-select>
     </mat-form-field>
   </div>
-  <div class="chart-container">
+  <div *ngIf="showChart" class="chart-container">
     <canvas id="diagnosesTrend" #diagnosesChart></canvas>
   </div>
   <h1 class="mat-title" i18n>Weekly Data</h1>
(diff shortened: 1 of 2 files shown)


Mutugiii:
@paulbert @lmmrssa I am having a little trouble here. It is supposed to show and hide the chart. The show & hide works. The problem is once it shows the 1st time it doesn't show again. The chart space is shown but the chart itself with its data doesn't

At a point the chart data doesn''t show and it renders to html as follows:
    `<canvas _ngcontent-uys-c16="" id="diagnosesTrend"></canvas>`

While the 1st time it renders as follows(where it works):
`<canvas _ngcontent-uys-c16="" id="diagnosesTrend" width="666" height="196" class="chartjs-render-monitor" style="display: block; width: 666px; height: 196px;"></canvas>`


I have tried as is in the PR and I also tried setting the context for the showChart boolean before data is defined and it still doesn't work(as below).
oletroubleshootingreports

paulbert:
The condition is not correct.  What you are checking is if there are any cases in the first week of the date range, but it should be for the entire period.  That's stored in `this.headlineData.conditions[condition]`.

You should include this with the initial condition, `this.weeklyHealthData.length === 0` because that block already includes the change which would fix the chart data no longer initializing.

PR #6614, opened 2020-06-20
HealthEventDialog: Countdown for "Edit" (fixes #6571)

### LINKED ISSUE #6571: HealthEventDialog: Countdown for "Edit"
### Current Behavior

After #6560 is fixed, the edit button in **HealthEventDialogComponent** will only show up for 5 minutes after the last update for the examination.

### Desired Behavior

While the **Edit** button is there, let's have a countdown in parentheses in the format `m:ss` for those five minutes.

jsschf:
Fixes #6571

### DIFF
--- a/src/app/health/health-event-dialog.component.html
+++ b/src/app/health/health-event-dialog.component.html
@@ -26,5 +26,7 @@
 </mat-dialog-content>
 <mat-dialog-actions>
   <button type="button" mat-raised-button mat-dialog-close i18n>Close</button>
-  <button type="button" color="primary" mat-dialog-close (click)="editExam(event)" *ngIf="canUpdate" mat-raised-button i18n>Edit ({{ minutes }}:{{ seconds }})</button>
+  <button type="button" color="primary" mat-dialog-close (click)="editExam(event)" *ngIf="canUpdate" mat-raised-button i18n>
+    Edit <ng-container *ngIf="minutes">({{ minutes }}:{{ seconds }})</ng-container>
+  </button>
 </mat-dialog-actions>
--- a/src/app/health/health-event-dialog.component.html
+++ b/src/app/health/health-event-dialog.component.html
@@ -26,5 +26,5 @@
 </mat-dialog-content>
 <mat-dialog-actions>
   <button type="button" mat-raised-button mat-dialog-close i18n>Close</button>
-  <button type="button" color="primary" mat-dialog-close (click)="editExam(event)" *ngIf="canUpdate" mat-raised-button i18n>Edit</button>
+  <button type="button" color="primary" mat-dialog-close (click)="editExam(event)" *ngIf="canUpdate" mat-raised-button i18n>Edit ({{ minutes }}:{{ seconds }})</button>
 </mat-dialog-actions>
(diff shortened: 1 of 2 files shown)


jsschf:
@paulbert @lmmrssa The timer works although I have a few questions:
1. I put the timer itself inside the edit button. If that isn't what you intended then I can display it somewhere else.
2. For less than a second after the dialog pops up, the "minutes" and "seconds" field aren't rendered, so the display looks like `Edit (:)` before it changes to `Edit (m:ss)`. Any guidance on how to resolve this would be great.
Annotation 2020-06-22 210628

src/app/health/health-event-dialog.component.html
```diff
@@ -26,5 +26,5 @@ <h4 class="primary-text-color" *ngIf="hasConditionAndTreatment" i18n>Other Notes
 </mat-dialog-content>
 <mat-dialog-actions>
   <button type="button" mat-raised-button mat-dialog-close i18n>Close</button>
-  <button type="button" color="primary" mat-dialog-close (click)="editExam(event)" *ngIf="canUpdate" mat-raised-button i18n>Edit</button>
+  <button type="button" color="primary" mat-dialog-close (click)="editExam(event)" *ngIf="canUpdate" mat-raised-button i18n>Edit ({{ minutes }}:{{ seconds }})</button>
```
paulbert, line 29: <ng-container *ngIf="minutes">({{ minutes }}:{{ seconds }})</ng-container>

src/app/health/health-event-dialog.component.ts
```diff
@@ -46,4 +49,18 @@ export class HealthEventDialogComponent implements OnInit {
     this.router.navigate([ 'event', { id: this.data.user, eventId: event._id } ], { relativeTo: this.data.route });
   }
 
+  timer() {
+    const intervalId = setInterval(() => {
+      const time = (300000 - (new Date(Date.now()).getTime() - new Date(this.event.date).getTime())) / 1000;
```
paulbert, line 54: Since `setInterval` doesn't run immediately, you could define this function, run it first, and then run it in the `setInterval`.

src/app/health/health-event-dialog.component.ts
```diff
@@ -46,4 +49,21 @@ export class HealthEventDialogComponent implements OnInit {
...
 
+  countdown() {
+    this.timer();
+    const intervalId = setInterval(() => {
+      this.timer(intervalId);
+    }, 1000);
+  }
+
+  timer(intervalId?) {
+    const time = (10000 - (new Date(Date.now()).getTime() - new Date(this.event.date).getTime())) / 1000;
```
paulbert, line 60: @jsschf This is off.  The number at the start should be 300000 (5 minutes in ms).

paulbert: It would be best to set this as a property and replace the 300000 in the constructor as well.  That way it'll be easy to change the edit time limit.

src/app/health/health-event-dialog.component.ts
```diff
@@ -46,4 +49,21 @@ export class HealthEventDialogComponent implements OnInit {
...
+    if (time <= 0) {
+      clearInterval(intervalId);
```
paulbert, line 62: When the time hits 0 you should set `canUpdate` to `false`, also, so the button disappears.

jsschf:
@paulbert Sorry about the time in ms, I lowered it so I wouldn't have to wait 5 min each time during testing and forgot to change it back. Everything is updated now.

PR #6654, opened 2020-06-25
6523issue fix

AlekseiKorolev:
Changed line 12 to ``element.last_synced || element.time`` instead of  ``element.time || element.last_synced`` because elements sorting by last_synced.
One document doesn't have last_synced key and this is why it doesn't sort 
image


Mutugiii:
Stale

PR #6714, opened 2020-07-01
6697 course team progress

singharpita:


src/app/courses/progress-courses/courses-progress-leader.component.ts
```diff
@@ -71,6 +78,21 @@ export class CoursesProgressLeaderComponent implements OnInit, OnDestroy {
     });
   }
 
+  getTeams(course) {
+    const selectors = {
+      teamPlanetCode: this.selectedPlanetCode,
+      courses: { '$elemMatch': { '_id': course._id } }
+    };
+    this.couchService.findAll('teams', findDocuments(selectors, 0))
+      .pipe(tap(teams => this.courseTeams = teams))
```
paulbert, line 87: This `tap` is unnecessary.  In fact both of the `pipes` here should just be performed in one subscribe.

src/app/courses/progress-courses/courses-progress-leader.component.ts
```diff
@@ -71,6 +78,21 @@ export class CoursesProgressLeaderComponent implements OnInit, OnDestroy {
...
+      .pipe(map(([ team ]) => {
```
paulbert, line 88: When we initialize the component we should only get the list of the teams to populate the dropdown.  When the selection changes we should check if we have the membership information and only then make the `getTeamMembers` request.

There's other problems with this block of code, also, but if you make the change above the below should be fixed, anyway.

I think you are trying to iterate over the array here, but the rxjs pipe `map` is different from `Array.map`.  Since the http request returns an array of values, this `map` will just have the `teams` array returned to it and go through once.

This only runs `getTeamMembers` for the first team, so the members are only stored for that one team.

paulbert (changes requested):
See my comments, thanks!

PR #6717, opened 2020-07-01
Enterprises: Finance not showing negative balance

### LINKED ISSUE #6690: Enterprises: warning for finances
For **Finances** section in **Enterprises**, a negative balance should trigger some warning similar to the **Finances** section in the **Community**.

### Screenshots
Screen Shot 2020-06-29 at 9 37 04 PM

Screen Shot 2020-06-29 at 9 40 58 PM

jsschf:
Fixes #6690

### DIFF
--- a/src/app/teams/teams-view-finances.component.ts
+++ b/src/app/teams/teams-view-finances.component.ts
@@ -53,9 +53,8 @@ export class TeamsViewFinancesComponent implements OnInit, OnChanges {
       if (transactions.length > 0 && transactions[0].filter !== this.filterString()) {
         transactions[0] = this.setTransactionsTable(transactions)[0];
       }
-      if (this.finances) {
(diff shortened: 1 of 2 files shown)


jsschf:
@paulbert After some testing I realized that the console error and issue with finance not showing the warning for a negative balance only happens with **public** enterprises. The "property of length" undefined error shows up when you click on a public enterprise. I'm struggling to figure out how the finance issue is related to the enterprise being public. Any advice/guidance would be appreciated.

paulbert:
@jsschf The reason the error is occurring is because when a team is private the **TeamsViewFinancesComponent** will only initialize after the finances (along with the user's membership status) have been fetched from the database.

Good debugging!  While I would say handling being passed undefined or null for `finances` would be a good fix and addition anyway, this gives some reassurance that there isn't a more difficult issue in there.

jsschf:
@paulbert Sorry for all the small commits, I'm not sure where the codebeat issue regarding **PlanetBetaDirective** is coming from

Issue #6736, opened 2020-07-02
Courses: Routing when editing a course


Mutugiii:
### Description
When editing/updating a course, on routing back to course details, it routes to the full course list.

#### Steps to Reproduce
Go to Courses -> Select a course -> Click the edit button -> Route back

![olecourseedit](https://user-images.githubusercontent.com/48474421/86370206-17daa280-bc88-11ea-9497-1c0059651b73.png)

paulbert:
Right now the routes are 

```
/courses/update/{id}
```

and

```
/courses/view/{id}
```

To work smoothly we'd need the route from clicking the edit button in **CoursesViewComponent** to route to

```
/courses/view/{id}/update
```

and improve the handling of the `navigateBack()` method in **CoursesAddComponent** to be relative to the current route.

PR #6743, opened 2020-07-03
Change routing from updateCourse to go back to coursesViewDetail.…

chtelmko:
… Issue #6736
Routing back is not smooth.

### DIFF
--- a/src/app/courses/add-courses/courses-add.component.ts
+++ b/src/app/courses/add-courses/courses-add.component.ts
@@ -231,7 +231,7 @@ export class CoursesAddComponent implements OnInit, OnDestroy {
   }
 
   navigateBack() {
-    this.router.navigate([ `../../${this.courseId}` ], { relativeTo: this.route });
+    this.router.navigate([ '../../../' ], { relativeTo: this.route });
   }
 
   removeStep(pos) {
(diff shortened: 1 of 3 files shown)


src/app/courses/courses-router.module.ts
```diff
@@ -14,7 +14,7 @@ import { CoursesEnrollComponent } from './enroll-courses/courses-enroll.componen
 const routes: Routes = [
   { path: '', component: CoursesComponent },
   { path: 'add', component: CoursesAddComponent },
-  { path: 'update/:id', component: CoursesAddComponent },
+  { path: 'view/:id/update', component: CoursesAddComponent },
```
paulbert, line 17: Both `update/:id` and `view/:id/update` paths should exist for `CoursesAddComponent`.

src/app/courses/add-courses/courses-add.component.ts
```diff
@@ -231,7 +231,7 @@ export class CoursesAddComponent implements OnInit, OnDestroy {
   }
 
   navigateBack() {
-    this.router.navigate([ '/courses' ]);
+    this.router.navigate([ `../../${this.courseId}` ], { relativeTo: this.route });
```
paulbert, line 234: This should not rely on the `courseId` because the routing should go back to the courses list from the `update/:id` route.

PR #6750, opened 2020-07-06
ReportsDetail: Filter by team (fixes #6731)

### LINKED ISSUE #6731: ReportsDetail: Filter by team
We have a request to add filtering by team to **ReportsDetailComponent**.

There should be a selection of teams in the top toolbar next to the date selection.  The teams should all match the Planet report being looked at (`teamPlanetCode` equals the `planetCode` property for the component).

The report data will then need to be filtered so that only data for users that are members of the team is shown in the reports.

robinGiri:
issue: #6731 
@paulbert Can you review once. I am stuck in filter and don't know if i am suppose to do filter in  ReportDetailFilter component.

### DIFF
--- a/src/app/manager-dashboard/reports/reports-detail-data.ts
+++ b/src/app/manager-dashboard/reports/reports-detail-data.ts
@@ -2,7 +2,7 @@ import { filterByDate } from './reports.utils';
 
 export interface ReportDetailFilter {
   app: 'planet' | 'myplanet' | '';
-  teams: any;
+  members: any;
   startDate?: Date;
   endDate?: Date;
 }
@@ -26,10 +26,10 @@ export class ReportsDetailData {
     this.dateField = dateField;
   }
 
-  filter({ app, teams, startDate, endDate }: ReportDetailFilter) {
+  filter({ app, members, startDate, endDate }: ReportDetailFilter) {
     const isCorrectApp = item =>
       app === '' || ((app === 'myplanet') !== (item.androidId === undefined))
-      || teams.forEach(team => team.userId.split(':')[1] === item.user);
+      || members.some(team => team.userId.split(':')[1] === item.user);
     this.filteredData = filterByDate(
       this.data,
       this.dateField,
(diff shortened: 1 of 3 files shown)


src/app/manager-dashboard/reports/reports-detail.component.ts
```diff
@@ -236,6 +239,17 @@ export class ReportsDetailComponent implements OnInit, OnDestroy {
...
 
+  getTeams() {
+    this.couchService.findAll('teams', { 'selector': { 'status': 'active' } }).subscribe((teams: any) => {
+      this.teams = teams.filter((team: any) => team.teamPlanetCode === this.planetCode);
+    });
+  }
+
+  onTeamsFilterChange(filterValue) {
+    this.couchService.findAll('teams', findDocuments({ teamId: filterValue._id, docType: 'membership' }))
+      .subscribe((teams: any) => this.filter.teams = teams);
```
paulbert, line 250: You also need to call `filterData` to trigger the data change based on the filter.

src/app/manager-dashboard/reports/reports-detail-data.ts
```diff
@@ -25,8 +26,10 @@ export class ReportsDetailData {
     this.dateField = dateField;
   }
 
-  filter({ app, startDate, endDate }: ReportDetailFilter) {
-    const isCorrectApp = item => app === '' || ((app === 'myplanet') !== (item.androidId === undefined));
+  filter({ app, teams, startDate, endDate }: ReportDetailFilter) {
```
paulbert, line 29: Calling the parameters `teams` is misleading.  Please call it `members`.

src/app/manager-dashboard/reports/reports-detail-data.ts
```diff
@@ -25,8 +26,10 @@ export class ReportsDetailData {
...
+    const isCorrectApp = item =>
+      app === '' || ((app === 'myplanet') !== (item.androidId === undefined))
+      || teams.forEach(team => team.userId.split(':')[1] === item.user);
```
paulbert, line 32: `forEach` always returns undefined.  You should use `some` which will return `true` if one of the values in the array matches.

This should also match if any of the first conditions match AND the member match is `true`.

PR #6752, opened 2020-07-06
Courses: Fix Exam error messages

### LINKED ISSUE #3784: ExamsQuestion: Clean up validation messages
Validation messages are not showing up in the best way in the exam question forms.

- When adding a second question after finishing the first, the validation messages show up. This is because it's the same form with the values reset to blank, but they are still considered "touched" for the form validation so errors show up.
- The "You must add a choice..." error should never show up for text question types.
- The main "Some required fields are missing" message should only show up after trying to submit and once the errors are fixed it should go away.

jsschf:
Fixes #3784

### DIFF
--- a/src/app/exams/exams-add.component.html
+++ b/src/app/exams/exams-add.component.html
@@ -39,7 +39,7 @@
             <span i18n>Editing Question</span>{{ ' ' + (this.activeQuestionIndex + 1) + ' ' }}
             <span i18n>of</span>{{ ' ' + questions?.length }}
           </span>
-          <planet-exam-question [(question)]="question" [(showChoicesError)]="showFormError" [examType]="examType" *ngIf="activeQuestionIndex > -1"></planet-exam-question>
+          <planet-exam-question [(question)]="question" [examType]="examType" *ngIf="activeQuestionIndex > -1"></planet-exam-question>
         </div>
       </planet-step-list>
     </form>
--- a/src/app/exams/exams-add.component.html
+++ b/src/app/exams/exams-add.component.html
@@ -20,7 +20,7 @@
           <button mat-menu-item type="button" (click)="addQuestion('selectMultiple')" i18n>Multiple Choice - multiple answer</button>
         </mat-menu>
         <button mat-raised-button color="accent" type="button" (click)="showPreviewDialog()" class="margin-lr-5" i18n>Preview {examType, select, exam {Test} survey {Survey}}</button>
-        <span *ngIf="showFormError" i18n class="mat-caption warn-text-color">Some required fields are missing.</span>
+        <span *ngIf="showFormError && !examForm.valid" i18n class="mat-caption warn-text-color">Some required fields are missing.</span>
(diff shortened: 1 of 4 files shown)


jsschf:
> 
> 
> Looking at this issue again, there are now only two of the four tasks left:
> 
>     1. The main "Some required fields are missing" message should go away after all errors are fixed.
> 
>     2. The error message "You must add a choice..." does not show up.  This should only show up for multiple choice question types, never text question types.
> 
> 
> I think the first one will be much more difficult than the second, so we might need to split this into two PRs.



@paulbert The "You must add a choice" error message does pop up when a choice is added and then removed.


Annotation 2020-07-06 173246

Also, I was unable to get that error message to pop up on text questions, it only showed up on multiple choice questions. 

I think the reason the main "Some required fields are missing error" won't go away dynamically is because it's based of a boolean in **ExamsAddComponent** that only changes when the form is trying to be submitted. I'm not sure how to fix that issue. I'm not sure if you could have that variable sent to **ExamsQuestionComponent**, and then have it's value emitted on change back to **ExamsAddComponent**

jsschf:
@paulbert My most recent commit is the idea I have to get the "Some required fields are missing" in **ExamsAddComponent** to automatically go away when the choices in **ExamsQuestionComponent** are valid. 

I'm not sure why it's not working. Any guidance would be appreciated.

jsschf:
Two things:
* I moved the "You must add a choice" error message below the "Add choices" button, otherwise it was shifting the choices as it showed up or disappeared from the view.
* I'm struggling to get the "You must select the correct choice" to show up

paulbert (commented):
In addition to the note below, it would be better to keep `showFormError` local to **ExamsAddComponent** only.

Using `statusChanges` on the form in **ExamsAddComponent** would be preferred.

I think the `statusChanges` is needed for the choices, but you can have a local `showChoicesError` on the **ExamsQuestionComponent** instead.  The other condition in the `*ngIf` for the choices error can be removed because it is not working.

jsschf:
@paulbert When you are talking about keeping _showFormError_ local to **ExamsAddComponent**, and _showChoicesError_ local to **ExamsQuestionComponent**, are you saying that those two properties shouldn't be bound to each other?

Also, do I need to put `statusChanges` into `ngAfterViewChecked`? If so, I'm having trouble preventing the "Some required fields are missing error" from displaying right as the page loads.

paulbert:
@jsschf Yes.  The 2-way binding is incomplete in this implementation, anyway, but **ExamsQuestionComponent** should not be passing its show error value back up to **ExamsAddComponent** and it's not necessary to pass the value down.

`statusChanges` does not need to go in `ngAfterViewChecked` in **ExamsAddComponent**.  You can actually put it right after the `examForm` assignment in `createForm()`.

I'm seeing the error message show up immediately when adding a question.  In `statusChanges` you should reset the `showFormError` to false if the form is invalid and otherwise don't change it.  That way the message will only show up after the user tries to submit the form.

PR #6773, opened 2020-07-08
Sort members alphabetically (fixes #6754)

### LINKED ISSUE #6754: members should be sorted alphabetically
right now members in team/enterprise or anywhere is shown based on the latest added members at top.

Screenshot_2020-07-07 Planet Learning

Haoran-Hu:
Fixes #6754 

Members are sorted alphabetically except for the **Team Leader**.

### DIFF
--- a/src/app/teams/teams-view.component.ts
+++ b/src/app/teams/teams-view.component.ts
@@ -25,7 +25,7 @@ import { DialogsResourcesViewerComponent } from '../shared/dialogs/dialogs-resou
 import { CustomValidators } from '../validators/custom-validators';
 import { planetAndParentId } from '../manager-dashboard/reports/reports.utils';
 import { CoursesViewDetailDialogComponent } from '../courses/view-courses/courses-view-detail.component';
-import { memberCompare } from './teams.utils';
+import { memberCompare, compareFunc, compareFunc } from './teams.utils';
 import { UserProfileDialogComponent } from '../users/users-profile/users-profile-dialog.component';
 
 @Component({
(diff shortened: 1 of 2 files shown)


src/app/teams/teams.utils.ts
```diff
@@ -1 +1,10 @@
 export const memberCompare = (member1, member2) => member1.userId === member2.userId && member1.userPlanetCode === member2.userPlanetCode;
+
+export function compareFunc(member1, member2) {
+  const name1 = member1.userDoc.doc.lastName;
+  const name2 = member2.userDoc.doc.lastName;
```
paulbert, line 5: There are situations where `userDoc` doesn't exist, and this will throw a breaking error.  You should handle that case and return `member2.userId` instead.

src/app/teams/teams-view.component.ts
```diff
@@ -204,7 +204,7 @@ export class TeamsViewComponent implements OnInit, AfterViewChecked, OnDestroy {
       const docsWithName = docs.map(mem => ({ ...mem, name: mem.userId && mem.userId.split(':')[1], avatar: src(mem) }));
       this.leader = docsWithName.find(mem => mem.isLeader) || { userId: this.team.createdBy, userPlanetCode: this.team.teamPlanetCode };
       this.members = docsWithName.filter(mem => mem.docType === 'membership')
-        .sort((a, b) => memberCompare(a, this.leader) ? -1 : 0);
+        .sort((a, b) => memberCompare(a, this.leader) ? -1 : 0).sort(compareFunc);
```
paulbert, line 207: You should add the new logic into the original `sort` instead of adding a new `sort` function.

src/app/teams/teams.utils.ts
```diff
@@ -1 +1,10 @@
...
+  if (name1 && name2) {
+    return name1 < name2 ? -1 : 1;
```
paulbert, line 7: When sorting strings, you `localeCompare` is very useful because it automatically returns a number that works in a sort function (negative, positive, or 0).

src/app/teams/teams.utils.ts
```diff
@@ -1 +1,10 @@
 export const memberCompare = (member1, member2) => member1.userId === member2.userId && member1.userPlanetCode === member2.userPlanetCode;
+
+export function compareFunc(member1, member2) {
+  const name1 = member1.userDoc.doc.lastName;
+  const name2 = member2.userDoc.doc.lastName;
+  if (name1 && name2) {
```
paulbert, line 6: This is a problem, because if one member has a name entered and the other does not it will sort both by id.  You should have `name1` equal `.lastName` if it exists and `.userId` if not and the same for `name2`.

src/app/teams/teams.utils.ts
paulbert: `(member1.userDoc && member1.userDoc.doc.lastName)`

paulbert:
Oops forgot to attach the image
image

Haoran-Hu:
@paulbert Did not move the entire sort function into `teams.utils.ts`, adding just one more logic still seems manageable.

PR #6776, opened 2020-07-08
Feedback: Remove chip transitions

### LINKED ISSUE #6765: Feedback: Chips with no action
Annotation 2020-07-06 210714

The buttons/labels in the "Priority" and "Status" columns don't have the `cursor-pointer` class, but still have an animation when the mouse is hovered over them and an animation on click. This seems misleading because they stand out on the page and have animations but have no functionality.

The Submissions page also has these buttons/labels. I'm not sure if there are other parts of the app that have them.

jsschf:
Fixes #6765


jsschf:
@paulbert `pointer-events: none` removes everything except the transition to the new background color change when the button is selected. The `:after` removes the transition, but the color still changes on click. I'm not sure how to fix that.

paulbert:
@jsschf I'm seeing the following as the style causing the darker color on click.  Cancelling that out should do it.  Thanks!

```
.mat-standard-chip:focus::after {
    opacity: .16;
}
```

PR #6808, opened 2020-07-10
Styles: Add class to remove mat-chip styles

### LINKED ISSUE #6798: MatChip: No style change if not clickable
Expand #6776 to use a reusable class for the `mat-chip` elements which do not have a clickable action.

jsschf:
Fixes #6798

### DIFF
--- a/src/app/feedback/feedback.component.html
+++ b/src/app/feedback/feedback.component.html
@@ -51,8 +51,8 @@
       <ng-container matColumnDef="priority">
         <mat-header-cell *matHeaderCellDef mat-sort-header i18n>Priority</mat-header-cell>
         <mat-cell *matCellDef="let element">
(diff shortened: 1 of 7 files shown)


jsschf:
I didn't add the class where `mat-chip` has a functionality.

However, I was unable to add the class to `mat-chip` where the chip has a remove icon, because it prevents the icon from being clickable.

The following are where I was unable to add the class:
 * **NewsListItemComponent**
    * The labels: _Help Wanted_, _Offer_, and _Request for Advice_ 
* **UsersTableComponent** 
    * The roles: _Health Provider_, _Leader_, and _Monitor_

jsschf:
@paulbert I was unable to ngClass because the condition was evaluating to true, so the class was being applied anyway.

Issue #6799, opened 2020-07-09
Certificate: Edit cretificate to add more courses


chtelmko:
There is a weird behavior when editing the certificate. when certificate already has courses, it won't let you add anymore. 
Some of my certificates work fine, not sure if its a glitch with this particular certificate.

https://www.loom.com/share/4c3c1876a0724f899bad6bbcdb8c0a40

paulbert:
@chtelmko This is not the process for adding courses to a certificate in the video.

When you select courses that are already added to the certificate, the possible action is to remove them.  Clicking **Add Courses** brings up a dialog with any courses not already added to the certificate.  If all courses are selected then there will not be any in the dialog.

I think we can make a change to disable the **Add Courses** button if all of the courses in the system have already been added to the certificate.

PR #6841, opened 2020-07-16
use deepEqual to check if All couses are selected. Issue #6799

chtelmko:
Issue #6799
need to find out how to compare all of the courses from the database. how they are stored.

### DIFF
--- a/src/app/manager-dashboard/certifications/certifications-add.component.ts
+++ b/src/app/manager-dashboard/certifications/certifications-add.component.ts
@@ -60,7 +60,9 @@ export class CertificationsAddComponent implements OnInit, AfterViewChecked {
             hasAllData = false;
           }
           if (hasAllData) {
-            const allCourseIds = this.courseTable.courses.data.map(v => v.id);
+            const allCourseIds = this.courseTable.courses.data.map((v: any) => {
+              return v.id; }
+            );
             this.enableAddCourses = deepEqual(allCourseIds, this.courseIds);
           }
           this.pageType = 'Update';
--- a/src/app/manager-dashboard/certifications/certifications-add.component.ts
+++ b/src/app/manager-dashboard/certifications/certifications-add.component.ts
@@ -9,6 +9,7 @@ import { CoursesComponent } from '../../courses/courses.component';
 import { showFormErrors } from '../../shared/table-helpers';
 import { ValidatorService } from '../../validators/validator.service';
 import { PlanetMessageService } from '../../shared/planet-message.service';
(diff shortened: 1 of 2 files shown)


src/app/manager-dashboard/certifications/certifications-add.component.ts
```diff
@@ -65,7 +70,6 @@ export class CertificationsAddComponent implements OnInit, AfterViewChecked {
       this.cdRef.detectChanges();
     }
   }
-
```
paulbert, line 68: Please put this line back in and be careful to not remove lines.

src/app/manager-dashboard/certifications/certifications-add.component.ts
```diff
@@ -49,6 +51,9 @@ export class CertificationsAddComponent implements OnInit, AfterViewChecked {
           this.certificateForm.patchValue(certification);
           this.certificateInfo._rev = certification._rev;
           this.courseIds = certification.courseIds || [];
+          // here find out all courses in the system. compare all courses ids with this.coursesIds
```
paulbert, line 54: CoursesComponent already fetches all of the course system so all courses are available in `this.coursesTable.dataTable.data`

src/app/manager-dashboard/certifications/certifications-add.component.ts
```diff
@@ -49,6 +51,9 @@ export class CertificationsAddComponent implements OnInit, AfterViewChecked {
...
+          const allCourseIds = []; // find out hhow to get values for this array.
+          this.enableAddCourses = deepEqual(allCourseIds, this.courseIds );
```
paulbert, line 56: There is an extra space before the last parenthesis that should be removed.

paulbert (changes requested):
@chtelmko See my comments, thanks!

src/app/manager-dashboard/certifications/certifications-add.component.ts
chtelmko: I am having a problem with my code, where the value of "hasAllData" variable is always "undefined" and the "if"  statement does not get executed(line 62-65).
After debugging I noticed that 
1. this.courseTable does not have a dataTable.data property ( used this.courseTable.courses.data, instead)
2. The first time you click on a course and click the edit button, the this.courseTable.courses.data is undefined. If you return back to certificate list and click on the same course again, this.courseTable.courses.data has 23 entries.
3. The this.courseTable.courses.data must be local database data, feel like there should be more than 23 entries.
not sure how to go forward, what to try next to fix this issue. Any help would be appreciated.

Issue #6852, opened 2020-07-16
Courses: Duplicate images in dialog


jsschf:
If you add the same images with the same title from your filesystem multiply times, each one gets saved in the dialog.

![Annotation 2020-07-16 090222](https://user-images.githubusercontent.com/54686871/87680676-3dc67380-c76d-11ea-8e19-04fad7563d41.png)

paulbert:
Short term quick fix: Reject images with the same filename as one already created and direct the user to rename the file.

Longer term fix: If the filename and the image match what's on the database, just use the existing image.  If it does not add it with a (2), etc added at the end of the `title`.  Means every image with the same `filename` needs to be checked each time.

paulbert:
Error message if file exists:

**An image with that filename exists.  Please rename or select other image.**

PR #6868, opened 2020-07-19
Add warning that an image with that name, has been already added. Issue #6852

chtelmko:
Issue #6852

Show alert, when user tries to add the same image( an image with the same name) to a course, when creating a new course.

### DIFF
--- a/src/app/shared/dialogs/dialogs-images.component.ts
+++ b/src/app/shared/dialogs/dialogs-images.component.ts
@@ -48,7 +48,7 @@ export class DialogsImagesComponent implements OnInit {
 
   uploadImage(event) {
     const file = event.target.files[0];
-    const imageExists = this.images.filter( img => file.name === img.filename).length > 0;
+    const imageExists = this.images.some( img => file.name === img.filename);
     if (imageExists) {
       this.planetMessageService.showAlert('An image with that filename exists. Please rename or select another image.');
       return;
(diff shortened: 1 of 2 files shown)


src/app/shared/dialogs/dialogs-images.component.ts
```diff
@@ -48,6 +48,12 @@ export class DialogsImagesComponent implements OnInit {
 
   uploadImage(event) {
     const file = event.target.files[0];
+    const imageExists = this.images.filter( img => file.name === img.filename).length > 0;
+    if (imageExists) {
+      this.planetMessageService.showAlert(`An image with that  ${file.filename}
```
paulbert, line 53: There is an extra space after `that`

src/app/shared/dialogs/dialogs-images.component.ts
```diff
@@ -48,6 +48,12 @@ export class DialogsImagesComponent implements OnInit {
...
+        filename exists. Please rename or select other image.`);
```
paulbert, line 54: Please put the string on one line.  When using backticks the newlines in the file are added to the string.

src/app/shared/dialogs/dialogs-images.component.ts
```diff
@@ -48,6 +48,12 @@ export class DialogsImagesComponent implements OnInit {
 
   uploadImage(event) {
     const file = event.target.files[0];
+    const imageExists = this.images.filter( img => file.name === img.filename).length > 0;
```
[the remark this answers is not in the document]
paulbert: In addition, it's more concise to use `.some` instead of `.filter`

src/app/shared/dialogs/dialogs-images.component.ts
paulbert: Please also remove `${file.filename}`

PR #6872, opened 2020-07-20
ReportsDetail: Filter members for count correction(fixes #6849)

### LINKED ISSUE #6849: ReportsDetail: Filter members for count
### Current Behavior

The counts of members in the **ReportsDetailComponent** have a couple of problems

1. The first admin is included, but that is not included in any of the other reports.
2. The counts are do not match the table found in **UsersComponent**

### Desired Behavior

The users for the counts in the report should be filtered to remove the first member (based on the `adminName` in the planet configuration) and any other members filtered by the table in **UsersTableComponent** (used in **UsersComponent**).

Image showing subset of members with only one in the **Members by Visits** list but two in the counts:
image

Total members in my local community according to reports & the members list (found in bottom right of table)
image

image

robinGiri:
issue: #6849 


paulbert:
This is better.  The count is still a little off, but I'm going to merge this so it gets in the latest release.

Issue #6830, opened 2020-07-14
ReportsDetail: Include team/member filter as choice for export CSV


paulbert:
We have a request to include the "Filter Members" as an option for filtering the **Export** options.

This should apply to everything except Health.

![image](https://user-images.githubusercontent.com/9203229/87375022-0b442d00-c559-11ea-8dd2-dd65df86d2fb.png)

PR #6874, opened 2020-07-20
[#6830] Add export team filter

singharpita:


src/app/manager-dashboard/reports/reports-detail.component.ts
```diff
@@ -420,4 +420,22 @@ export class ReportsDetailComponent implements OnInit, OnDestroy {
...
   }
+
+  filterByMember(filterValue, filterData?: []) {
+    let members = [];
+    const isSelectedMember = item => members.length === 0 ||
+      members.some(member => (member.userId === item.userId || member.userId.split(':')[1] === item.user));
+    if (filterValue !== 'All') {
+      this.couchService.findAll('teams', findDocuments({ teamId: filterValue._id, docType: 'membership' }))
+        .subscribe((teamMembers: any) => {
+          members = teamMembers;
```
paulbert, line 431: @singharpita This is a request to the database that will happen asynchronously to the rest of the function.  The function will return a value long before `members` is reassigned with the `teamMembers` from this request.

PR #6898, opened 2020-07-21
ReportsDetail: Fix member count on load (fixes #6883)

### LINKED ISSUE #6883: ReportsDetail: Two different member counts load
### Current Behavior

When the report loads, one set of member counts is shown first, and then another is used.

### Desired Behavior

The counts should always match the second one shown.  In **ReportsDetailComponent** we can remove `getTotalUsers` and the initial `setUserCounts` that follows to stop this first count from showing up.

https://www.loom.com/share/ba1dc29a9d16471392c1bc2d257fd885

Haoran-Hu:
Fixes #6883 

Work in progress

Fix member counts on **Report Detail** page


Haoran-Hu:
@paulbert On `reports-detail.component.html` we have corresponding text for first count, isn't that the reason the first member count is showing up?

Issue #6912, opened 2020-07-23
Enterprise- Add report: validation is missing in End date field


singharpita:
![Screenshot_2020-07-23 Planet Learning(1)](https://user-images.githubusercontent.com/27276132/88292773-cc149980-cd19-11ea-92f3-26577618fbfa.png)

paulbert:
The function is added to the field but it is not implemented correctly.  `endDateValidator` returns a validator function.  It is not a validator function itself.

PR #6937, opened 2020-07-23
Enterprise Reports: Validate number fields

### LINKED ISSUE #6910: Enterprise Reports: Income & Expenses only positive numbers allowed
Do not allow negative value
Lets make it same as financial transaction

jsschf:
Fixes #6910

### DIFF
--- a/src/app/shared/dialogs/dialogs-form.component.html
+++ b/src/app/shared/dialogs/dialogs-form.component.html
@@ -5,7 +5,7 @@
       <mat-checkbox *ngIf="field.type === 'checkbox'" formControlName="{{field.name}}">{{field.placeholder}}</mat-checkbox>
       <!-- input field -->
       <mat-form-field *ngIf="field.type === 'textbox'" class="full-width">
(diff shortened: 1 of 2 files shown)


jsschf:
@paulbert I'm in **TeamsReportsComponent** and I'm confused on how to validate the fields as there isn't a Form Control. I see the fields in `openAddReportDialog()`, but I can't find any relevant input fields. 

paulbert:
@jsschf In `addFormInitialValues` there is a `formControl` function.  This creates arrays for setting the **FormControl** values with **FormBuilder.group**.

I used that function since all of the fields are required and it cut down on some repetition to set things that way.  We already have one additional validator for `endDate`, but you could add another to that array if the field name matches any of the four in the issue.

https://angular.io/api/forms/FormBuilder#group

jsschf:
@paulbert Ok thanks that makes more sense

jsschf:
This should also make it easier to add the endDate validation for this issue #6912

PR #6944, opened 2020-07-27
CoursesView: Leave/join doesn't work after refresh

### LINKED ISSUE #6926: CoursesView: Join/Leave does not work after refresh
## Description
While in a certifications, trying to join/leave a course you must refresh the page for the action to be noted

### Steps to reproduce
Visit certifications in the manager settings -> Select a course (You are redirected to another page) -> Click join or leave (Refreshing the page shows the action)

jsschf:
Fixes #6926

### DIFF
--- a/src/app/courses/courses.service.ts
+++ b/src/app/courses/courses.service.ts
@@ -196,7 +196,8 @@ export class CoursesService {
     }));
   }
 
-  courseResignAdmission(courseId, type) {
+  courseResignAdmission(courseId, type, courseTitle?) {
+    const title = courseTitle ? courseTitle : this.getCourseNameFromId(courseId);
     const courseIds: any = [ ...this.userService.shelf.courseIds ];
     if (type === 'resign') {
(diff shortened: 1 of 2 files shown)


jsschf:
@paulbert It seems that the put request in `updateShelf()` in **UserService** isn't successful after a refresh. The snackbar that should pop up to notify the user the course was added/removed doesn't show up, and the "Http put request completed" message doesn't show up in the console. However it does show that the put request is made, just not successfully completed.

Other than that, I don't see any differences in the network tab of the requests that work as intended and the requests that don't work after a page refresh

paulbert:
@jsschf Thanks for doing some debugging.  It's an annoying silent error because it should throw something.

The issue is that `this.courses` is not set yet after the refresh.  Add an optional parameter for `courseTitle` in `courseResignAdmission()` and set it to `this.courseDetail.courseTitle` when calling the function in **CoursesViewComponent**.

When setting the `admissionMessage` use `courseTitle` if it's set, and use `getCourseNameFromId` if not.

PR #6945, opened 2020-07-27
Collections: Update count based on filter

### LINKED ISSUE #6822: Courses: collection selection
Is it confusing that the number of courses in a collection, does not match the selected collection?

It happens when you select the grade level first and then click on "collection" and select from the list.

number of courses in a collection, does not match the selection

jsschf:
Fixes #6822

### DIFF
--- a/src/app/shared/forms/planet-tag-input.component.ts
+++ b/src/app/shared/forms/planet-tag-input.component.ts
@@ -188,7 +188,7 @@ export class PlanetTagInputComponent implements ControlValueAccessor, OnInit, On
       initTags: this.initTags.bind(this),
       reset: this.resetDialogData.bind(this),
       startingTags,
-      tags: this.addTagCount(this.tags.filter(this.tagsService.filterOutSubTags), this.selectMany),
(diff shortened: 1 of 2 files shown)


src/app/shared/forms/planet-tag-input.component.ts
```diff
@@ -215,14 +215,11 @@ export class PlanetTagInputComponent implements ControlValueAccessor, OnInit, On
...
+  addTagCount(tags) {
     return this.mode === 'add' ? tags : tags.map((tag) => {
-      return !selectMany ? tag : ({
+      return ({
         ...tag,
-        count: this.filteredData.reduce(
-          (count, item: any) => count + ((item.tags || []).findIndex((itemTag: any) => itemTag._id === tag._id) > -1 ? 1 : 0),
-          0
-        )
+        count: this.filteredData.length
```
paulbert, line 222: This just shows the same number for each collection.

You should filter the `filteredData` for items that have the `tag` and then take the length of that.

PR #6990, opened 2020-07-30
CoursesProgress: Use UserProfileDialog

### LINKED ISSUE #6986: CoursesProgress: Use UsersProfileDialog
### Current Behavior

The column labels at the bottom of this view show the member profile image and the member's username.  Clicking does nothing.

### Desired Behavior

Either clicking on the member profile image or the member's username should open **UsersProfileDialog** with the member's information.

image

jsschf:
Fixes #6986

### DIFF
--- a/src/app/courses/progress-courses/courses-progress-chart.component.html
+++ b/src/app/courses/progress-courses/courses-progress-chart.component.html
@@ -38,9 +38,9 @@
   </div>
 </div>
 <div class="errors-container errors-user-total" #errorsUserTotal>
-  <div class="errors" *ngFor="let user of users">
-    <div>{{setTotal}}</div>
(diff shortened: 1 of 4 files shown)


jsschf:
@paulbert So I passed user information from submissions in **CoursesProgressLeaderComponent** to **CoursesProgressChartComponent** to be able to open the UserProfileDialog, but I ran into some issues:

The _set_ iterator in **CoursesProgressChartComponent** (lines 41-47) has a "total" value, which isn't present in the users object that I passed. So I need to have two *ngFor statements to iterate through the avatar and name; one for _sets_ and one for _users_ to have the dialog work correctly for each user. However I'm not sure how to implement that. My current implementation isn't working.

The only field in _set_ to match user information is "label", which is only a first name. So I could use `set.label === user.name`, but that would cause issues for people with the same first name.

In short, I'm stuck at the end. I thought about replacing _sets_ with _users_, but then I won't have the "total" field.

jsschf:
@paulbert It should be working now. Code Climate is failing because I copied the `openMemberDialog()` function from a different component.

Issue #7067, opened 2022-03-24
When screen size is too small we are not scrolling so it hides part of login screen.


lmmrssa:
Lets test with some small mobile sizes and if content fixs there then we can leave it for now. Else we need to be able to scroll when screen size is smaller than our design break point. i.e change overflow

rheam97:
@lmmrssa I address this in the branches dedicated to the past two issues, but I need to figure out why the mq stopped working first

Mutugiii:
Resolved in #7073

PR #7065, opened 2022-03-21
added media queries for login.scss (fixes #7061)

rheam97:
### Problem
#7061
no break line for "on android, try myPlanet"
login-container does not fit to size of the wrapper when mobile

### Steps to reproduce the problem

N/A refer to #7061

### Screenshots

Screen Shot 2022-03-21 at 7 52 29 AM

### Proposed solution

added media queries to add flex properties to myplanet div with "on android" and myplanet link for tablet
removed padding on wrapper and changed height of login container for mobile

PR #7073, opened 2022-03-28
responsive login screen (fixes #7062 and #7067)

### LINKED ISSUE #7062: Convert login design for mobile something like this
image
image

rheam97:
Fixes #7062 #7067

### Description
made mobile portrait, works for small screens
made padding 0 for portrait and landscape
also added landscape query, works for small screens as well
added div to wrap text in login left tile container so that display is row in portrait, does not affect  mobile landscape or web layout 

### Result
Screen Shot 2022-03-28 at 9 43 14 AM
Screen Shot 2022-03-25 at 1 37 53 PM

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -8,7 +8,7 @@ declare let gtag: Function;
 @Component({
   selector: 'planet-app',
   template: '<div i18n-dir dir="ltr"><router-outlet></router-outlet></div>',
-  styles: ['div { height: 100%; margin: 0} ']
+  styles: [ 'div { height: 100%; margin: 0} ' ]
 })
 export class AppComponent {
   constructor(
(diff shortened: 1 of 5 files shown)


Mutugiii:
@lmmrssa In some cases it seems that the form still jumps out of the `login-tile`

ole_login_overlap

In this case, when the height is at 650 and the width is in the range 400(value when the media query is triggered) to 550, some elements of the form are outside the white `login-tile`. The width range changes based on the height of the screen

Mutugiii:
@lmmrssa I've pushed some changes to incorporate the additions we discussed

lmmrssa:
@Mutugiii for mobile sizes it looks good. For tablet size make it looks similar to desktop but remove padding around. i.e. blue space around form

lmmrssa:
this is not as expected when you keep changing with of screen it jump between one with padding and without padding and again with padding. there is should be on break point and after that it should have padding.

Mutugiii:
@lmmrssa I've updated the PR incorporating a line break for the `On android? try Planet` line that was implemented in the #7065

lmmrssa (approved):

PR #7111, opened 2022-05-30
Dashboard: Add accordion Layout for smaller screens

### LINKED ISSUE #7068: Change dashboard list to accordion
When we do not have enough space on screen mostly for mobile (potrait) we can change the scrolling items to accordion
Show only list of item types Library, Courses and then when clicked show items added below as rows

Also for screen with smaller height on landscape we should either try to shrink the rows length or make it scrollable

Mutugiii:
Fixes #7068 

Adding an accordion layout for smaller screens in the dashboard


Mutugiii:
@lmmrssa Regarding the background color change issue for the accordion, it happens because there is a hover for the `mat-expansion-panel-header` selector that sets the background to inherit when the accordion is expanded.

I have tried using the code below to overwrite the class which works when a color is explicitly specified but it is unable to unset it to the previous colour when using revert/unset/initial

```
mat-expansion-panel-header:hover {
    background: revert;
]
```

Mutugiii:
stale

PR #7126, opened 2022-06-17
responsive resource view

### LINKED ISSUE #7122: responsive resource view
eng/resources/view/
Show detail should not split column but show as row in mobile design Try putting it below resource content and also put other version before content to compare which looks good

Mutugiii:
Fixes #7122 

Adding media query to change the grid style for mobile screens


Mutugiii:
This PR has the details below the resource

lmmrssa (approved):

PR #7166, opened 2023-06-13
adding responsive breakpoints for tablets (fixes #7164)

### LINKED ISSUE #7164: Suggestion: Introduce new breakpoint for Large Mobile phones & Tablets
Suggestion to introduce a newer breakpoint for responsive screens Large Mobile Phones & Tablets (780 -  1199px)

Mutugiii:
- New breakpoint between 780 to 1024px for tablets & large mobile device
- Add the mobile & tablet breakpoint variables to the environment file to allow for dynamic maintenance

Fixes #7164


src/environments/environment.ts
```diff
@@ -14,5 +14,7 @@ export const environment = {
   uPlanetCode: 'guatemala',
   uParentCode: 'guatemala@earth',
   upgradeAddress: window.location.origin + '/upgrade',
-  syncAddress: window.location.protocol + '//localhost:5984'
+  syncAddress: window.location.protocol + '//localhost:5984',
+  mobileBreakpoint: 780,
```
lmmrssa, line 18: is there reason for using it as env? It could be constant if it does not change for each deployment

Mutugiii: No reason for that. Ok

src/app/shared/checkMobile.service.ts
```diff
@@ -16,6 +19,11 @@ export class CheckMobileService {
 
   checkIsMobile(): boolean {
     this.setScreenWidth();
-    return this.getScreenWidth() <= 780 ? true : false;
+    return this.getScreenWidth() <= mobileBreakpoint ? true : false;
+  }
+
+  checkIsTablet(): boolean {
+    this.setScreenWidth();
+    return this.getScreenWidth() <= tabletBreakpoint ? true : false;
```
lmmrssa, line 27: not sure about the implication but instead of checking two different case we can use one function to check the type. As this function will also return true for mobile size if checked against table.

Mutugiii: That's true. 

In a unified function to check the device type, what other device info would you look out for, orientation?

src/app/shared/device-info.service.ts
```diff
@@ -0,0 +1,33 @@
...
+    this.screenWidth = window.innerWidth;
+    window.addEventListener('resize', this.updateScreenWidth.bind(this));
+  }
+
+  updateScreenWidth(): void {
+    this.screenWidth = window.innerWidth;
+  }
+
+  getDeviceSize(): DeviceSize {
+    const isMobile = this.screenWidth <= mobileBreakpoint;
```
lmmrssa, line 28: It will be only one true so instead of returning array and  we can just return one so you do not need to calculate all cases. And may be use like 'MOBILE'|'TABLET' in uppercase so it is like keyword.

Mutugiii: @lmmrssa Updated to use an enum instead, so the device type can be returned

src/app/landing/landing-home/landing-home.component.ts
```diff
@@ -10,7 +10,8 @@ import { LandingEventsService } from './landing-event/landing-events.service';
 })
 export class LandingHomeComponent implements OnInit {
 
-  isMobile: boolean = this.checkMobileService.checkIsMobile();
+  deviceType: DeviceType = this.deviceInfoService.getDeviceType();
```
lmmrssa, line 13: value initialize can go to constructor. Only defination can be left here

src/app/landing/landing-nav/landing-nav.component.ts
```diff
@@ -28,14 +28,15 @@ export class LandingNavbarComponent {
 
   menuState = 'out';
   isExpanded = false;
-  isMobile: boolean = this.checkMobileService.checkIsMobile();
+  deviceType: DeviceType = this.deviceInfoService.getDeviceType();
```
lmmrssa, line 31: same as previous, initialization can be moved to constructor

src/app/community/community.component.ts
```diff
@@ -83,7 +84,8 @@ export class CommunityComponent implements OnInit, OnDestroy {
   }
 
   @HostListener('window:resize') onResize() {
-    this.isMobile = this.checkMobileService.checkIsMobile();
+    this.deviceType = this.deviceInfoService.getDeviceType();
+    console.log(this.deviceType);
```
lmmrssa, line 88: remove console log

src/app/community/community.component.ts
```diff
@@ -46,7 +46,8 @@ export class CommunityComponent implements OnInit, OnDestroy {
   shareTarget: string;
   servicesDescriptionLabel: 'Add' | 'Edit';
   resizeCalendar: any = false;
-  isMobile: boolean = this.checkMobileService.checkIsMobile();
+  deviceType: DeviceType = this.deviceInfoService.getDeviceType();
```
lmmrssa, line 49: initialize in constructor

lmmrssa (changes requested):

Mutugiii:
@lmmrssa Initialized the deviceType variable in the constructor and removed the console message

lmmrssa (approved):

dogi (approved):

Issue #4714, opened 2019-08-29
Dashboard: myTeams should be only teams


paulbert:
Let's filter out the Enterprises from myTeams.

![image](https://user-images.githubusercontent.com/9203229/63958808-e33efe80-ca58-11e9-981b-7bf59900d3e7.png)

PR #7172, opened 2023-06-22
dashboard: add enterprises in myTeams list (fixes #7167)

### LINKED ISSUE #7167: myDashboard: myTeams does not show user enterprises
myDashboard does not show the users enterprises under the myTeams tab

Mutugiii:
Fixes #7167 

Take note of #4714

### DIFF
--- a/src/app/dashboard/dashboard.component.ts
+++ b/src/app/dashboard/dashboard.component.ts
@@ -139,7 +139,7 @@ export class DashboardComponent implements OnInit, OnDestroy {
         this.getData('teams', memberships.map((doc: any) => doc.teamId), { titleField: 'name', linkPrefix: '/teams/view/', addId: true })
       ])),
       map(([ memberships, teams ]: any[]) =>
-        teams.map(team => ({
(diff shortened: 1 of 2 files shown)


src/app/dashboard/dashboard.component.ts
```diff
@@ -139,7 +139,7 @@ export class DashboardComponent implements OnInit, OnDestroy {
         this.getData('teams', memberships.map((doc: any) => doc.teamId), { titleField: 'name', linkPrefix: '/teams/view/', addId: true })
       ])),
       map(([ memberships, teams ]: any[]) =>
-        teams.filter(team => team.type === undefined || team.type === 'team').map(team => ({
+        teams.map(team => ({
```
lmmrssa, line 142: there are teams without type as well. These are usually members and there are also records like credit. So removing filter might display other records as well.

lmmrssa (changes requested):

src/app/dashboard/dashboard.component.ts
Mutugiii: Reinstated the filter and additionally filtering for type === enterprises

lmmrssa (approved):

dogi (approved):

PR #7363, opened 2023-12-06
courses: improve data consistency (fixes #7362)

### LINKED ISSUE #7362: Hide subject and grade level when empty

Mutugiii:
Fixes #7362

Allow subjectLevel and gradeLevel to be empty on initialization. This fixes the issue where it was being initialized with an object and not a string.

Only display the gradeLevel and subjectLevel when the data is present


Mutugiii:
Closed in favour of #7367

PR #7464, opened 2024-05-22
manager: smoother fetch items (fixes #7463)

### LINKED ISSUE #7463: double popups across courses, resources and manager fetch
Handle the double popups on initialization when:
- pushing course/resource to community
- Fetching a course

Mutugiii:
Fixes #7463

In the manager fetch items, we have a situation where the popups open twice on page initialization.
Resolve it to only open one.

### DIFF
--- a/src/app/shared/sync.service.ts
+++ b/src/app/shared/sync.service.ts
@@ -82,7 +82,9 @@ export class SyncService {
   }
 
   confirmPasswordAndRunReplicators(replicators) {
-    return this.openPasswordConfirmation().pipe(switchMap((credentials) => {
+    return this.openPasswordConfirmation().pipe(
+      take(1),
+      switchMap((credentials) => {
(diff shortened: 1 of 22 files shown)


Mutugiii:
currently taking only the first observable in the replicators array. 

Could compare the 2 received and then take it only if they are the same

src/app/shared/sync.service.ts
```diff
@@ -137,7 +137,9 @@ export class SyncService {
   }
 
   replicatorsArrayWithTags(items, type: 'pull' | 'push', planetField: 'local' | 'parent') {
-    return this.stateService.getCouchState('tags', planetField).pipe(map(tags => this.createReplicatorsArray(items, type, tags)));
+    return this.stateService.getCouchState('tags', planetField).pipe(
+      take(1),
```
Mutugiii, line 141: currently taking only the first observable in the replicators array.

Could compare the 2 received and then take it only if they are the same.

dogi (approved):

PR #7646, opened 2024-10-24
all: smoother submissions list

### LINKED ISSUE #7630: submissions: table responsiveness
Responsiveness options for the submissions table.

image

RheuX:
Fixes #7630 

### Horizontal Scroll Screen Option:
image
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.15.0",
+  "version": "0.15.5",
   "myplanet": {
-    "latest": "v0.20.55",
-    "min": "v0.19.55"
+    "latest": "v0.20.64",
+    "min": "v0.19.64"
   },
   "scripts": {
     "ng": "ng",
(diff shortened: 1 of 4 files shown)


Gavinp14:
I like this idea. We would have to figure out how to style it, so it looks neat and organized

jessewashburn:
I think it's fine if people have to scroll a bit to see the very right hand side on narrow screens. Obviously, it's not ideal, but many pages on the site currently work that way, and it may not be the best use of time reinventing everything for the very few people who will be using a mobile browser.

sahilvunnam:
I personally like the horizontal scroll screen option, I think it is cleaner compared to stacking each row formation. Perhaps we could decrease the margin between each column or decrease the width of each column so the user doesn't have to scroll so much horizontally.

RheuX:
okay thansk for the feedback, ill stick with the horizontal scroll option then :)

Mutugiii (approved):

dogi (approved):

PR #7262, opened 2023-09-21
resources: more responsive mobile view (fixes #7163)

Mutugiii:
Implement responsive view for the courses table  list.

image

PR #7649, opened 2024-10-25
courses: smoother submissions toolbar (fixes #7631) 

### LINKED ISSUE #7631: submissions: toolbar responsiveness
Responsiveness for the Submissions toolbar

image

jessewashburn:
Added a filter list to add screen real estate on mobile. 

Fixes #7631 

image
image
image

Note that you have to scroll a little to the right to see everything on very small screens, but that’s consistent with how many of the pages work on mobile.


RheuX:
When I tried to test the code, it came out like this

I use the Samsung Galaxy S20 Ultra (I'm using a mac system - not sure if that makes a different)
image

Not sure if this is intended; since I'm guessing its because the picture is a higher dimension that what I tested.

But I would also recommended adding like the `filter_list` icon, so you have more space in the bottom, cause even in the picture you give, i feel its too cluttered.

Screenshot 2024-10-25 at 9 13 10 AM

It's something similar with #7262

RheuX (changes requested):
When I tried to test the code, it came out like this

I use the Samsung Galaxy S20 Ultra (I'm using a mac system - not sure if that makes a different)
image

Not sure if this is intended; since I'm guessing its because the picture is a higher dimension that what I tested. Could you please just double check this, if its work, then maybe its just me, and you can ignore this request :)

But I would also recommended adding like the `filter_list` icon, so you have more space in the bottom, cause even in the picture you give, i feel its too cluttered.

Screenshot 2024-10-25 at 9 13 10 AM

It's something similar with #7262

Mutugiii:
Agree with @RheuX using a filter_list would be a neater option

jessewashburn:
I have incorporated the filter list:
image
image
image

RheuX (commented):
Tablet looks great!
image

I still have the issues with the small screen for some reason. here's a screenshot of what I have with the dimension and the device type. To my knowledge, I have the latest version of your code, so it you can't replicate this, could someone replicate this issue and see if its just in my end

Screenshot 2024-10-30 at 1 46 23 PM

The only issue in the picture above its that you can't really use the search filter.

Screenshot 2024-10-30 at 1 51 56 PM

jessewashburn:
> I still have the issues with the small screen for some reason. here's a screenshot of what I have with the dimension and the device type. To my knowledge, I have the latest version of your code, so it you can't replicate this, could someone replicate this issue and see if its just in my end

I can't replicate your issue. Even on the narrowest screens, the worst that happens is the user has to scroll a bit to the right to get all the content, which is consistent with the behavior of the rest of the app on small screens. I am definitely open to suggestions if there's something specific about the code I could improve.  

[submissions_test.webm](https://github.com/user-attachments/assets/c6c72ebf-55c8-4284-9cb2-9cdf7d0d32e7)

RheuX (approved):

sahilvunnam (commented):
This is what I'm getting

Image 11-7-24 at 6 11 AM

jessewashburn:
> This is what I'm getting
> 

This branch has been around for a while; are you sure you're on the most recent version? Would you mind running git fetch origin before checking out the branch and testing again?

Mutugiii (approved):

Mutugiii:
Was able to replicate @RheuX issue. Opted to introduce an additional toolbar row for mobile

image

dogi (approved):

PR #7662, opened 2024-10-28
step-view: smoother view (fixes #7661)

### LINKED ISSUE #7661: courses-view: add ellipsis when resource name overflows
During a course step view in the resource selector(multiple resources) ensure that the name is truncated with an ellipsis if the name is very long

In the screenshot - resource 2
image

sahilvunnam:
fixes #7661

@Mutugiii The original code already had ellipsis if the title was too long, I just added a max-width to make sure there is no overflow of the title, also I couldn't get an image for this pull request since I don't have any resources within my courses on Planet

### DIFF
--- a/src/app/courses/step-view-courses/courses-step-view.scss
+++ b/src/app/courses/step-view-courses/courses-step-view.scss
@@ -57,4 +57,12 @@
     max-width: 100%;
   }
 
+  .ellipsis-title {
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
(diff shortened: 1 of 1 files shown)


Mutugiii:
You can try to create a course step with a few resources(could help with beta testing & issue finding of related features)

RheuX (changes requested):
Hello!

so, i added 5 resources to my step course, and this is what it looks like

image

it's still overflowing and I don't think i see any ellipsis.

Mutugiii:
stale

PR #7698, opened 2024-11-01
resouces: smoother search (fixes #7689)

### LINKED ISSUE #7689: courses and library: search should treat accented and unaccented vowels the same
e.g í should be treated as equivalent to i, else learners will have to be extremely specific about their search queries.

jessewashburn:
Fixes #7689 
When searching courses and library, characters a and á, e and é, n and ñ, etc, are now treated as equivalent. This makes searching Spanish titles easier and requires less precision from the user.

image
image
image
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.15.17",
+  "version": "0.15.32",
   "myplanet": {
     "latest": "v0.20.69",
     "min": "v0.19.69"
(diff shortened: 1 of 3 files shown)


Gavinp14 (approved):

sahilvunnam (commented):
For some reason my search still treats the characters the same. But mine might be the only one that does this.

Image 11-1-24 at 7 47 AM

jessewashburn:
> For some reason my search still treats the characters the same. But mine might be the only one that does this.
> 

That is actually the desired behavior. Treating them as equivalent deals with both possible cases: 

1) Someone made a course with a Spanish title but incorrectly used n in place of ñ. Searching the title using ñ will still deliver the result.

2) Someone is searching a course with a Spanish title containing ñ, but typed n instead. The result will still be delivered.

RheuX (approved):

Mutugiii (approved):

dogi (approved):

PR #7708, opened 2024-11-01
community: smoother services (fixes #5615)

### LINKED ISSUE #5615: Community Service: Add/remove button should be visible on page all time
when no. of link increases, button for Add/remove is hidden from the current page.
Screenshot from 2020-01-23 20-33-57

RheuX:
fixes #5615 

Now both action buttons for the links are sticky, so if you have a lot of links, you can click on them anytime
image

I also change the position of both actions buttons to make them sync.
image

I'm not sure if putting sticky on the bottom of the page is intuitive, or looks nice, but if you think I should do that instead of top let me know


Mutugiii (approved):

sahilvunnam (commented):
I would add color to the buttons, the rest of the pages within the Finances Tab or Reports tab has a color scheme for their buttons (Yellow and Blue). It makes the UI look cleaner

jessewashburn (approved):

Gavinp14 (approved):

dogi (approved):

PR #7711, opened 2024-11-04
manager: smoother report steps filter (fixes #7679)

### LINKED ISSUE #7679: manager: report when time changes is not done on courses and steps
image

RheuX:
Fixes #7679 

Oct - Nov Filter Range
image

Just November
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.15.24",
+  "version": "0.15.33",
   "myplanet": {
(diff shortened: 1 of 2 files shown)


jessewashburn (approved):

jessewashburn:
Definitely an improvement. I think it would be nice if when changing the date range, the charts would zoom in on that range. Currently I'm not sure that there's a lot of value in filtering this way.

Gavinp14 (approved):

Mutugiii (approved):

dogi (approved):

PR #7714, opened 2024-11-04
community challenge: popup only appears on fresh login

jessewashburn:
I fixed the issue where the popup happens every time the community page is refreshed. Now only happens on a new login. 

Still need to add a persistent notification.

### DIFF
--- a/src/app/community/community.component.ts
+++ b/src/app/community/community.component.ts
@@ -101,11 +101,18 @@ export class CommunityComponent implements OnInit, OnDestroy {
   }
 
   communityChallenge() {
+    // Set the challengeActive status
     this.challengeActive = ((new Date() > new Date(2024, 9, 31)) && (new Date() < new Date(2024, 11, 1)));
-    if (this.challengeActive) {
-      this.openAnnouncementDialog();
+
+    // Check if the popup was shown for this login session
+    const popupShown = localStorage.getItem('announcementPopupShown');
+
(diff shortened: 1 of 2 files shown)


Gavinp14 (approved):
LGTM, it only pops up on a fresh login and not every time when you refresh a page

jessewashburn:
I am wondering if this Logout() function is called when users are logged out automatically. Also, how often are users logged out automatically? My concern would be if they are allowed to just stay logged in and thus only see the popup once and are not encouraged to do the challenge. @Mutugiii

RheuX:
If you are concern with popup only appearing once since they have to log out and log back in, you could do either a int stored in a session storage (data would persist even on reload, only when closed they would restart, but at that point, you would have to log back in also), where every time they go to dashboard for the 3rd or 5th time it would pop up also. If not, we should implement the notification system and challenge tab for them.

jessewashburn:
It seems when the session times out, the Login() function is not called, so I am now switching the flag upon login. 

This still leaves the issue of making it obvious to users that there's an ongoing challenge even if they have already seen the popup during that session. I will work on that tonight.

Mutugiii:
If it is a question of permanence(assurance of not extending the session) then sessionStorage should do the trick like @RheuX has pointed out.

Though in the current implementation it seems like it would do the trick with the the removal being set on the login method

RheuX (approved):

Mutugiii (approved):

PR #7716, opened 2024-11-04
7712 challenge success message trigger mechanics (Not final ui)

Gavinp14:
announcmentSuccess

This is the first iteration of the success message after all steps are completed. Obviously, this needs to be changed and updated based off of the specific user reqs. Just figured I would start working on it to get an idea of where we want to go with this.

### DIFF
--- a/src/app/shared/dialogs/dialogs-announcement.component.html
+++ b/src/app/shared/dialogs/dialogs-announcement.component.html
@@ -32,7 +32,7 @@
   </div>
 
   <div *ngIf="checkAllStepsCompleted" class="steps-complete-message">
-    <span>Congratulations! You have completed all the steps!</span>
+    <span i18n>¡Felicidades, has completado todos los pasos!</span>
   </div>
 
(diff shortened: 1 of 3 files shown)


RheuX (approved):

jessewashburn (changes requested):
This message will need the i18 tag, or else it won't translate. Since the audience for this course speaks Spanish, we could maybe just hardcode it too. "¡Felicidades, has completado todos los pasos!" would be the translation.

Gavinp14:
good catch Jessie. I will make these changes once the team decides on what they want to do for the success message Ui.

sahilvunnam (commented):
If PR #7721 get's approved then we would have to choose how we would format this because this PR has progress indicators that are also above the thermometer

Gavinp14:
Exactly, I'm going to put a discussion label on this if that is the case.

jessewashburn (approved):

sahilvunnam (approved):

PR #7717, opened 2024-11-04
community: dynamic current and target goal

RheuX:
Fixes Group Challenge:

image

2 of my user just finished the challenge, so now it will show $2 out of $500. Also, i change some of the function so that you can change the goal more easily, since its hardcoded right now. (theres 2 function to get and set the goal of the challenges)


RheuX:
oops wrong target branch

PR #7718, opened 2024-11-04
challenge: dynamic current and target goal

RheuX:
Fixes Group Challenge:

image

2 of my user just finished the challenge, so now it will show $2 out of $500. Also, i change some of the function so that you can change the goal more easily, since its hardcoded right now. (theres 2 function to get and set the goal of the challenges)

### DIFF
--- a/src/app/shared/dialogs/dialogs-announcement.component.html
+++ b/src/app/shared/dialogs/dialogs-announcement.component.html
@@ -40,7 +40,7 @@
     </div>
   </div>
   <div class="thermometer-goal">
-    <span>$0</span>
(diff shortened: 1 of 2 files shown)


jessewashburn (approved):

Mutugiii:
I prefer the implementation in #7720

sahilvunnam (commented):
I also prefer #7720 implementation of the thermometer

RheuX:
I'll change this so that it will only give this the function. I see the main issue now.

Mutugiii (approved):

PR #7720, opened 2024-11-04
challenges: smoother progress bar label

RheuX:
Problem:
Screenshot 2024-11-04 at 12 01 41 PM

Fixes:
image

If there is enough space for the label, it will act as previous code

image

### DIFF
--- a/src/app/shared/dialogs/dialogs-announcement.component.html
+++ b/src/app/shared/dialogs/dialogs-announcement.component.html
@@ -34,7 +34,7 @@
   <div class="thermometer-container">
   <div class="thermometer">
     <div class="thermometer-bar" [style.width.%]="getGoalPercentage()">
-      <div class="thermometer-label">
+      <div class="thermometer-label" [ngClass]="{'outside': getGoalPercentage() < 8}">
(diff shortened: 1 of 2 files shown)


jessewashburn:
What is the number in the parentheses next to the percentage?

RheuX:
its the number of how many people finished with the challenges, i was considering to removing it, but what if each challenge does not equal 1 participant = 1 dollar, so i just keep it in there

RheuX:
also, you saw 10% there, but its still (2), because i just changed the numerical value of the bar, when testing it, in practice, its should be equal to the ratio we set

Gavinp14:
Do you think there should be a way so instead of the learners just seeing "(2)" or any given number in parenthesis, it says like "(2 entries)". This way perhaps people are not confused by the number. Just my thoughts though.

sahilvunnam (commented):
I personally prefer the percentage over the other format for the thermometer, also I would say "2 entries completed". You could switch the word entries, but I think you should keep the word "completed".

jessewashburn (approved):

Mutugiii (approved):

PR #7723, opened 2024-11-05
challenges: notifications system

RheuX:
image

Right now, notifications will show everytime that dialog challenge comes up (#7714 - should work with this, if this only pops up once).

~~The only thing that isn't working right now is the notification from the page (I think the router and click event handler is messing with each other. But the dropdown should be working for now. Ill try to fix the notifications also~~

Both notifications should be working right now, the only part I would like to fix is making it more neat. I had to import the dialogNotifications, and functions in 2 different places, wondering if i could make it more tidier

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.15.24",
+  "version": "0.15.37",
   "myplanet": {
-    "latest": "v0.20.76",
-    "min": "v0.19.76"
+    "latest": "v0.20.86",
+    "min": "v0.19.86"
   },
   "scripts": {
     "ng": "ng",
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.15.17",
+  "version": "0.15.24",
   "myplanet": {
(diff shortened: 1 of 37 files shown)


RheuX:
Oh, when the challenges is finished, it should stop sending notifications (if it does prompt it), i didn't check for that edge case, unless a PR has taken care that if a challenge is finished, it wouldn't prompt the challenges again

jessewashburn (changes requested):
Generally, this works nicely. I incorporated my popup fix and found some issues: 

1) The little red number that shows how many notifications there are is not updating until I refresh the page.

2) A new notification is sent when I click the community page, regardless of whether the popup happens or not. 

3) I have completed the entire challenge and still get the popup/notifications, but that should probably be fixed by a separate PR. 

[challenge_notifications_test.webm](https://github.com/user-attachments/assets/41505d99-c43c-476d-8d90-67820f3c92ea)

RheuX:
Cool thanks for the review, ill change some of them later!

Gavinp14 (changes requested):
I am also getting the same issues as Jesse with this. Not exactly related but is it necessary to include the exact time in the notification (ex: 12:41:02 PM)? Maybe just the date (ex. Nov. 6th) is enough.

RheuX:
I will close down this PR, it will be finished by #7746

PR #7737, opened 2024-11-06
courses/library: disable rating without enrollment

RheuX:
Currently WIP

Main Issues:
- [ ] When you're enrolled - when you first start rating it, it would not let you do it first, but then it would work after that.
- [ ] Opening the details courses and library would still allow you to rate it.

Other than that -> features that are working (for me to keep track)
- You can't rate courses (and the stars won't get written nor updated) if you haven't enrolled yet
- If you're enrolled, you can rate courses (but only after the second try) (**trying to fix this right now**)


Gavinp14 (approved):

jessewashburn (approved):

sahilvunnam (approved):

Gavinp14 (approved):

jessewashburn (dismissed):
This works great but this message doesn't really make sense since you don't join a resource. Should say something like "Please add the resource to your library before rating!"
image

dogi (approved):

Issue #7741, opened 2024-11-07
challenge upgrades


Mutugiii:
upgrades on the included planets 
notification system

RheuX:
remove label

PR #7746, opened 2024-11-07
challenges: notification system changes

RheuX:
Notification now works for the challenges, 

Notes:
> only edge case is that the status only first initalized when youre in the community post. If you refresh anywhere and you never got the dialog challenged open from community, if you open it from the notification, it would say you didn't do any of them.
- need to confirm if this is fix in the merging (meaning, just refresh at any point, and open the dialog from the notification. -> if you see the status check, that means its fixed)

P.S. If you dont understand the edge case, then let me know, so i can clarify it

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.15.37",
+  "version": "0.15.38",
   "myplanet": {
     "latest": "v0.20.86",
     "min": "v0.19.86"
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.15.24",
+  "version": "0.15.37",
   "myplanet": {
-    "latest": "v0.20.76",
-    "min": "v0.19.76"
+    "latest": "v0.20.86",
+    "min": "v0.19.86"
   },
   "scripts": {
     "ng": "ng",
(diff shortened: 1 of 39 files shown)


RheuX:
i personally cannot test it out, since the code is in the included planets (i think the code for checking the status being check twice or thrice, so even if i brute force it then it wouldn't work, but if there is some issues, please let me know, so i can probably check it out myself

jessewashburn:
The notification system plays nicer with my popup code. It no longer duplicates notifications when navigating to the community page - only when there's a new popup upon login. 

**Some problems(Not necessarily related to this PR. Feel free to create separate issues or I can do so on Monday):** 

the notification number only updates on refresh.  
image

My progress is not updating. 

image

image

I'm allowed to join multiple times, because the Unirse button doesn't go away.. 

image

RheuX:
@jessewashburn could you have a video on how you do that (i meant the progress not updating, is it just from the community post to begin with)? sorry, for the inconvenience.  

and the notification thing, mutugi said he would look into it because im not sure if its supposed to be responsive or not (some notifications require you to refresh first), and i cannot test out if thats the case because most of the notification i have requires 2 users (unless i run in 2 different ports, but probably should wait for mutugi

RheuX:
> @jessewashburn could you have a video on how you do that (i meant the progress not updating, is it just from the community post to begin with)? sorry, for the inconvenience.

also, if you could, let me know which planet did you tested on, since it only works in some planets (not our planet)

Mutugiii:
Merging with #7741

PR #7755, opened 2024-11-08
adding label when loading courses/library resources (fixes #7750)

### LINKED ISSUE #7750: resources/courses: `No content provided` label loading
Either display a loading label or no label when the data is loading

Gavinp14:
Added "loading courses" label when loading courses. Need to do this for resources as well, then it will be ready for review.

### DIFF
--- a/src/app/courses/courses.component.html
+++ b/src/app/courses/courses.component.html
@@ -243,8 +243,9 @@
   </ng-container> 
 
   <ng-template #loadingMessage>
-    <div class="view-container" i18n>Loading Courses</div>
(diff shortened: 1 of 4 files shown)


RheuX (approved):

jessewashburn (approved):

jessewashburn:
This works nicely.

I think I remember Dogi mentioning that he doesn't like when there are so many unneeded line changes. It's a bit difficult to see what lines from the HTML file were actually changed because the whole thing is highlighted red/green, which I don't think is needed.

sahilvunnam (approved):

PR #7774, opened 2024-11-13
cannot add blank course (fixes #5040)

### LINKED ISSUE #5040: Teams: Do not allow adding an empty course

Gavinp14:
An Alert will show if a user tries to a add a course with no steps. Courses cannot be blank
fixes #5040


Gavinp14:
Good catch there, I will fix that. This has nothing to do with the issue, rather something on my end where every time I do ng-serve it does this to the json file. I will have to look into this

sahilvunnam (approved):

RheuX (approved):

jessewashburn (approved):

PR #7776, opened 2024-11-14
resources: smoother collections display (fixes #7772)

### LINKED ISSUE #7772: collections: name length collections issues
library/courses collections problems happen if name of a collection is getting long

image

image

**The second picture happened because one long non-space word -- if you have spaces it will act as the first photo**
not sure if we should limit or just truncate the name of the collections

sahilvunnam:
fixes #7772

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.15.44",
+  "version": "0.16.2",
   "myplanet": {
     "latest": "v0.20.96",
     "min": "v0.19.96"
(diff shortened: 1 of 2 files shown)


jessewashburn (approved):

RheuX (approved):

Gavinp14 (approved):

Mutugiii (changes requested):
This is only supported in the creation form, however you can edit the name and bypass the limit

image

Mutugiii (approved):

jessewashburn (approved):

dogi (approved):

PR #7790, opened 2024-11-15
Courses: more accurate label

### LINKED ISSUE #4019: "Manage" for updating
I think it's a little strange that the "Manage" button is used to update courses but it's used for another purpose for teams. It's also the wording of "Manage" that bothers me but this might be just me. I just thought it was a little nonintuitive since whenever I think "Manage" I think of admin rights, beyond just updating.

Capture

Proposed solution, if this is a problem at all:
Rename instances where "Manage" is used to "Update" when the function is to update.

jessewashburn:
fixes #4019

The old manage button just took you edit the course, so changed the label to Edit Course and logo changed to reflect it.

image

### DIFF
--- a/src/app/courses/courses.component.html
+++ b/src/app/courses/courses.component.html
@@ -180,8 +180,8 @@
               <span i18n>Feedback</span>
             </a>
             <a mat-menu-item *ngIf="element.canManage" (click)="updateCourse(element.doc)">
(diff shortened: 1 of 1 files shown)


Gavinp14 (approved):
Looks good.

Mutugiii (approved):

dogi (approved):

PR #7797, opened 2024-11-19
removed quotations from notifications (fixes #4446)

### LINKED ISSUE #4446: Notification link consistency
either quotation ""  should be applied in every type of message or can be ignored for all
right now its kind of mixup .
Screenshot from 2019-07-30 07-18-45

Gavinp14:
fixes #4446


jessewashburn (approved):

jessewashburn:
I think this format is preferable for all notifications going forward. 

Too bad it doesn't work retroactively, but I imagine that would involve some CouchDB manipulation and would be more trouble that it's worth.

RheuX (approved):

PR #7806, opened 2024-11-19
First Iteration of Acheivements Page Revamp (fixes #7639)

### LINKED ISSUE #7639: achievements page revamp
Improve the styling of the achievements page. More in line with the myDashboard page styles and theme

image

Gavinp14:
Revamped the Achievements page to take up more real estate, open to feedback and suggestions to improve this even further. 
revamp

fixes #7639

### DIFF
--- a/src/app/users/users-achievements/users-achievements.component.html
+++ b/src/app/users/users-achievements/users-achievements.component.html
@@ -23,19 +23,24 @@
   </mat-toolbar>
   <div class="view-container">
     <div *ngIf="achievements !== undefined" class="achievements-container">
-      <h2>{{user.firstName}} {{user.middleName}} {{user.lastName}}</h2>
-      <div>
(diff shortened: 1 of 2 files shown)


jessewashburn:
This is definitely an improvement. If it's something they print out, it would be good to have the OLE logo somewhere on there, so it looks less like something a person could make for themselves in a word doc. Maybe have it centered at the bottom.

Gavinp14:
I agree @jessewashburn, I would think since this has to do with printing the page rather than revamping the Ui it would be its own issue. I'm going to add a logo somewhere on the Ui because I think it would look nice.

Mutugiii:
While the printing is handled with the generatePDF as @RheuX points out, I do agree with @jessewashburn on having the OLE logo and maybe also the users profile picture to improve the page visuals, also maybe having the separator lines with a light blue color for added visual effects

Gavinp14:
Sounds good I will make some of these changes.

Gavinp14:
updatedRevamp

Here is what the Revamp now looks like

jessewashburn (approved):

Mutugiii (approved):

dogi (approved):

PR #7814, opened 2024-11-20
chat: smoother provider switching (fixes #7638)(fixes #7809)

### LINKED ISSUE #7638: chat: changing provider starts new chat
Changing the providers should start a new chat with the new provider

image

### LINKED ISSUE #7809: chat: toggle the right provider for each chat
if you start a chat with perplexity or gemini, and go back to that chat later on, the chat model is openAI, since it's hardcoded to be openAI, from the beginning of the chat window

RheuX:
Fixes #7638
Fixes #7809 

New chat:
If you are currently on a chat with a specific provider, then switching provider will create a new chat.
If you haven't selected an existing chat with a model, than switching chat does nothing.

Select existing chat:
If you select an existing chat, it will switch over to the model that the chat previously used (since we want to let a chat be tied to a specific model)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.15.55",
+  "version": "0.15.64",
   "myplanet": {
     "latest": "v0.20.99",
     "min": "v0.19.99"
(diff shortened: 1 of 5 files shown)


RheuX:
actually i need to change some things here

RheuX:
I might make another PR that handles another issue (#7809) to combine into this PR, since I might refactor a couple stuff in this PR to have a smoother flow

Mutugiii (approved):

Mutugiii (approved):

dogi (approved):

dogi (approved):

PR #7846, opened 2024-11-27
More Secure Password Criteria (fixes #6768)

### LINKED ISSUE #6768: Password: Security Issue
When I create an account by clicking "Become a member", the app currently allows any length of password & combinations to sign in. Wouldn't it better to make some restrictions? For example, the password length needs to be 4<password<16 or needs to be combined with any upper letters or symbols.

I know "Planet Communities" is on a local network, but the user might share their devices with others.

Gavinp14:
Will Not let you create an account unless the password meets the following criteria

- Minimum 8 characters long
- Must contain at least one number
- Must contain at least one special character

This is something that should be up for discussion, so please comment and leave some thoughts on the password criteria and if it should be adjusted.

weakpass

### DIFF
--- a/src/app/shared/forms/form-error-messages.component.ts
+++ b/src/app/shared/forms/form-error-messages.component.ts
@@ -35,7 +35,7 @@ import { AbstractControl, AbstractControlDirective } from '@angular/forms';
       invalidEndTime {End time cannot be before start time}
       dateInPast {Cannot be before current date}
       invalidPassword {Password is not valid}
-      weakPassword {Password is too weak}
+      weakPassword {Weak password. Hover for more info}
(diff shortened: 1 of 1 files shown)


jessewashburn:
I think this is a great suggestion. Users could potentially be keeping sensitive health and financial information on the app, so I think security is definitely a concern. 

Only suggestion would be a more a descriptive message about what the password criteria is, but this is already in line with how a lot of other apps handle this.

jessewashburn (approved):

RheuX (approved):

Gavinp14:
I agree Jesse, I am going to edit the error message, so it displays the password criteria.

PR #7871, opened 2024-12-02
feedback: smoother titles (fixes #2280)

### LINKED ISSUE #2280: Feedback title in general
when feedback is given in general the title if feedback shows as "Question regarding/".
Is it okay or can be changed into something "Question regarding general feedback" when not attached with any course,library or any other subject.
screenshot from 2018-10-04 20-05-04

jessewashburn:
Fixes #2280

Feedback that lacks a url (general feedback submitted from the home page) now has a more descriptive title instead of just "/" Note that I moved the feedback title logic from the html to the ts. 

Works in list view, detailed view, and delete click pop up. 

image
image


Gavinp14 (approved):

RheuX (approved):

Mutugiii:
- You need to ensure you're updating couchdb with the new title format: when editing the title form field it defaults to `Question regarding /` -> Seems to be getting this from couchdb. 

- I was also thinking about the titles and I'm not sure if we want question regarding general feedback -> Not too sure about the general grammar. Maybe we can go for something like General Bug/Question/Suggestion

jessewashburn:
> * You need to ensure you're updating couchdb with the new title format: when editing the title form field it defaults to `Question regarding /` -> Seems to be getting this from couchdb.
> 
>     * I was also thinking about the titles and I'm not sure if we want `question regarding general feedback`(Not too sure about the general grammar). Maybe we can go for something like General Bug/Question/Suggestion

image
image

Thank you for the feedback @Mutugiii . Turns out I should have just been editing the feedback.directive.ts file. The placeholder now updates couchdb. I also improved the placeholder grammar. It doesn't work retroactively though.

RheuX (approved):

jessewashburn:
Actually, I recommend closing this one. I have incorporated the functionality into this other feedback related PR:  #7907

Mutugiii:
continued by #7907

PR #7873, opened 2024-12-02
courses:  limit step titles length (fixes #7867)

### LINKED ISSUE #7867: long titles in course creation
image

Gavinp14:
fixes #7867
Added a max length for creating a step title at 150 characters. If we feel like this is too short or too long, I can always make adjustments.  I don't think that there will be an actual need to create titles over 150 chars for steps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.15.72",
+  "version": "0.15.81",
   "myplanet": {
     "latest": "v0.21.12",
     "min": "v0.20.12"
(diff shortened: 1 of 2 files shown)


RheuX (changes requested):
I think this would work on larger screen,

for smaller screen it would still look like this.

image

I figured we should probably truncate or anything based on screen sizes

Gavinp14:
smallscreenstepsfix
It will now be responsive for smaller screens. Ellipses will also be added on all sized screens if step title exceeds a certain length.

RheuX (approved):

jessewashburn (approved):

Mutugiii (approved):

dogi (approved):

PR #7931, opened 2024-12-12
chat: smoother ux (fixes #7915)

### LINKED ISSUE #7915: chat: save state different user
create a chat, switch account, chat history still persist on the chatbox

image

RheuX:
Fixes #7915 

When opening up the chat for the first time, the chat would be clear. Also when switching between user, it would be clear.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.15.97",
+  "version": "0.16.5",
   "myplanet": {
     "latest": "v0.21.34",
     "min": "v0.20.34"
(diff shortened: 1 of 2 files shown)


Gavinp14 (approved):

Gavinp14:
It worked, the pouchdb error is not causing any direct errors related to this pr right?

RheuX:
not sure, because to my knowledge, its not supposed to happened like this, everytime you open chat, its supposed to be a clean slate, but its somehow getting the last conversation always.

my solution just avoid the chat updating for the first time, which avoids initial problem, but im not sure why its happening, upon testing, i saw the error, I'm not sure if its related or not

jessewashburn (approved):

Mutugiii (approved):

dogi (approved):

PR #7932, opened 2024-12-12
resources: smoother collections (fixes #7928)

### LINKED ISSUE #7928: collections name if too long, overflows on library/courses
when selecting collection in library/courses and if its too long

image

related to #7772, I might just put ellipsis instead of imposing a maxLength, so this would be an issues if someone were to put a collection name bigger than the screen

RheuX:
Fixes #7928 

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.16.5",
+  "version": "0.16.10",
   "myplanet": {
     "latest": "v0.21.40",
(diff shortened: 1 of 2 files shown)


Gavinp14 (changes requested):
For me, it is still overflowing on mobile or $screen-sm. Desktop looks good.
axelissue

Gavinp14 (approved):

jessewashburn:
lgtm. Still overflows in this window, but that could be a separate issue 
image

dogi (approved):

PR #7952, opened 2024-12-17
dashboard: smoother UI (fixes #7739)

### LINKED ISSUE #7739: myDashboard: myChat and myProgress buttons are crowded
mychat - Since the chat is not really related content-wise to myLibrary, I don't think they should share a square. It's too easy to click it  by accident. I would recommend moving it to the myLife row with its own row. 

myProgress - Courses and Progress are somewhat related content-wise, but it's still to easy to click the wrong one by accident. I would recommend moving it to the myLife row or somewhere inside the courses page.

image

jessewashburn:
fixes #7739

Moved myChat and myProgress to their own tiles instead of crowding the myCourses and myLibrary tiles. 

admin view

image

learner view

image


Gavinp14 (approved):
This layout makes more sense.

PR #7953, opened 2024-12-17
courses: smoother deletion snackbar (fixes #7941)

### LINKED ISSUE #7941: courses: long non-spaced titles on snack bar
Snack bar Title overflows when it is a non-spaced long title.
Screenshot 2024-12-16 120506

Gavinp14:
fixes #7941

### DIFF
--- a/src/app/courses/courses.component.ts
+++ b/src/app/courses/courses.component.ts
@@ -251,6 +251,14 @@ export class CoursesComponent implements OnInit, OnChanges, AfterViewInit, OnDes
     });
   }
 
+  truncateMessage(message: string, maxLength: number=50): string {
+    if (message.length <= maxLength) {
+      return message;
+    }
+    const truncatedMessage = message.slice(0, maxLength - 3);
+    return truncatedMessage + '...';
+  }
+
   deleteCourse(course) {
     const { _id: courseId, _rev: courseRev } = course;
     return {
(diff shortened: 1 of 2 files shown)


jessewashburn (approved):

jessewashburn (changes requested):
Unfortunately, this will truncate not just the title, but the information following it. E.G if you join a course, you would want it to say "aaaaaaaaa.... added to your dashboard" not just "aaaaaaaaaaa..."  The universal approach may not work after all. 

image

Gavinp14:
I see I will make this change just for the courses snack bar instead

RheuX (commented):
is it supposed to like this? or is it supposed to be truncated?

image

image

Gavinp14:
hmmm, it does not do this for my screen. @Mutugiii could you check this pr out on your side to see if anything is wrong?

jessewashburn:
> is it supposed to like this? or is it supposed to be truncated?

It's because the change for this PR is only related to the course deletion snackbar. For adding/removing courses to/from myLibrary, that's related to the courses.service.ts, similar to PR #7956

jessewashburn (approved):

Mutugiii:
@Gavinp14 Getting a similar issue to that of @RheuX

jessewashburn:
Fixed by #8088

PR #7970, opened 2024-12-19
smoother member role title (fixes #7966)

### LINKED ISSUE #7966: teams: member role overflow
roleOverflow
nonspacedrole

Gavinp14:
fixes #7966

### DIFF
--- a/src/app/teams/teams-member.component.html
+++ b/src/app/teams/teams-member.component.html
@@ -10,7 +10,7 @@
       {{ member?.doc?.leadershipTitle | slice:0:40 }}
       <span *ngIf="member?.doc?.leadershipTitle.length > 40">...</span>
     </p>
-    <p class="primary-text-color" *ngIf="member?.role">{{member?.role}}</p>
+    <p class="break-word primary-text-color" *ngIf="member?.role">{{member?.role | slice:0:485}}<span *ngIf="member?.role.length > 485">...</span></p>
   </mat-card-subtitle>
   <button *ngIf="actionMenu.length" class="top-right-icon" mat-icon-button [matMenuTriggerFor]="memberMenu">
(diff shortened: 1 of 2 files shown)


jessewashburn (changes requested):
If you scroll down, there is still overflow. 
image

Gavinp14:
@jessewashburn I decided to remove word break and just make it fixed on one line, I realized you need to leave room for the tasks user is assigned to. Let me know if it is still overflowing.

jessewashburn (changes requested):
I think it should allow for slightly longer titles since there is plenty of space
image

Gavinp14:
@jessewashburn I agree there is plenty of room, however when a member is assigned more than two tasks, a member title that is more than two lines will cause overflow for the tasks as pictured below. I guess maybe this could be a separate issue in regard to the tasks list maybe. 
jesseResponse

jessewashburn:
> @jessewashburn I agree there is plenty of room, however when a member is assigned more than two tasks, a member title that is more than two lines will cause overflow for the tasks as pictured below. I guess maybe this could be a separate issue in regard to the tasks list maybe. Could you please try to replicate this? 

I see. Yes, that does happen, but it can also happen when there is a long name. 
image
Rather than catering to the task overflow issue, I would set up the title truncation assuming that tasks will not over flow and then fix the tasks issue separately.

Gavinp14:
I will take a look and see what I can do, I might have to adjust some stuff in the task's component.

jessewashburn (changes requested):
image

I think you still need to add word break. Not sure if you intended to fix the tasks overflow here, but this is what it looks like for me.

jessewashburn:
Fixed by PR #8240. Recommend closing

PR #7982, opened 2024-12-20
all: smoother navigation icons (fixes #7980)

### LINKED ISSUE #7980: nav-bar - more distinctive chat icon
Chat and feedback icons look very similar. 
image

Consider using this mat icon for chat, which evokes the back-and-forth nature of an AI chat. 
image

jessewashburn:
image
Richard mentioned some confusion with the icons in the top right of the nav-bar, and I agree. I think this new icon is much clearer than the simple text bubble. 

fixes #7980


Gavinp14 (approved):
I agree, this looks a lot better than the previous icon.

Mutugiii:
Makes sense, potentially you can look into some of the other icons around the app particularly during mobile screens and tackle them in this PR as well

image

The ones I find a bit unclear are the courses, resources and maybe dashboard...

Gavinp14 (approved):
The dashboard icons are a lot crisper here. I am not sure how many options are out there for the ai chat icon but were you able to find anything with like an ai logo to let the user know that it is an ai message rather than talking to another person?

jessewashburn:
> The dashboard icons are a lot crisper here. I am not sure how many options are out there for the ai chat icon but were you able to find anything with like an ai logo to let the user know that it is an ai message rather than talking to another person?

I considered this, but the options for more specific AI chat icons would either have the word AI, which doesn't work for Spanish since they call it IA, or some kind of robot image, which I don't think suits our style. 
chatbot(1)
chatbot

jessewashburn:
I opted to put all the icon options in this doc. If anyone wants to suggest new ones, feel free to add to it. https://docs.google.com/document/d/1tmiofPLZW1uve1n1z_5sOzE2EucfsghaeYUDEAcVWtI/edit?usp=sharing

Mutugiii:
I think the suggested ones are an overall improvement on the current ones and we should implement them, for the resources I prefer option 1

jessewashburn:
> I think the suggested ones are an overall improvement on the current ones and we should implement them, for the resources I prefer option 1

Great, then this PR should be good to go.

Mutugiii:
For reference:

image

dogi (approved):

PR #7991, opened 2024-12-20
achievements: smoother phone number (fixes #7985)

### LINKED ISSUE #7985: achievements: allows non-numbers for phone number
invalidNumebr

Phone number input should only allow numbers, not anything

Gavinp14:
fixes #7985


Mutugiii:
This is a design choice that aligns with the phone number field in the profiles page

Issue #8037, opened 2025-01-03
myHealth: more visible sections in examinations


jessewashburn:
Sections like "Observations & Notes, " etc are bolded but still blend in too much with the text. Recommend making the text a little bigger and adding a horizontal line.

![image](https://github.com/user-attachments/assets/1fee3415-77aa-4edc-9709-f67c7f9b9c66)

Issue #8012, opened 2024-12-26
myHealth - add option to export examination(s)


jessewashburn:
Would be helpful for users who want to store their health records here. Exporting allows them to easily send them to a new doctor, print out proof of vaccinations, etc.

jessewashburn:
@Mutugiii Would you like me to work on this? I think I could create an export to pdf feature without too much difficulty.

jessewashburn:
@Mutugiii Would you like me to work on this? I think I could create an export to pdf feature without too much difficulty.

Mutugiii:
Sure, go ahead

jessewashburn:
fixed by #8043

PR #8043, opened 2025-01-06
myHealth: smoother examinations (fixes #8036, #8037, #8012)

### LINKED ISSUE #8036: myHealth: add profile info to the header of each exam
Should say something about who the test is regarding in the header next to the date , certainly if we were to export this. 

image

jessewashburn:
fixes #8036, #8037, #8012

Added export functionality. 

Examinations now contain user info.
image

Added bolded section labels and a conditional hr
image


Gavinp14 (approved):

RheuX (approved):

Mutugiii:
Not very sure about implementing #8036 fix since we might want the data sent anonymously, we need to discuss more on that

jessewashburn:
> Not very sure about implementing #8036 fix since we might want the data sent anonymously, we need to discuss more on that

I see. I thought Richard had requested we add the profile info to the exams, which is why I went in that direction.

PR #8048, opened 2025-01-07
all: activate alternative login (fixes #8047)

### LINKED ISSUE #8047: bypass auth service for some routes

Mutugiii:
Fixes #8047

Allow specific child routes to bypass the auth guard service i.e be accessible without authentication


Mutugiii:
Yes, we will be revamping it, but we might hold off on starting until we get some feedback from the gutamela team

dogi (approved):

PR #8115, opened 2025-01-17
mylife: smoother myachievements links (fixes #8114)

### LINKED ISSUE #8114: achievement link validation
ensure we have link validation for an achievement in the myAchievement page

Mutugiii:
Fixes #8114

Link validation for an achievement under the myAchievements


RheuX (commented):
I think its working well (for non http examples), but I did test out some edges case, something like this

image

image

image

It will allow it even though all of these are invalid links

Mutugiii:
@RheuX I've tackled the specific edge cases you mentioned

jessewashburn (changes requested):
It seems that anything after https:// or http:// (as long as it's not blank) is allowed. 

image

image


image

RheuX:
the edge case I mentioned is now blocked which is great,

i know we can't block every link that invalid, but probably another one is to check if at least have a common tld (https://tld-list.com/). so things like the one jesse mentioned shouldn't be possible. If they passed all the basic link validation, and have a tld, but just not a valid link (doesn't link to anywhere at the end), than thats something that we probably shouldn't handle unless you want to do a DNS Lookup

So TL;DR, I think we should implemented a regex that handle
- [x] no spaces
- [ ] only 1 period
- [ ] should contain a common tld at the end

this should be good enough for a basic validation link @Mutugiii

jessewashburn:
I pushed some additional edge case handling with a regex. All the cases I mentioned plus support for multipart TLDs like .co.uk should now be handled appropriately.

jessewashburn (approved):

dogi (approved):

PR #8129, opened 2025-01-22
mylife: smoother myhealth immunizations info (fixes #8126)

### LINKED ISSUE #8126: myHealth - immunizations rework
It doesn't make sense that `Immunization Dates` is a section in an examination, which happens on a single date. `Immunization(s)` should be the section in an examination (for example if the exam involved a flu shot). `Immunization Dates` should be a section in the profile that can be easily exported to send to a doctor.

jessewashburn:
fixes #8126


RheuX (approved):

Mutugiii:
There seems to be an issue with the alignment:

image



On Master it looks like this:
image

Mutugiii (approved):

PR #8152, opened 2025-01-27
courses: smoother chat persistence (fixes #8107)

### LINKED ISSUE #8107: course chat - conversation gets cleared when you hide course chat
It would be nice if the conversation stayed in memory until you exit the course.

RheuX:
Fixes #8107 

Now when you type something in course chat, it persist until you switch to another step, or when you navigate away.


RheuX:
also, im not sure, if we want to have the whole chat persist in the course, if you guys think we should do that, then let me know since right now, i erase the chat if you move on to another step, since i think closing and opening the window on the same step should persist, but im 50/50 on persisting it for the whole course

jessewashburn (approved):

Mutugiii (approved):

dogi (approved):

PR #8158, opened 2025-01-29
courses: smoother steps chat (fixes #8151)

### LINKED ISSUE #8151: course chat smaller display
Image

On a smaller display, having it split into 2 columns is not really that great, perhaps, we should have a seperate window open up? like a `bottom sheet`, so the learner can still interact with the chat, and then quickly navigate to the courses

RheuX:
Fixes #8151

image

Not sure how I feel about this, probabl would work on it more for a bit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.16.84",
+  "version": "0.16.87",
   "myplanet": {
-    "latest": "v0.22.51",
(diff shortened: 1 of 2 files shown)


RheuX:
i cant seem to find a better solution to make this look better, so i can take any feedback or advice

Mutugiii (approved):

Mutugiii:
Think we can merge for now since it is an improvement on the current view

dogi (approved):

PR #8228, opened 2025-02-12
community: smoother public access (fixes #8220)

### LINKED ISSUE #8220: community: make voices and calendar public
There's a desire to have some public facing content for communities to be able to promote their work.

When someone who is not logged in goes to the homepage (`/` route), they should not be rerouted to the login but instead a "community" page with only the voices and calendar tabs available. When they try to go to a different route they are sent to the login page before being directed to that route.

If possible, leave this behind a "beta" flag because there are additional steps needed to roll this out to all Planet instances.

Image

paulbert:
Makes the community page available to a not logged in client if the community is set to beta mode.

1. Set Planet to be in beta mode (advanced options in configuration).
2. Logout
3. Go to `localhost:3000` (or whatever your domain is)

Voices & Calendar can be viewed.

Images are not showing up in voices, this can be a follow up.
User menu in upper right is changed only have "Login" option in the dropdown. Could be a better UX to just have a login button. Also good for a follow up.

image

Fixes #8220

### DIFF
--- a/src/app/home/home-router.module.ts
+++ b/src/app/home/home-router.module.ts
@@ -1,7 +1,6 @@
 import { NgModule } from '@angular/core';
 import { Routes, RouterModule } from '@angular/router';
 import { DashboardComponent } from '../dashboard/dashboard.component';
-import { HomeComponent } from './home.component';
 import { NotificationsComponent } from '../notifications/notifications.component';
 import { UpgradeComponent } from '../upgrade/upgrade.component';
 import { UsersAchievementsComponent } from '../users/users-achievements/users-achievements.component';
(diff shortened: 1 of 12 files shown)


src/app/home/home-router.module.ts
```diff
@@ -12,70 +12,75 @@ import { CommunityComponent } from '../community/community.component';
...
 import { CoursesProgressLearnerComponent } from '../courses/progress-courses/courses-progress-learner.component';
 import { LandingComponent } from '../landing/landing.component';
+import { AuthService } from '../shared/auth-guard.service';
+import { BetaThenAuthService } from '../shared/beta-then-auth-guard-service';
 
 export function dashboardPath(route): string {
   return `${myDashboardRoute}/${route}`;
 }
 
+const alwaysGuardedRoutes = [
```
paulbert, line 22: I can revert this if needed. I think this is a little more organized but also I see it is cluttering the diff.

Mutugiii (approved):

dogi (approved):

PR #8241, opened 2025-02-14
teams: smoother survey submission (fixes #8238)

### LINKED ISSUE #8238: surveys: custom post survey form
Initially show the gender and the age.

Mutugiii:
Fixes #8238 

Focus on getting the gender and age post survey, hide other fields but can still be filled

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.17.16",
+  "version": "0.17.27",
   "myplanet": {
-    "latest": "v0.23.5",
-    "min": "v0.22.5"
+    "latest": "v0.23.29",
+    "min": "v0.22.29"
   },
   "scripts": {
     "ng": "ng",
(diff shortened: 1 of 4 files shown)


jessewashburn (changes requested):
image
Getting some weird results. I did 3 surveys using age 22 (although a random character may have crept in there resulting in NAN. I'd double check for validation.), 100, and 1. It seems to be showing birthdate instead of age.

Mutugiii:
Nice catch! Needed to handle the age value on exporting. You'll need to retest and disregard the already added submissions

Not sure how you managed to get 1925 though would be interested to know if you can replicate it again

jessewashburn (changes requested):
Age is handled appropriately, but NAN values are still allowed. 

> Nice catch! Needed to handle the age value on exporting. You'll need to retest and disregard the already added submissions
> 
> Not sure how you managed to get 1925 though would be interested to know if you can replicate it again

image
image
image

jessewashburn (approved):

jessewashburn:
It works great, but do we want to validate the form to prevent non-numbers in the age section? Maybe that's for another pull request.

Issue #8225, opened 2025-02-11
community: Prompt user to login from community click to other content


paulbert:
depends on #8220 

When the community page is public, the user should be prompted to login before being routed to other content in Planet if they click on any content that requires authorization.

The two options for implementation is redirect or dialog.

Redirect
- Simpler to route
- May be some complication making sure the user is redirected to the correct post login content

Dialog
- Might be a cleaner user experience
- Would require creating a new login dialog, not sure if that would work well with post login routing

Mutugiii:
https://github.com/open-learning-exchange/planet/issues/8283#issue-2885478378

paulbert:
Assigned myself because I'm doing some explorations into the login dialog option.

PR #8275, opened 2025-02-26
community: smoother voice url (fixes #8226)

### LINKED ISSUE #8226: voices: Link directly to specific posts
Add routing to allow a user to paste in a link with something like `/voices/{id}` and be directed to the community page with that specific post as if they had clicked "See replies" on that post.

Image

Mutugiii:
Fixes #8226

- Change the routing structure of voice replies to match 'voice/:id'
- Icon to copy the voice and share that link

image


jessewashburn (changes requested):
The linking works, but replying and editing messages require a refresh to see the changes. Also, deleting a reply doesn't seem to work at all. Eventually we should add a "Create link" button, but that could be a separate PR down the road.

[screen-recorder-wed-feb-26-2025-16-14-30.webm](https://github.com/user-attachments/assets/684d3cc2-e6ca-4ac5-b7a9-472aaa1f1234)

Mutugiii:
Technically those actions from non-authed users shouldn't be allowed and we should redirect for login (or a popup for login)  -> #8225

jessewashburn:
> Technically those actions from non-authed users shouldn't be allowed and we should redirect for login (or a popup for login) -> #8225

Ah I see

jessewashburn (approved):

src/app/shared/auth-guard.service.ts
```diff
@@ -43,15 +43,27 @@ export class AuthService {
     );
   }
 
+  private requiresNoAuth(route: ActivatedRouteSnapshot): boolean {
```
paulbert, line 46: This function flips the current logic which previously traveled up the route's parent along the tree to now going down the route's children. We should keep it to parents because allowing a child route to overwrite a parent's auth requirement is an easy way to end up with unintended side effects when opening up pages.

src/app/home/home-router.module.ts
```diff
@@ -76,8 +77,14 @@ const alwaysGuardedRoutes = [
 ];
 
 const routes: Routes = [
-  { path: '', component: CommunityComponent, canActivate: [ BetaThenAuthService ] },
-  { path: '',
+  {
+    path: '',
+    component: CommunityComponent,
+    children: [ { path: 'voices/:id', component: NewsListComponent } ],
```
paulbert, line 83: This should be added to `backgroundRoute` so the black to blue background doesn't go away when a voices reply is shown.

paulbert (changes requested):
Looks like a good start. Still need to implement the initial state for a linked voices post. If I paste in a copied link it shows the main news page instead of the specific post.

src/app/news/news-list.component.ts
```diff
@@ -41,9 +42,23 @@ export class NewsListComponent implements OnChanges {
...
+    private router: Router,
+    private route: ActivatedRoute
   ) {}
 
+  ngOnInit() {
+    const childRoute = this.route.firstChild;
+    if (childRoute) {
+      const voiceId = childRoute.snapshot.paramMap.get('id');
+      if (voiceId) {
+        this.newsService.requestNewsItem(voiceId).subscribe(news => {
```
paulbert, line 55: We already fetch all of the news on the CommunityComponent. That's sent to this component in `items` so you can just find it there rather than making another request.

paulbert (approved):
Looks good just one more change to remove the extra request to the news database.

One thing to follow up on is now it seems a little odd to be on one of the other community tabs but still have a "voices" route, but we can tackling having routes for all of the different states the community page could be in on another time.

Screenshot from 2025-03-28 11-11-54

dogi (approved):

PR #8303, opened 2025-03-03
community: smoother authentication dialogs (fixes #8284)

### LINKED ISSUE #8284: community first page: disable ...
... voice comments
and add events

Mutugiii:
Fixes #8284

Authenticate the dialog actions for the open community page
- Reply
- Add Event
- Feedback

### DIFF
--- a/src/app/feedback/feedback.directive.ts
+++ b/src/app/feedback/feedback.directive.ts
@@ -147,25 +147,25 @@ export class FeedbackDirective {
   }
 
   @HostListener('click')
+  checkAuthentication() {
+    this.authService.checkAuthenticationStatus().subscribe(this.openFeedback);
+  }
+
   openFeedback() {
-    this.authService.isAuthenticated().subscribe(isAuthenticated => {
-      if (isAuthenticated) {
-        const title = $localize`Feedback`;
-        const type = 'feedback';
-        const fields = dialogFieldOptions;
-        const formGroup = {
-          priority: [ this.priority, Validators.required ],
-          type: [ this.type, Validators.required ],
-          message: [ this.message, CustomValidators.required ]
-        };
-        this.dialogsFormService
-          .confirm(title, fields, formGroup)
-          .pipe(debug('Dialog confirm'))
-          .subscribe((response) => {
-            if (response !== undefined) {
-              this.addFeedback(response);
-            }
-          });
+    const title = $localize`Feedback`;
+    const type = 'feedback';
+    const fields = dialogFieldOptions;
+    const formGroup = {
+      priority: [ this.priority, Validators.required ],
+      type: [ this.type, Validators.required ],
+      message: [ this.message, CustomValidators.required ]
+    };
+    this.dialogsFormService
+      .confirm(title, fields, formGroup)
+      .pipe(debug('Dialog confirm'))
+      .subscribe((response) => {
+        if (response !== undefined) {
+          this.addFeedback(response);
         }
       });
   }
--- a/src/app/feedback/feedback.directive.ts
+++ b/src/app/feedback/feedback.directive.ts
@@ -9,6 +9,7 @@ import { PlanetMessageService } from '../shared/planet-message.service';
 import { debug } from '../debug-operator';
 import { StateService } from '../shared/state.service';
 import { CustomValidators } from '../validators/custom-validators';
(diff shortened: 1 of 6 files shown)


jessewashburn (approved):

jessewashburn:
Works well. The only other thing maybe worth restricting is the profile info popup on voices, as I'm not sure you want users' phone numbers and emails visible to those who aren't logged in.  

image

Mutugiii:
Nice catch!

Updated to restrict that dialog as well

src/app/feedback/feedback.directive.ts
```diff
@@ -146,20 +148,24 @@ export class FeedbackDirective {
...
-      type: [ this.type, Validators.required ],
-      message: [ this.message, CustomValidators.required ]
-    };
-    this.dialogsFormService
-      .confirm(title, fields, formGroup)
-      .pipe(debug('Dialog confirm'))
-      .subscribe((response) => {
-        if (response !== undefined) {
-          this.addFeedback(response);
+    this.authService.isAuthenticated().subscribe(isAuthenticated => {
```
paulbert, line 151: It might be a little more readable or quickly understandable that we're checking authentication here if the listener function is a `checkAuthentication` and we move the open feedback logic to a separate function that is called here.

I'm thinking something like:

```
@HostListener('click')
checkAuthentication() {
  this.authService.isAuthenticated().subscribe(this.openFeedback); // might need to be () => this.openFeedback(); for "this" context
}

openFeedback() {
  ...
}
```

src/app/feedback/feedback.directive.ts
```diff
@@ -146,20 +148,24 @@ export class FeedbackDirective {
...
+      if (isAuthenticated) {
```
paulbert, line 152: See AuthService but I think we should avoid all these ifs.

src/app/shared/auth-guard.service.ts
```diff
@@ -70,4 +70,17 @@ export class AuthService {
     );
   }
 
+  isAuthenticated(): Observable<boolean> {
```
paulbert, line 73: Not sure what's best but I think `checkAuthentication` or `handlerAuthGuard` better describes this.

src/app/shared/auth-guard.service.ts
```diff
@@ -70,4 +70,17 @@ export class AuthService {
     );
   }
 
+  isAuthenticated(): Observable<boolean> {
+    return this.getSession$().pipe(
+      take(1),
```
paulbert, line 75: Is this needed? I'm not seeing this return more than one value.

src/app/shared/auth-guard.service.ts
```diff
@@ -70,4 +70,17 @@ export class AuthService {
     );
   }
 
+  isAuthenticated(): Observable<boolean> {
+    return this.getSession$().pipe(
+      take(1),
+      map(sessionInfo => {
+        const isLoggedIn = !!sessionInfo.userCtx.name;
+        if (!isLoggedIn) {
+          this.router.navigate([ '/login' ], { queryParams: { returnUrl: this.router.url } });
```
paulbert, line 79: I think we should throw an error here. A simple "Not authorized" message would probably work. This will break the observable and avoid running the subscription function when we call this. Makes it so we don't need to check if the user is logged in every time this gets called.

src/app/shared/auth-guard.service.ts
```diff
@@ -70,4 +70,17 @@ export class AuthService {
...
 
+  isAuthenticated(): Observable<boolean> {
+    return this.getSession$().pipe(
+      take(1),
+      map(sessionInfo => {
+        const isLoggedIn = !!sessionInfo.userCtx.name;
+        if (!isLoggedIn) {
+          this.router.navigate([ '/login' ], { queryParams: { returnUrl: this.router.url } });
+        }
+        return isLoggedIn;
```
paulbert, line 81: If we throw an error above, no need to return a value here.

paulbert (commented):
Looks like it's working as intended, but adding some thoughts on things that could improve.

paulbert (approved):
Looks good!

I'm assuming the plan is to follow up with a PR for the login dialog?

Mutugiii:
yes, that is the plan!

dogi (approved):

dogi (approved):

PR #8318, opened 2025-03-10
community: smoother login dialog (connects #8225)

paulbert:
WIP branch for using a dialog for the login form instead of a separate route.

#8225


jessewashburn (approved):

jessewashburn (changes requested):
If configuration type is set to nation, login prompt still works, but I get the below snackbar message. Also, I can't get into manager settings even though I'm on an admin account. The rest of the functionality seems to work fine. 

[screen-recorder-mon-mar-24-2025-09-44-24.webm](https://github.com/user-attachments/assets/bd2dc54d-fa31-4d04-be77-dfae5840e451)

Mutugiii:
@jessewashburn the snackbar is not directly related to the login but rather to do with your dev community not being able to connect to the planet earth

jessewashburn:
> @jessewashburn the snackbar is not directly related to the login but rather to do with your dev community not being able to connect to the planet earth

I see. It's just that I don't see the snackbar when I log in any other way. Also, there's the issue where I don't have manager permissions when I log in this way.

paulbert:
Updating the toolbar brings up a second issue where some pages that rely on StateService subscriptions don't load.

To replicate login via the dialog then open up a course or go into manager settings and open the list of users. Content will not load but on refresh it'll load up.

jessewashburn (approved):

dogi (approved):

munishpatel (approved):

PR #8345, opened 2025-03-14
manager: smoother reports courses overview export (fixes #8329)

### LINKED ISSUE #8329: manager: smoother report summary courses export
Image

jessewashburn:
fixes #8329
image
image
image

The link builder doesn't work on a dev environment, but I think it should be functional on Planet Guatemala for example. Please let me know if that is not the case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.17.69",
+  "version": "0.17.74",
   "myplanet": {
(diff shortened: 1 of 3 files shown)


munishpatel (commented):
image

Works fine in my dev environment, course overview reports is visible and when clicked on it select date range also working fine. After selecting the date range and clicking on submit, it automatically download a .csv file. 

Nice!

munishpatel (approved):

PR #8380, opened 2025-03-21
all: replace ngx img package (fixes #8379)

### LINKED ISSUE #8379: replace ngx-img package
`ngx-img` is no longer being maintained, is no longer in npm package directory as well and was tested upto ng 9.1.9.

Given plans to upgrade to ng 12+, we need to replace ngx-img

Mutugiii:
Fixes #8379

Switch to lightweight and maintained [ngx-image-cropper](https://www.npmjs.com/package/ngx-image-cropper) package.


3011089503 (approved):
Good change! And the ngx-img has some differences with ngx cropper. In the future we should solve some potential issues of that.  Make sure the benefits are solid enough before we merge it.

jessewashburn (approved):

paulbert (approved):
Overall I think the code is good and the cropper is working as intended, so I'll approve!

See the one comment about button text. Also the layout isn't great. All of the white space on the right and bumping the image preview lower isn't great. Maybe having the cropper in a dialog instead of on the page directly would be better? I think that is good to decide on and do in a follow up.

image

dogi (approved):

PR #8462, opened 2025-04-03
community: smoother url switching tabs (fixes #8443)

### LINKED ISSUE #8443: community tabs url
In the community page, after viewing the replies of a voice, navigating the community page, switching between tabs should clear any extra URL segments. 

Steps to reproduce:
- Go to a voice with a reply and view the replies
- Go to another tab e.g `Community Leaders`

Image

munishpatel:
fixes: #8443 

https://github.com/user-attachments/assets/10f03ea8-991b-45b0-ac57-8dfa6c968b5c

The community url is working fine now!


Mutugiii (changes requested):
Currently this only works when you switch to the `Community Leaders` tab only while we want it to work for all the other tabs in the community page

Also when switching back to the `Our Voices` tab, it currently shows the reply but the url route is empty. we should either show the relevant url route for the reply voice or go to the main conversation

munishpatel (commented):
Now it work sfor all the other tabs in the community page.

Need help when switching back to the Our Voices tab.

Mutugiii (approved):

dogi (approved):

PR #8491, opened 2025-04-08
teams: smoother join requests and applicants (fixes #8481)

### LINKED ISSUE #8481: teams: combine join requests and members tabs
Combine the 'Join Requests' tab with the 'Members' tab

munishpatel:
fixes #8481 

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.18.6",
+  "version": "0.18.34",
   "myplanet": {
     "latest": "v0.24.42",
     "min": "v0.23.42"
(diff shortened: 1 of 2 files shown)


jessewashburn (changes requested):
I would recommend putting the Join Requests section above the members. Otherwise, team leaders have to scroll down to see requests and might miss them. 

Since that section is conditional on having data anyway, it won't get in the way when there are no requests. What do you think @Mutugiii ?

Mutugiii:
makes sense @jessewashburn

munishpatel (commented):
Screenshot 2025-04-15 at 9 29 37 PM

Made the required changes as mentioned @jessewashburn @Mutugiii

jessewashburn (approved):

dogi (approved):

PR #8497, opened 2025-04-09
community: smoother new landing page (fixes #8482)

### LINKED ISSUE #8482: community: public community release (remove beta)
Time to release the public community page & remove the beta checks which redirect to the community page or login page. Also removing beta checks for the login dialog.

paulbert:
Removes the beta requirement for the public community. I left in the `BetaThenAuthComponent` which was only used for this, but I think we should wait to remove that until we are sure we do not need to revert this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.18.11",
+  "version": "0.18.16",
   "myplanet": {
     "latest": "v0.24.45",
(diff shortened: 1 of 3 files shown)


paulbert:
fixes #8482 

I'm open to merging after #8498 but I think this could also wait a week.

jessewashburn (approved):

Mutugiii (approved):

dogi (approved):

PR #8504, opened 2025-04-10
manager: smoother myplanet reports date filtering (fixes #8502)

### LINKED ISSUE #8502: myPlanet report: add date range dropdown

jessewashburn:
fixes #8502

Added a dropdown menu for date filtering to match the other reports pages. Currently, All Time is the default range, but this can be adjust if something else makes more sense or is more convenient. 

image
image

### DIFF
--- a/src/app/manager-dashboard/reports/reports-myplanet.component.ts
+++ b/src/app/manager-dashboard/reports/reports-myplanet.component.ts
@@ -42,6 +42,15 @@ export class ReportsMyPlanetComponent implements OnInit {
   versions: string[] = [];
   selectedVersion = '';
   disableShowAllTime = true;
+  selectedTimeFilter = 'all'; // Default to All Time
+  timeFilterOptions = [
+    { value: '24h', label: $localize`Last 24 Hours` },
+    { value: '7d', label: $localize`Last 7 Days` },
+    { value: '30d', label: $localize`Last 30 Days` },
+    { value: 'all', label: $localize`All Time` },
+    { value: 'custom', label: $localize`Custom` },
+  ];
+  showCustomDateFields = false;
 
   constructor(
     private csvService: CsvService,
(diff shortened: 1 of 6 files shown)


src/app/manager-dashboard/reports/reports-myplanet.component.ts
```diff
@@ -42,6 +42,15 @@ export class ReportsMyPlanetComponent implements OnInit {
...
   selectedVersion = '';
   disableShowAllTime = true;
+  selectedTimeFilter = 'all'; // Default to All Time
+  timeFilterOptions = [
+    { value: '24h', label: $localize`Last 24 Hours` },
+    { value: '7d', label: $localize`Last 7 Days` },
+    { value: '30d', label: $localize`Last 30 Days` },
+    { value: 'all', label: $localize`All Time` },
+    { value: 'custom', label: $localize`Custom` },
+  ];
```
Mutugiii, line 52: Since we have a similar block of code in the myPlanet logs, it would be ideal to place it in a constanst.ts file and import to both myPlanet logs & reports

src/app/manager-dashboard/reports/reports-myplanet.component.ts
```diff
@@ -149,6 +160,44 @@ export class ReportsMyPlanetComponent implements OnInit {
...
+    }
+    this.startDate = newStartDate;
+    this.endDate = newEndDate;
+    this.reportsForm.patchValue({
+      startDate: newStartDate,
+      endDate: newEndDate,
+    });
+    this.applyFilters();
+  }
+
```
Mutugiii, line 200: Similar to above, since we have a similar block of code in the myPlanet logs, this could be refactored and placed in the reports service and called in both places

Mutugiii (approved):

dogi (approved):

PR #8513, opened 2025-04-10
all: smoother content loading (fixes #8500)

### LINKED ISSUE #8500: various - wait for data to load before populating toolbar
Library, Teams, likely others

jessewashburn:
fixes #8500

Anywhere there is a counter or some section where the presence of data affects the placement of elements, we now wait for the data before displaying these elements. This prevents inaccurate info from flashing on the screen or elements shifting once the data fully loads.

### DIFF
--- a/src/app/courses/courses.component.html
+++ b/src/app/courses/courses.component.html
@@ -42,28 +42,30 @@
 
 <div [ngClass]="{ 'space-container': !isForm }" class="primary-link-hover">
   <mat-toolbar class="primary-color font-size-1" *ngIf="!isDialog && !isForm">
-    <ng-container *ngIf="isAuthorized">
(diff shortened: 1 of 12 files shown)


jessewashburn:
To test, you can go to all of these page and let them load with network throttling so you have chance to see the change. Places to test: 

courses list, courses view, resources list , resources view, dashboard, manager dashboard, exam or survey view, teams view.

munishpatel (approved):
Nice work!

3011089503 (approved):

Mutugiii:
Not sure we want this particularly in the courses and the resources where we have disabled buttons and an items count. The buttons and count provide visual cues to users which can be especially helpful if the resources/courses are taking time to load

PR #8517, opened 2025-04-11
mylife: smoother achievements loading (fixes #8518)

### LINKED ISSUE #8515: achievements: certifications takes longer to load and pushes other fields around
Either make it faster so it can load with everything else or at least move it to the bottom so it doesn't push the achievements around

https://github.com/user-attachments/assets/5ea568e7-a0fc-4780-9168-2d8d5de1e212

### LINKED ISSUE #8518: dashboard: myPersonal seems to be obsolete
As far as I can tell, the resources page linked to via myPersonals on the dashboard will not have any resources.

The "myPersonals" version of the resources page filters by:

```
{
  private: true,
  privateFor: { users: *user._id* }
}
```

The two cases where we set a `privateFor` field are when adding a resource to a team or adding an image to a message on the teams message board. In both cases, `privateFor` is set to `{ teams: *team._id* }`.

Not sure if we should remove this, but just wanted to bring it up for consideration.

jessewashburn:
fixes #8515

The whole achievements page loads at once, instead of prematurely loading without certifications.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.18.14",
+  "version": "0.18.27",
   "myplanet": {
     "latest": "v0.24.45",
     "min": "v0.23.45"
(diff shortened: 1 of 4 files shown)


munishpatel (approved):
image

Mutugiii (approved):

PR #8578, opened 2025-04-23
dashboard: smoother tile visibility (fixes #8577)

### LINKED ISSUE #8577: myDashboard: distinguish white tiles from background
I recommend using a slightly different shade of white or maybe a subtle border so the last white tile in a row is more clearly visible. 

Image

jessewashburn:
fixes #8577

If the very last tile in the dashboard is white, I added a subtle right border to prevent it from blending in with the background. 

Image

The linter also prompted me to make this unrelated fix to users-achievements.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.18.46",
+  "version": "0.18.50",
   "myplanet": {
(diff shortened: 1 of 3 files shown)


3011089503 (approved):
截屏2025-04-25 下午3 24 56

Mutugiii (approved):

PR #8586, opened 2025-04-24
manager: smoother reports export (fixes #8583)

### LINKED ISSUE #8583: manager: exports diabled when start date is before min date
There is data in this range, so it should be exportable. 

Image

jessewashburn:
fixes #8583

I removed the unnecessary min date export restriction. Now even if the start date is before the earliest data, it will still export data if there is data in that range.


munishpatel (approved):

3011089503 (approved):

Mutugiii:
Only issue with this approach is that it allows selection of any date.

image

PR #8591, opened 2025-04-25
manager: smoother configuration (fixes #8590)

### LINKED ISSUE #8590: implement a safeConfig option
Currently the configurations db object has the risk of exposing the chat keys if incorrectly logged.

Mutugiii:
Fixes #8590

Safer config to prevent unintended logging of keys


jessewashburn (approved):

3011089503 (commented):
The achievement file seems like making the same change as in 8588. I am not sure if it will have any impact.
截屏2025-04-25 下午2 25 52

dogi (approved):

PR #8604, opened 2025-04-28
reports: add PDF button(fix #8603)

### LINKED ISSUE #8603: reports: the report can also be exportable as a PDF
Image

3011089503:
fix #8603
截屏2025-04-28 下午12 57 31


munishpatel (approved):
image

Works fine!! Good

PR #8616, opened 2025-05-01
community: implement news pagination (fixes #8608)

### LINKED ISSUE #8608: community: add news view
Since the community is now the landing page we should add pagination to the voices for a better user experience

Will initially start with a view for the data

Mutugiii:
Fixes #8608

Add news db view for voices.

To test in the local dev env first run:  `./couchdb-setup.sh -u adminuser -w adminpassword`.  Then test the voices

To test using a production system:

```
docker pull treehouses/planet-tags:
docker pull treehouses/planet-tags:db-init-
```


design/news/news-design.js
```diff
@@ -0,0 +1,39 @@
+module.exports = {
+  "views": {
+    "by_section_and_id_and_date": {
+      map: function (doc) {
+        if (doc.docType !== 'message' || !Array.isArray(doc.viewIn)) { return; }
+
+        payload = {
```
paulbert, line 7: Should be a `var` here (CouchDB design docs use older JS syntax before `let` and `const`). Seems like it's working without it, but we should have that in there to avoid unintended behavior.

src/app/community/community.component.ts
```diff
@@ -57,7 +57,8 @@ export class CommunityComponent implements OnInit, OnDestroy {
   resizeCalendar: any = false;
   deviceType: DeviceType;
   deviceTypes = DeviceType;
-  isLoading: boolean;
+  isLoading = true;
```
paulbert, line 60: Let's have these variables be `isLoadingInit` and `isLoadingMore`. Gives a little more clarity on the two to have both be verbose.

paulbert (changes requested):
Good start! Needs some work, though:

- The amount of posts shown on each "Show More" is inconsistent, likely due to the replies getting lumped in with all the posts. The request should only pull posts at the level being viewed, or when `replyTo` equals null for toplevel or the post id when viewing replies to a post.
- When adding a new post it doesn't show up anymore after submitting the form. This should be fixed.
- When all posts are loaded the "Show More" button should be hidden.
- Going from the community to a team page throws an error (see below). I wonder if the view could be done as a filter function? `[ section, id, date ]` is a pretty complex key to be indexing off of. We set up an index on simply `time`, and it'd be ideal to use that. Doing a filter should fix the first point.
- We should use the `time` field instead of `updatedDate`. `updatedDate` is just there to show when a post is updated but we sort them based on the time the post was created.

Error on teams page:
image

Mutugiii:
> * The amount of posts shown on each "Show More" is inconsistent, likely due to the replies getting lumped in with all the posts. The request should only pull posts at the level being viewed, or when `replyTo` equals null for toplevel or the post id when viewing replies to a post.

With the startKey & endKey queryString in place it is proving tricky to ensure the consistency of the number of responses without meaningfully modifying the current view.

Opened a different PR to initially tackle the UI pagination as we discuss on potentially moving to a simpler time-based index. #8652

Mutugiii:
Closing in favour of #8652

PR #8618, opened 2025-05-01
actions: smoother release prebuild on change (fixes #8529)

### LINKED ISSUE #8529: build: prebuilder not running
Prebuild is not running on any PRs or commits currently. This is the step where the `npm install` runs for packages needed in the build process.

Two things need to be implemented ideally:
1. Prebuild should run if the packages change. Could be difficult since we change the version on every commit (and even if we didn't a version change may not mean a change in packages). The prebuild step is long so it's important to only run it when necessary.
2. PR builds should not push to the `planet-base:latest` docker image, that should be reserved for commits. PR builds should also use their own `planet-base` image if built or fall back to the latest build if not.

paulbert:
WIP update the prebuilder script to run if new packages are added or packages are updated.

fixes #8529


.github/workflows/planet.yml
```diff
@@ -85,6 +99,16 @@ jobs:
       - name: Docker login
         run: docker login -u ${{ secrets.DOCKERUSERNAME }} -p ${{ secrets.DOCKERAPIKEY }}
 
+      - name: Tag prebuild as latest
+        run: |
```
paulbert, line 103: Might need this to conditionally only run if NOT master

paulbert:
After this merges I can create a follow up with a small package change to confirm that this works as intended on master when upgrading or adding packages.

Mutugiii (approved):

3011089503 (approved):

dogi (approved):

PR #8632, opened 2025-05-05
courses: smoother editing (fixes #8631)

### LINKED ISSUE #8631: courses: put general course info in a collapsed container when editing
Similar to editing surveys, the general info like title and description, which is relatively rarely edited, should be in a collapsed container to allow more screen real estate to the steps, which are edited more frequently.
Image

jessewashburn:
fixes #8631

Added a collapsible container for course title and description and moved the steps below the container instead of cramming them in a separate column to the right.

If editing a course with steps, the container is collapsed by default, as the steps are more likely to be edited
image

When you add a new course or edit a course with no steps, the container is expanded by default. 
image

### DIFF
--- a/src/app/courses/add-courses/courses-add.component.html
+++ b/src/app/courses/add-courses/courses-add.component.html
@@ -15,7 +15,7 @@
     <mat-expansion-panel [expanded]="isFormExpanded" class="course-details-panel">
       <mat-expansion-panel-header>
         <mat-panel-title i18n class="course-title-header">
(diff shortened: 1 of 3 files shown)


JunhaoZhou66 (changes requested):
That's a good idea. There's just a small bug — these two clicks trigger the same function.
截屏2025-05-07 上午11 37 15

Mutugiii:
Closing in favour of #9978

PR #8640, opened 2025-05-06
teams: smoother list sorting priority (fixes #8635)

### LINKED ISSUE #8635: teams: sort for joined teams
In the teams/enterprises, we need a sort by a user's team.(Similar to the sort by name, visits, type). This could be placed on top of the action button.

Image

3011089503:
fixes #8635

https://github.com/user-attachments/assets/2b4fae62-dbc2-455e-b978-790554a22b99

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.18.61",
+  "version": "0.18.93",
   "myplanet": {
-    "latest": "v0.25.13",
-    "min": "v0.24.13"
+    "latest": "v0.25.47",
+    "min": "v0.24.47"
   },
   "scripts": {
     "ng": "ng",
--- a/src/app/teams/teams.component.html
+++ b/src/app/teams/teams.component.html
@@ -84,7 +84,7 @@
       <ng-container matColumnDef="action">
         <mat-header-cell *matHeaderCellDef i18n>
           <span>Action</span>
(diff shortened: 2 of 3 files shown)


jessewashburn:
Instead of sort, I think a filter button would be more intuitive. "Show My Teams" and "Show All Teams". What do you think @Mutugiii

jessewashburn (changes requested):
Instead of sort, I think a filter button would be more intuitive. "Show My Teams" and "Show All Teams". What do you think @Mutugiii ?

3011089503:
> Instead of sort, I think a filter button would be more intuitive. "Show My Teams" and "Show All Teams". What do you think @Mutugiii ?

I think both designs work.  But this one requires a bit more logic.

jessewashburn (approved):

3011089503 (commented):
@Mutugiii Is this position the same as the one you originally mentioned?

Mutugiii:
@jessewashburn Given we already have a myTeams route accessible via myDashboard we can go ahead with just a sort on the teams/enterprises.

@3011089503 The initial idea was to have the button on the table header row similar to the table header filter buttons.

jessewashburn:
@Mutugiii another option might be to have a "Joined" column that says yes or no. Then the user can just use the already familiar table sorting function and sort by that column.

munishpatel (approved):

Mutugiii (approved):

dogi (approved):

dogi (approved):

PR #8652, opened 2025-05-07
teams: smoother voices pagination (fixes #8608)

### LINKED ISSUE #8608: community: add news view

Mutugiii:
Fixes #8601 

Add news UI pagination to the community and teams components

### DIFF
--- a/src/app/community/community.component.html
+++ b/src/app/community/community.component.html
@@ -16,7 +16,7 @@
               <div class="news-container">
                 <planet-news-list [items]="paginatedNews" [shareTarget]="shareTarget" (viewChange)="toggleShowButton($event)" [viewableId]="teamId"></planet-news-list>
               </div>
-              <div class="full-width" style="text-align: center;" *ngIf="!isLoadingInit && !isLoadingMore && hasMoreNews && !this.newsService.getActiveReplyId()">
+              <div class="full-width" style="text-align: center;" *ngIf="!isLoadingInit && !isLoadingMore && hasMoreNews && !isActiveReplyThread">
                 <button mat-raised-button color="primary" (click)="loadMoreNews()">
                   <span i18n>Load More</span>
                 </button>
--- a/src/app/community/community.component.html
+++ b/src/app/community/community.component.html
@@ -16,7 +16,7 @@
               <div class="news-container">
                 <planet-news-list [items]="paginatedNews" [shareTarget]="shareTarget" (viewChange)="toggleShowButton($event)" [viewableId]="teamId"></planet-news-list>
               </div>
-              <div class="full-width" style="text-align: center;" *ngIf="!isLoadingInit && !isLoadingMore && hasMoreNews && !this.newsService.getActiveReplyId()">
+              <div class="full-width" style="text-align: center;" *ngIf="!isLoadingInit && !isLoadingMore && hasMoreNews && !newsService.getActiveReplyId()">
                 <button mat-raised-button color="primary" (click)="loadMoreNews()">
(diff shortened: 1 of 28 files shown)


src/app/community/community.component.html
```diff
@@ -8,19 +8,29 @@ <h3 style="text-align: right; margin-right: 0.5rem;">
...
+          <ng-container *ngIf="isLoadingInit; else loaded">
             <span i18n>Loading voices...</span>
           </ng-container>
           <ng-template #loaded>
             <ng-container *ngIf="news?.length > 0; else noVoices">
-              <planet-news-list [items]="news" [shareTarget]="shareTarget" (viewChange)="toggleShowButton($event)" [viewableId]="teamId"></planet-news-list>
+              <div class="news-container">
+                <planet-news-list [items]="paginatedNews" [shareTarget]="shareTarget" (viewChange)="toggleShowButton($event)" [viewableId]="teamId"></planet-news-list>
+              </div>
+              <div class="full-width" style="text-align: center;" *ngIf="!isLoadingInit && !isLoadingMore && hasMoreNews && !isActiveReplyThread">
```
paulbert, line 19: Let's make a follow up issue to make this a shared component.

Also not sure why we're not showing this for viewing replies? Is that just to push this out or is there another reason that we wouldn't limit the posts viewed for replies?

paulbert (commented):
Looks like you're actively working on this, but just wanted to add the comment below.

At one point the deep linking to view replies to a post did not work. Not sure if it's because the app had just recompiled or if it was actually an issue. I'm not able to reproduce but just wanted to note that so you keep an eye on that feature.

jessewashburn (changes requested):
I'm not seeing the load more button after 10 voices, even though there are more that aren't displaying.

src/app/community/community.component.html
paulbert, line 19: The function needs to be called for this to work properly (`!isactiveReplyThread()`).

I think it's not optimal for rendering reasons to call a function in an `ngIf`. Better to have a variable because Angular can detect when that changes and optimize when it runs a rerender.

paulbert:
Just a thought: there is the Angular Material pagination component. We could implement that on `NewsListComponent` instead of writing our own. That should work smoothly with the reply lists and there should be an easy way to implement linking directly to a page for easier navigation to older posts.

I think this is working pretty well, too, but it's also adding a bit of complexity and duplication which we might be able to avoid with the pagination component.

Mutugiii:
Updated the implementation to use the news list component which makes it easier to track the replies pagination by leveraging the displayedItems array. This should be cleaner than the previous one.

jessewashburn (approved):

src/app/news/news-list.component.ts
```diff
@@ -197,4 +195,35 @@ export class NewsListComponent implements OnInit, OnChanges {
...
+    const page = list.slice(start, end);
+    return {
+      items: page,
+      endIndex: start + page.length,
+      hasMore: end < list.length
+    };
+  }
+
+  loadPagedItems(initial = true) {
+    const news = this.getCurrentItems();
```
paulbert, line 216: Would be a little better to set this to a variable (`this.filteredNews`, maybe?) rather than running this filter every time "Show More" is clicked, I think.

paulbert (approved):
Looks like a good first step! I noticed some follow up issues for improvements, but I don't want to hold this up any longer. Hopefully it'll improve performance on the community page.

PR #8678, opened 2025-05-13
all: smoother the logout function (fixes #8547)

### LINKED ISSUE #8457: reports: dynamically load the charts based on available data

### LINKED ISSUE #8547: navbar: logout should route to community page
Now that the public community page is out of beta, we can update the logout button to route to the community page instead of the login page.

Some things to consider:

- If the user is on a voices post, try to maintain that url instead of routing to `/`.
- If the user is on the community page, we should make sure the private tabs (i.e. "Community Leaders", "Services", etc.) are removed from view. If the user was actively on one of those tabs make sure it transitions to voices after removal.

3011089503:
fixes #8547

Uploading 录屏2025-05-13 下午1.16.54.mov…

### DIFF
--- a/src/app/home/home.component.ts
+++ b/src/app/home/home.component.ts
@@ -188,26 +188,38 @@ export class HomeComponent implements OnInit, DoCheck, AfterViewChecked, OnDestr
 
   logoutClick() {
     const configuration = this.stateService.configuration;
+  
     const errorCatch = error => {
-      console.log(error);
+      console.log('Logout error:', error);
       return of({});
     };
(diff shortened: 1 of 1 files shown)


Mutugiii (changes requested):
A good start, but currently I'm not being rerouted to the community page
image


Also a few things to to point out with regards to the PR format:
- Looks like you referenced the wrong issue in the title & description
- Your demo video didn't fully upload
- Watch out for unnessary whitespaces and indentation changes

Mutugiii:
Closing in favour of #8700

Issue #8695, opened 2025-05-21
Prebuild not running on master


paulbert:
The script to run prebuild to build the base image with new npm packages is not running on the master branch, only on PR builds.

Need to update the script so it correctly detects changes for package.json and runs on master branch.

Also consider readding functionality to manually run an update of the `planet-base:latest` image.

paulbert:
Fixed in #8696

PR #8696, opened 2025-05-21
actions: smoother release prebuild

paulbert:
Makes a few changes for the prebuild release:

- When merged to master, looks back one commit to compare package.json against previous master instead of current master
- Adds a condition to allow the prebuild step to run manually again
- Rename the script to a more descriptive script name

It also reverts the ramda package to the previous version which should trigger the prebuild again.


paulbert:
Fixes #8695

dogi (approved):

PR #8726, opened 2025-05-29
community: smoother voices creation (fixes #8721)

### LINKED ISSUE #8721: voices: make New Voice button more prominent
Elsewhere on the app, adding a course, resource, survey, etc is represented by an eyecatching yellow '+' button. I like the 'New Voice' label for branding, but maybe the button could be yellow to catch the eye and encourage more posts. 

Image

3011089503:
fixes #8721
截屏2025-05-29 下午2 06 58


jessewashburn (changes requested):
I think this looks good, but you should use the variable from _variables.scss instead of hard-coding the color. The variable for this yellow color is called "accent".

3011089503 (commented):
I have created a new commit by changing the scss and HTML file

PR #8764, opened 2025-06-10
dashboard: smoother profile avatar buttons (fixes #8756)

### LINKED ISSUE #8756: Cancel the selection
We can add cancel the selection at Edit Member Profile page

JunhaoZhou66:
fixes #8756
截屏2025-06-10 下午12 14 51


3011089503 (commented):
Do we need to delete the Image Crop Box after remove?

jessewashburn (changes requested):
Currently the crop section has the image persist even after removing it. Also the remove button should go away if there is no longer an image to remove. 
image

It would be cleaner if removing the image reverted us back to this kind of view
image

3011089503 (commented):
@jessewashburn good suggestions! I have created a new commit to improve that

jessewashburn (approved):

PR #8832, opened 2025-06-30
all: less images is more (fixes #8831)

### LINKED ISSUE #8831: Icons cleanup
Cleanup on icons. Experiment with a few newer icons as per @PlataformasInformaticas suggestions

Image

Mutugiii:
Fixes #8831

- Assets cleanup
- Update icons for myLibrary and feedback list

image


jessewashburn:
I think these changes are fine - the feedback list is definitely an improvement. I think for resources, the folder icon made more sense than something that looks like a document, as resources can be various different formats. Also, this new icon looks like `copy` icon.

Also notice this alignment issue in dashboard on mobile view (not an issue with this PR specifically). Maybe a separate issue. 

image

dogi (approved):

PR #8834, opened 2025-06-30
teams: smoother survey views (fixes #8823)

### LINKED ISSUE #8823: survey: mobile ui out of bound
Image

The next/last question button is out of bounds on mobile, making it hard for the user to click

JunhaoZhou66:
fixes #8823
截屏2025-06-30 下午12 50 41


jessewashburn (changes requested):
I think the UI could make better use of the space. Also, double check that there aren't unnecessary line changes. 
image

image

3011089503 (commented):
@jessewashburn  I have created commits to remove unnecessary line changes. For the UI , I think that the change question button in a line will be better

Mutugiii (approved):

PR #8888, opened 2025-07-11
manager: smoother dashboard (fixes #8882)

### LINKED ISSUE #8882: manager dashboard: data mess up
Image

JunhaoZhou66:
截屏2025-07-11 上午5 53 27


3011089503 (commented):
When under 480 it still the same Ui like before, I trying to fix this issue

jessewashburn (changes requested):
These new boxes look great, but they should replace the old data fields and be visible on all screen sizes. Currently only see them on screen sm and xs. 

image

3011089503:
> These new boxes look great, but they should replace the old data fields and be visible on all screen sizes. Currently only see them on screen sm and xs.
> 
> image

yes I think this is related with the ismobile function's problem, I also meet this problem in other pr

3011089503 (commented):
I have push two commits to solve this problem

Mutugiii (approved):

PR #8953, opened 2025-07-28
courses: smoother creation workflow (fixes #8952)

### LINKED ISSUE #8952: courses: improve draft workflow
Currently in the courses add/edit we currently automatically save the draft content.

We could improve the workflow to:
1. be more intuitive for the learners
2. accomodate the current unsaved changes dialog

Mutugiii:
Fixes #8952

Improve the courses add/edit draft workflow. This is more intuitive for users & accomodates the unsaved changes dialog. Previously, changes were being auto saved.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.19.76",
+  "version": "0.19.80",
   "myplanet": {
-    "latest": "v0.27.60",
-    "min": "v0.26.60"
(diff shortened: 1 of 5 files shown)


jessewashburn (changes requested):
It would be nice if there was a way to delete a draft. 

Also, the unsaved changes guard is firing when a course is added. 

image

Mutugiii:
Fixed the unsaved changes dialog issue on course add

Also added a delete draft workflow

image

dogi (approved):

PR #9069, opened 2025-09-01
manager: smoother currency configuring (fixes #9064)

### LINKED ISSUE #9064: Currency configuration.
Add Currency configuration for reports and interface

Image

jessewashburn:
fixes #9064

Added a currency configuration tab in manager settings.

image

image

Reflects in the finances tab of the community page

image

And in enterprises

image

Added proper spanish translations for all

image


jessewashburn:
The only hitch is that I think existing communities will need to go and configure the currency before they can add more finance reports.

src/app/manager-dashboard/manager-currency.component.ts
```diff
@@ -0,0 +1,53 @@
+import { Component, OnInit } from '@angular/core';
+import { UntypedFormBuilder, UntypedFormGroup, Validators } from '@angular/forms';
```
Mutugiii, line 2: We should use typed forms here.

jessewashburn:
@Mutugiii  I think I have incorporated all of  your suggestions.

dogi (approved):

PR #9070, opened 2025-09-01
exams: add rating scale question type (fixes #9068)

### LINKED ISSUE #9068: add matrix question type in exams component
In the exams component, add support for the new matrix question type.

Mutugiii:
Fixes #9068

- Add default 1-9 rating scale question type to surveys.
- Add custom rating scale type representation in PDF export.

image

image

image

### DIFF
--- a/src/app/exams/exams-view.scss
+++ b/src/app/exams/exams-view.scss
@@ -1,3 +1,5 @@
+@import '../variables';
+
 :host {
 
   .view-container {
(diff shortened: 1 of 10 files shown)


jessewashburn (approved):

jessewashburn:
Works nicely. My only note is that Matrix- number keypad` sounds pretty technical. Maybe `Rating Scale – 1 to 9` would be more intuitive.

dogi (changes requested):
good job @Mutugiii 

for the analysis we should flip the X and the Y

src/app/exams/exams-view.scss
```diff
@@ -48,4 +50,40 @@
...
+  .rating-scale-keypad {
+    display: flex;
+    justify-content: flex-start;
+    margin: 20px 0;
+  }
+
+  .rating-scale-grid {
+    display: grid;
+    grid-template-columns: repeat(3, 1fr);
+    gap: 6px;
```
paulbert, line 63: This gap feels a little tight to me.

Material says 8px (which is close enough to 6 to not bother)
https://m2.material.io/design/layout/spacing-methods.html#containers-and-ratios

This article which links to a study says 24-36px for that size button
https://uxmovement.com/mobile/optimal-size-and-spacing-for-mobile-buttons/

You mentioned this is targeted for phones so accuracy might be an issue if the buttons are too close together.

src/app/submissions/submissions.service.ts
```diff
@@ -671,6 +695,38 @@ export class SubmissionsService {
     };
   }
 
+  aggregateRatingScaleResponses(question, submissions) {
```
paulbert, line 698: Is there a reason we can't use `aggregateQuestionResponses` instead? There's a bit of duplication here that I think we can avoid.

This function could call that function and then do some data manipulation/sorting if needed.

Mutugiii: Just combined the `aggregateRatingScaleResponses` with the `aggregateQuestionResponses`

PR #9076, opened 2025-09-04
Fix Title from the username data

PlataformasInformaticas:
I deleted the username from the title of the final step of the survey.


jessewashburn (changes requested):
I'm not sure I understand this accomplishes what is stated in your description. Can you give a little more info about what the goal is here? This only gets rid of the username in the user update page toolbar, and does not affect surveys as far as I can tell. 

image

PR #9133, opened 2025-09-23
fix: disable 'Remove Description' button when no description exists (#9095)

### LINKED ISSUE #9095: community: disable "Remove Description" icon when no description exists
**What is the current behavior?**
On the Community -> Services page, the "Remove Description" icon is always clickable, even when there is no description text to remove. This creates a confusing user experience.

**What is the expected behavior?**
The "Remove Description" icon should be disabled (greyed out and non-clickable) when there is no description available for a service. It should only be enabled if there is text to remove.

Image

**Possible Solution**
A front-end fix to add a condition in the template to disable the button.

Saby-Bishops:
Fix #9095.

- Disabled the "Remove Description" button in community.component.html when team.description is empty or whitespace.
- Ensures button only enabled when a description exists.
image

Tested:
- Empty description → button disabled.
- Non-empty description → button enabled and works as expected.

### DIFF
--- a/src/app/community/community.component.html
+++ b/src/app/community/community.component.html
@@ -70,7 +70,7 @@
               <button class="toggle-button" (click)="openDescriptionDialog()" mat-stroked-button i18n>
                 { servicesDescriptionLabel, select, Edit {Edit} Add {Add}} { configuration.planetType, select, community {Community} nation {Nation} center {Earth}} Description
               </button>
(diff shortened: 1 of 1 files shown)


emmanuelbaa:
Hi, it would be really helpful if you could include a screenshot in the PR to showcase the feature you worked on.

PR #9134, opened 2025-09-23
dashboard: smoother events creation (fixes #9109)

### LINKED ISSUE #9109: community events time field
When creating an event, the time field is entered manually. We should support a timepicker similar to the tasks in teams.

Image

Teams Tasks:

Image

emmanuelbaa:
fixes #9109 

added a time picker to the add event dialogue.

timePicker


jessewashburn (changes requested):
I like the idea, but currently I'm seeing some weird margins. 

On Firefox:
image

On Chrome the issue is more subtle, but you can see the top of the 0 and the 2 are slightly cut off. 
image

Saby-Bishops:
Hi @emmanuelbaa,
I think double-checking the CSS for padding/line-height on the input field might be a good idea? It might just need a small adjustment to ensure consistent rendering across browsers.
Otherwise, the functionality looks solid.

Mutugiii:
stale

PR #9155, opened 2025-09-26
Prune unused global styles – impact 60, feasibility 50

dogi:
Scope or remove unused SCSS so the generated styles.css is dramatically smaller and blocks less rendering.

## Summary
- trim `styles.scss` to drop unused utility selectors and only emit the margin helpers referenced in templates
- move gradient, rating, chat, and other niche styles into component-level SCSS with matching `styleUrls`
- add feature styles for meetups/resources/courses and relocate shared helpers out of the global bundle
- introduce a production-only PurgeCSS step with a safelist to protect dynamic Angular Material classes

## Testing
- `CI=1 npx ng build --configuration production --stats-json`

------
https://chatgpt.com/codex/tasks/task_e_68d5d2d17810832b990de35f6ec2efed

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -57,6 +57,7 @@
               "node_modules/@fortawesome/fontawesome-free/css/all.css"
             ],
             "scripts": [],
+            "postcssConfig": "postcss.config.js",
             "i18nMissingTranslation": "ignore",
             "poll": 2000,
             "vendorChunk": true,
--- a/src/app/courses/courses.scss
+++ b/src/app/courses/courses.scss
@@ -1,67 +1,11 @@
 @import '../variables';
+@import '../shared/styles/collection-layout';
 
 $toolbar-height: 25vh;
 $label-height: 1rem;
 $list-title-height: 36px;
 
-.list-content-menu,
-.table-selection-top {
-  align-self: start;
-}
-
-.table-selection-top {
-  height: $list-title-height;
-  min-height: $list-title-height;
-}
-
-.list-content-menu {
-  display: grid;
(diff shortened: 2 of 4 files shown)


src/app/news/news-list-item.scss
```diff
@@ -71,6 +71,72 @@ mat-chip {
...
+  border: none;
+  padding: 5px;
+  cursor: pointer;
+  color: #f8f8f2;
+  transition: color 0.3s;
+}
+
+:host ::ng-deep .copy-btn.copied {
+  color: $primary;
+}
```
Mutugiii, line 138: This is a duplication of what is in the chat-window.scss

src/styles.scss
```diff
@@ -132,191 +103,6 @@ body {
...
-    .tags-list {
-      grid-area: tags;
-    }
-
-    .course-progress {
-      grid-area: progress;
-      margin-bottom: 0.5rem;
-    }
-
-  }
```
Mutugiii, line 184: Removing this from the `styles.scss` results in a duplication in the `courses.scss`, `resources.scss` & `meetups.component.scss`

Mutugiii:
The PurgeCSS change is okay as long as we maintain a suitable safelist, as codex review has pointed out with mdc-*.

But I think lazy-loading those shared styles into components isn’t worth it, especially since the chat/news styles will already always be loaded (because community landing uses news), and the courses/resources (and planet rating in the courses & resources) styles are core features. This adds maintenance overhead for marginal gain in my opinion.

PR #9156, opened 2025-09-26
Bundle third-party editors locally – impact 70, feasibility 65

dogi:
## Summary
- add easymde and codemirror to the Angular bundle and remove the unpkg script/style tags
- import EasyMDE styles through the global stylesheet and lazily hydrate the editor component with a dynamic import
- disable Angular CLI analytics to avoid prompting during automated lint runs

## Testing
- npm run lint
- npm run test -- --watch=false --browsers=ChromeHeadless --progress=false *(fails: existing spec files reference removed modules and deprecated rxjs imports)*
- npm run build -- --progress=false

------
https://chatgpt.com/codex/tasks/task_e_68d5d2c751a8832b83b418754cee0c5c


emmanuelbaa (approved):

jessewashburn (approved):

src/app/shared/forms/planet-markdown-textbox.component.ts
```diff
@@ -11,6 +24,24 @@ import { DialogsImagesComponent } from '../dialogs/dialogs-images.component';
...
+function ensureEasyMDELoaded(): Promise<void> {
+  if (!easymdeLoader) {
+    easymdeLoader = import('easymde')
+      .then((module) => {
+        const easyMdeConstructor = module.default ?? module;
+        (globalThis as any).EasyMDE = easyMdeConstructor;
+      })
+      .catch((error) => {
+        easymdeLoader = null;
+        throw error;
```
paulbert, line 38: I think just throwing an error here will result in the whole app crashing if easymde doesn't initialize.

Good to catch the error here, but to then just throw an error isn't great. Would be better to display a message that the editor didn't load correctly.

package.json
```diff
@@ -46,7 +46,9 @@
     "@fullcalendar/daygrid": "^6.1.10",
     "@fullcalendar/interaction": "^6.1.10",
     "chart.js": "^3.9.1",
+    "codemirror": "^5.65.16",
```
paulbert, line 49: Not sure what this is. I think it'd be better to let whatever dependency relies on this to handle its version instead of declaring it here.

paulbert (commented):
On the whole I'm not sure what the benefit of this change would be?

The covalent editor will already gracefully switch to a text input if easymde is not there. I suppose there is a benefit to pulling easymde out of the index.html where it's a bit hidden, but that will add another package we need to manually update the version for whenever we upgrade Angular.

PR #9159, opened 2025-09-26
Lazy-load PDF/markdown toolchain – impact 75, feasibility 60

dogi:
Dynamically import pdf/markdown libraries so they load only for export features.

------
https://chatgpt.com/codex/tasks/task_e_68d5d2750f98832b8604b2f48dac9eb1


package.json
```diff
@@ -82,6 +88,7 @@
     "@types/pouchdb-find": "^6.3.4",
     "@typescript-eslint/eslint-plugin": "^5.43.0",
     "@typescript-eslint/parser": "^5.43.0",
+    "bytebuffer": "^5.0.1",
     "eslint": "^8.28.0",
     "eslint-plugin-import": "^2.29.1",
     "eslint-plugin-jsdoc": "^46.10.1",
```
jessewashburn, line 94: I don't see any evidence that these new dependencies are needed, unless there are plans to incorporate them in the future.

jessewashburn (changes requested):

PR #9174, opened 2025-09-30
Refactor Angular Material theming configuration

dogi:
## Summary
- wrap the Angular Material palettes in a theme configuration map that includes color, typography, and density details
- consume the shared theme map when generating section-specific themes and avoid re-emitting typography and density styles

## Testing
- `CI=1 NG_CLI_ANALYTICS=false npm run build -- --progress=false`

------
https://chatgpt.com/codex/tasks/task_e_68dbdbe350cc832b9116cf6ea6655783


emmanuelbaa (approved):

jessewashburn (changes requested):
Seems to me that since we're using the default Material typography, wrapping it in a map adds no value and just makes the code more verbose.

Mutugiii:
#9723

PR #9233, opened 2025-10-15
teams: smoother survey navigating (fixes #9130)

### LINKED ISSUE #9130: team/enterprises: hide the table headers if there are no surveys
Image

emmanuelbaa:
fixes #9130

feat(surveys): Hide table headers when no surveys are present

header hide

This commit fixes an issue where the table headers in the surveys view were visible even when there were no surveys to display.

The fix involves conditionally adding a class to the `mat-table` element, which is then used to hide the header row with CSS. This ensures that the headers are only displayed when there is data in the table.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.20.57",
+  "version": "0.20.58",
   "myplanet": {
-    "latest": "v0.35.42",
-    "min": "v0.34.42"
+    "latest": "v0.35.64",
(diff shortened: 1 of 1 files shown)


Saby-Bishops (approved):

jessewashburn (changes requested):
Good fix, but I think you can do it without so many line changes.
Recommend either adding *ngIf="surveys.data.length > 0" directly to the mat-header-row, or use the existing *matNoDataRow pattern that other components in the codebase already use (e.g. courses.component.html).

PR #9252, opened 2025-10-22
log remote details in pre-push hook

Saby-Bishops:
## Summary
- reintroduce the remote and url assignments at the top of the pre-push hook
- print the remote name and URL before running the lint steps so the values are used

## Testing
- PATH="/tmp/fake-bin:$PATH" git-hooks/pre-push origin https://example.com/repo.git

------
https://chatgpt.com/codex/tasks/task_e_68f7d45a0008832d8fd6b1962b359134

### DIFF
--- a/git-hooks/pre-push
+++ b/git-hooks/pre-push
@@ -22,6 +22,10 @@
 remote="$1"
 url="$2"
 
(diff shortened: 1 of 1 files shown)


Mutugiii:
Remote var removed in #9253

Issue #9166, opened 2025-09-29
Migrate to new Material components as of Angular 15


paulbert:
Many of the Angular Material components were refactored as of Angular 15. When going through the automatic upgrade process for Angular from 14 -> 15 these components were replaced with "Legacy" versions. We need to migrate to the newer component versions.

See https://v15.material.angular.dev/guide/mdc-migration

PR #9263, opened 2025-10-23
all: smoother input forms material autocompleting (connects #9166)

paulbert:
Update script grouped together FormField, Autocomplete, Input, and Select (probably because they are very similar and may share some styles). This changes inputs and form fields quite a bit.

A couple of initial observations:
- The markdown textbox shouldn't have that background. I think I added CSS to remove the form fields styles from this before, so hopefully I can just edit that and it'll be OK again.
- Form fields generally take up more space, so we may need to adjust some of the layouts (top bar on voices page stands out for this).
- Placeholder text only shows up when the field is empty right now. I'll check to see if adding a label isn't too hard.
- I think there's extra padding now making less of the text visible when there is overflow on a field (i.e. the title of a resource)
- There are two options for how a field is styled, the default "fill" or a simpler "outline" (shown below in the resource title). Not sure which I like better. Fill is a little more minimal, but the color change stands out vs outline which maintains the white background.

image


paulbert:
@Mutugiii Made those updates and replaced most of the placeholders with labels.

The only places I didn't were when the placeholder makes sense to keep in addition to a label.

uj-sxn (approved):

Mutugiii (approved):

dogi (approved):

PR #9387, opened 2025-11-26
manager: smoother survey submissions viewing (fixes #9353)

### LINKED ISSUE #9353: surveys page loading
The submissions fetch request on planet learning takes a long time to load; this needs to be investigated

Mutugiii:
Fixes #9353

- Add submissions db view and index
- Use maps to improve lookups performance
- Handle missing subscriptions cleanup


Saby-Bishops (approved):

emmanuelbaa (approved):

paulbert (commented):
I think this is adding more complication and I'm not sure it'll have the performance improvement we're going for.

Is it possible to do this with one view instead of two? That's adding a database request so I'm not sure there's a performance gain there.

I see you're adding in a lot of Maps. It's adding more looping over the arrays (though removing some of the previous loops within the map is good), and the main source of the additional complication. This might give a little performance boost, but I think it'd be better to stick to the arrays at this stage and reduce the amount of times we loop over them (maybe combining different parts of the processing into `reduce` functions?).

jessewashburn (approved):

paulbert (approved):
This is simpler, thanks! I'm going to approve so we can see if we get a performance improvement from this. If we don't or it's not enough, we could try a MapReduce for the view instead of just a Map to move some of the aggregation from the client to the server.

dogi (approved):

PR #9424, opened 2025-12-10
resources: smoother icon downloading  (fixes #9307)

### LINKED ISSUE #9307: File downloadable shouldn’t be selectable if no file has been uploaded
When creating/editing a resource

Image

JackTheSparow:
Fixes #9307 

Fixed the checkbox enabled problem when file is deleted after selecting from local files. Checkbox automatically gets refreshed if file is deleted and not uploaded.

image

### After Deleting File

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,9 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.21.15",
+  "version": "0.21.20",
   "myplanet": {
-    "latest": "v0.41.2",
+    "latest": "v0.42.56",
     "min": "v0.37.60"
   },
   "scripts": {
--- a/src/app/resources/resources-add.component.html
+++ b/src/app/resources/resources-add.component.html
@@ -109,7 +109,7 @@
           </div>
         </div>
         <div class="inner-gaps full-width">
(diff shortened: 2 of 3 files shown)


Saby-Bishops (commented):
Please add a screenshot of the output, and include the relevant issue number in the PR description for easier tracking.

Saby-Bishops (approved):

jessewashburn (dismissed):
It's nice that it refreshes after you remove an attachment, but you should not be able to do this, per the original issue (checked even when there is no attachment):

image

Also, If you include fixes #9307 in the body of the PR, it is much easier for the reviewers to directly link to the original issue.

emmanuelbaa (approved):

Mutugiii (changes requested):
Image

Steps to replicate:
1. Upload item
2. Select `File downloadable`
3. Delete the item

Mutugiii (changes requested):
Additional note, the `showDownloadCheckbox` and `disableDownload` variables are basically inverses most of the time. Sticking to one is ideal to reduce maintenance overhead

Mutugiii (approved):

dogi (approved):

PR #9426, opened 2025-12-10
Re-enable beta authentication guard

Saby-Bishops:
## Summary
- add BetaThenAuthService to root routes so activation and child routes honor beta gating
- register the guard in the app module and document how beta toggles affect auth fallback
- ensure the guard supports child activation paths

## Testing
- not run (not requested)

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_6939ce7a9700832d92391f719983029c)


Mutugiii:
The `betaThenAuthService` is in place for archival reasons, but we currently don’t have any routes that require beta access.

PR #9445, opened 2025-12-16
chat: enter key full functionality (fixes #9428)

### LINKED ISSUE #9428: chat: enter key does not send chat message
Currently, pressing the Enter key creates a new line instead of sending the chat message. The message is only sent when you click the send icon.

It would be more intuitive if pressing **Enter** sent the message and **Shift + Enter** created a new line, similar to the behavior in other chat applications like ChatGPT.

Image

emmanuelbaa:
fixes #9428 

make conf

This commit introduces several key improvements to the chat feature to enhance user experience and fix bugs.

1.  **Enter-to-Send Functionality:**
    - Implements the standard chat behavior where the "Enter" key sends the message and "Shift + Enter" creates a new line.
    - Adds a `keydown` event listener to the chat textarea to handle this logic.
    - Replicates the existing form validation from the send button to prevent empty messages or multiple submissions.

2.  **Immediate UI Feedback (Optimistic Update):**
    - The chat UI is now updated optimistically. The user's message is added to the conversation list and the input field is cleared as soon as the message is submitted, rather than waiting for the server response. This makes the chat feel more responsive.

3.  **Sidebar Conversation History Fix:**
    - Fixes a bug where new conversations were not appearing in the sidebar history until after the first response was received.
    - A signal is now sent to the `ChatService` as soon as a message is sent in a new conversation, ensuring the sidebar updates immediately.
    - Fixes a subsequent bug where this signal was being sent twice.

4.  **UI Rendering Fix:**
    - Restores the `role: 'user'` property to the conversation object when performing an optimistic update, preventing potential UI rendering issues for user messages.


Mutugiii (changes requested):
This brings about some regressions in the chat:
- On submit, we see the last created chat instead of the question asked
- Newly created chats are no longer showing up in the sidebar
- Enter submission disregards the chat disabled state

emmanuelbaa:
> This brings about some regressions in the chat:
> 
> * On submit, we see the last created chat instead of the question asked
> * Newly created chats are no longer showing up in the sidebar
> * Enter submission disregards the chat disabled state

great reviews, i will work on that

Saby-Bishops (changes requested):
There still seems to be some regressions where Enter triggers submission even when the action is disabled, and where new chats don’t always update the sidebar as expected.

Mutugiii:
stale

PR #9510, opened 2026-01-22
all: smoother cards material handling (connects #9166)

Mutugiii:
Initial observations
- Card is flex by default, making some of our styling redundant
- Padding is not applied to the root, but to content areas, similarly, making some of our custom styling redundant & requiring some new padding adjustments


uj-sxn (approved):

paulbert (commented):
It's mostly looking good, but a few of the components need some adjustments to their spacing.

The spacing is too tight here. Let's extend the card to fill the space better and leave space between the labels and values in the content
Image

Feedback no longer has any padding
Image

paulbert (approved):
Thanks for the changes looks good now!

Saby-Bishops (approved):

dogi (approved):

PR #9587, opened 2026-02-11
Remove unused deferred promise from Protractor onPrepare

Saby-Bishops:
### Motivation
- Remove unused `protractor.promise.defer()` and `defer.fulfill()` from `onPrepare` since the deferred value was never awaited and `onPrepare` can return the `user.create()` promise directly.

### Description
- Deleted the `var defer = protractor.promise.defer();` line and removed the `defer.fulfill();` call, and updated the `.then` handler to `return res;` so `onPrepare` returns the `user.create()` promise chain directly.

### Testing
- Ran `node --check protractor.conf.js` which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_698b9e19d7d8832d8b8ed61abcf06c42)

### DIFF
--- a/protractor.conf.js
+++ b/protractor.conf.js
@@ -26,7 +26,6 @@ exports.config = {
     print: function() {}
   },
   onPrepare() {
-    var defer = protractor.promise.defer();
     require('ts-node').register({
       project: './e2e/tsconfig.e2e.json'
     });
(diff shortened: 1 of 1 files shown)


protractor.conf.js
Mutugiii: Not prioritizing test updates right now, as current tests are unmaintained.

PR #9604, opened 2026-02-16
Tighten chat service return types and simplify call sites

Saby-Bishops:
### Motivation
- Ensure `chat` and `chatNoSave` have strict success types (no `undefined`) because errors are thrown via a single error path. 
- Reflect that the error handler always throws by marking it `never` so TypeScript can infer tighter return behavior. 
- Remove redundant optional chaining at call sites now that success returns are non-optional.

### Description
- Change `handleChatError` signature to `: never` to indicate it always throws. 
- Update `chat` in `chatapi/src/services/chat.service.ts` to return `Promise<{ completionText: string; couchSaveResponse: DocumentInsertResponse }>` (removed `| undefined`). 
- Update `chatNoSave` in `chatapi/src/services/chat.service.ts` to return `Promise<string>` (removed `| undefined`). 
- Update `chatapi/src/index.ts` to remove unnecessary optional checks and chaining: removed the `if (chatResponse)` guard in the websocket final path and removed `response?.completionText` / `response?.couchSaveResponse` in the POST save response.

### Testing
- Ran `npm run build` in `chatapi` which compiles TypeScript and succeeded. 
- Ran `npm run lint` in `chatapi` which passed with no lint errors.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_698b9e1bc8a0832d8d8ba6e2e61b09fa)


Mutugiii:
nit

PR #9610, opened 2026-02-17
Remove deprecated assistant-mode payload handling

Saby-Bishops:
### Motivation
- Remove legacy "assistant mode" code paths and the `assistant` request flag to simplify chat handling and consolidate on standard chat completions flows. 
- Protect the API from accidental assistant-mode usage by making the request contract explicit: any payload including `assistant` should be rejected. 

### Description
- Deleted assistant-only branches and removed imports relying on the assistants API from `chatapi/src/utils/chat-helpers.utils.ts`, so streaming/non-streaming now always use regular chat completions. 
- Simplified the internal API surface by removing the `assistant` parameter from `aiChat` (`chatapi/src/utils/chat.utils.ts`) and from callers in `chatapi/src/services/chat.service.ts`. 
- Updated HTTP and WebSocket entrypoints in `chatapi/src/index.ts` to explicitly reject requests containing `data.assistant` with `400 Bad Request`. 
- Updated frontend types and callers to stop emitting the flag by removing `assistant` from `src/app/chat/chat.model.ts`, `src/app/chat/chat-window/chat-window.component.ts`, and from the AI analysis call in `src/app/submissions/submissions.service.ts`. 
- Updated `chatapi/README.md` to remove `assistant` from examples and mark it as deprecated/unsupported in the API contract. 

### Testing
- Attempted a TypeScript build for the chat API with `npm --prefix chatapi run build`, which could not complete in this environment because project dependencies/type declarations are not installed locally and the build failed with missing modules/type errors (examples: `openai`, `nano`, `express`, and Node type declarations).

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_699381e61bf0832d8370d0b38b8e8cb8)

### DIFF
--- a/chatapi/README.md
+++ b/chatapi/README.md
@@ -66,7 +66,6 @@ If you are using an 32bit arm device and encounter a fatal error while running t
     "data": {
       "user": "admin",
       "content": "Hello",
-      "assistant": false,
       "context": "",
       "aiProvider": {
         "name": "openai",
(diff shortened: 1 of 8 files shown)


Mutugiii:
since assistants API is in use in planet, we should switch from that on planet before deprecating its usage in the chatapi

PR #9614, opened 2026-02-18
incorrect age calculation in ageFromBirthDate(fixes #9616)

### LINKED ISSUE #9616: incorrect age calculation in ageFromBirthDate function
- Problem:
The function ageFromBirthDate in src/app/shared/utils.ts used getDay() (day of week) instead of getDate() (day of month) when comparing the current day and the birth day. This produced incorrect ages when the birthday in the current year had not yet occurred.

- Fix
Use getDate() instead of getDay() for day-of-month comparison.

uj-sxn:
The `ageFromBirthDate` function in `src/app/shared/utils.ts` was using `getDay()` (day of week) instead of `getDate()` (day of month) to compare the current day with the birth day. This caused incorrect age calculations when the birthday hadn't passed in the current year but the day of week index was greater than or equal to the birth day of week index.

This PR fixes the issue #9616  by using `getDate()` and adds a regression test in `src/app/shared/utils.spec.ts`.

---
*PR created automatically by Jules for task


Saby-Bishops (changes requested):
You can remove the spec file as they are not being maintained currently.

PR #9617, opened 2026-02-18
community: clean up css class (fixes #9478)

### LINKED ISSUE #9478: community: clean up css class
There's a class, `sticky-button` on the following div that shouldn't really be there. The `sticky-button` class is for making the voices toolbar stick to the top of the page while scrolling through the voices posts.

https://github.com/open-learning-exchange/planet/blob/3ea74e8911450f4d1463de7a3d3d81fe898a9547/src/app/community/community.component.html#L69

There should be a generic class that can be added to that div to add the following properties to it to maintain alignment fixed after #9420 

```
display: flex;
gap: 0.5rem;
```

After that remove those properties from the `sticky-button` class.

uj-sxn:
Refactor the sticky-button class in community.component by decoupling layout properties (flex and gap) into a new generic class .flex-row-gap. This fixes an alignment issue where the sticky-button class was being misused for layout on a specific div. Also addresses sass-lint warnings for the gap property.

Addresses issue #9478

### DIFF
--- a/src/app/community/community.component.html
+++ b/src/app/community/community.component.html
@@ -68,7 +68,7 @@
             <ng-template #noTeamDesc><p i18n>No description available.</p></ng-template>
           </ng-container>
           <ng-container *planetAuthorizedRoles="''">
-            <div *ngIf="!planetCode" class="action-buttons sticky-button">
+            <div *ngIf="!planetCode" class="action-buttons flex-row-gap">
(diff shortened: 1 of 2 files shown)


Saby-Bishops (changes requested):
The sass-lint disables around gap seem inconsistent, might be worth standardizing or handling via lint config.

PR #9618, opened 2026-02-18
Document legacy helper scripts and add CI/deploy invocation points

Saby-Bishops:
### Motivation

- Preserve and document legacy operational scripts that are still referenced by deployment and e2e flows so operators understand their purpose and prerequisites.
- Ensure deployments explicitly toggle the `/srv/starthub` marker on the remote host to match existing `starthub-*` behavior and fail early if the target cannot support it.
- Provide an explicit CI invocation for pinning the legacy ChromeDriver used by Protractor-based e2e so manual runs are repeatable and scoped.

### Description

- Added a new `Script Maintenance & Operational Helpers` section to `README.md` that documents the purpose and prerequisites for `starthub-true`/`starthub-false` and `webdriver-set-version`, including the writable `/srv/starthub` requirement and ChromeDriver `2.37` compatibility expectations for legacy Protractor e2e.
- Updated `.github/workflows/deploy.yml` to verify `/srv/starthub` writability on the remote host, to write `true` before restarting services, and to write `false` after restart, and to exit with an error if the path is not writable.
- Added a `legacy-e2e-webdriver-setup` manual-dispatch job to `.github/workflows/planet.yml` that checks out code, sets up Node, runs `npm ci`, and executes `npm run webdriver-set-version` to pin the legacy ChromeDriver for Protractor e2e runs.

### Testing

- Validated the modified workflow YAML with `ruby -e "require 'yaml'; YAML.load_file('.github/workflows/deploy.yml'); YAML.load_file('.github/workflows/planet.yml')"`, which completed successfully.
- Attempted to validate YAML with Python using `yaml.safe_load` but the `PyYAML` module was not available in the environment, so that check was not executed.
- Performed local file diffs and basic syntax inspection of `README.md`, `.github/workflows/deploy.yml`, and `.github/workflows/planet.yml` to confirm the intended insertions were applied without syntax errors.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_699381e9ddd0832d9199f91776dd5418)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -55,10 +55,16 @@ jobs:
           echo "planet.dev.ole.org ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBFqXBJPFe+pH3L57o1ildAxHssG4lpkloTcw3Wbs64c7bL8M6hR0rre4ufpCKboVLn4trJqbKOPWtFgBJHsgqXA=" > ~/.ssh/known_hosts
           # echo "planet.earth.ole.org ...."
           ssh root@$SERVER_HOST <<EOF
+          if [ ! -w /srv ] && [ ! -w /srv/starthub ] && [ ! -e /srv/starthub ]; then
+            echo "Error: /srv/starthub is not writable on deployment target"
+            exit 1
+          fi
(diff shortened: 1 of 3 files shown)


Mutugiii:
This PR doesn't address a current need

- hubs state toggling is not part of our active deployment requirements today
- added manual legacy webdriver setup doesn't improve workflow paths
- the legacy e2e webdriver support will be removed once tests start being maintained

PR #9620, opened 2026-02-19
all: less material legacy core is more (closes #9166)

### LINKED ISSUE #9166: Migrate to new Material components as of Angular 15

Mutugiii:
Closes #9166

Post MDC migration cleanup and prep for the next Angular version

- Positional args deprecated in the Theme definition API
- Migrated some pending TODOs to mdc styles and removed some that are no longer needed
- Updated the form fields' highlight and hover behaviour to mimic pre-mdc behaviour
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,9 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.21.59",
+  "version": "0.21.62",
   "myplanet": {
-    "latest": "v0.46.34",
+    "latest": "v0.47.0",
     "min": "v0.37.60"
   },
   "scripts": {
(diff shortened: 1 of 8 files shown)


Saby-Bishops (approved):

uj-sxn (approved):
Tested locally: Reports detail filters + toggles + export button render correctly; verified dropdowns open/hover/focus states look consistent.

paulbert (approved):
Just the one space to revert, but otherwise this looks good to me!

dogi (approved):

PR #9634, opened 2026-02-23
Make dialogs announcement examId internal-only

Saby-Bishops:
### Motivation
- Reduce exported surface by making `examId` internal to `dialogs-announcement.component.ts` because its usage is internal only and no cross-file references are required.

### Description
- Replaced `export const examId = ...` with `const examId = ...` in `src/app/shared/dialogs/dialogs-announcement.component.ts` so the constant is module-private and runtime behavior is unchanged.

### Testing
- Ran an import search and diff checks with `rg` and `git diff` to confirm there are no cross-file imports of `examId` and that the only change is the export removal, and both checks succeeded.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_699cc062aeb8832da25b92b1eb8b80d5)


Mutugiii:
examId is not used outside the dialogs announcement component

PR #9643, opened 2026-02-24
Resolve nation APK version metadata before writing apkversion

Saby-Bishops:
### Motivation
- The script previously attempted to write an undefined `$tag` to `/usr/share/nginx/html/fs/apkversion`, so the version needs to be resolved from nation-hosted metadata instead of relying on an undefined variable.
- Prefer authoritative nation metadata (a nation-hosted `/fs/apkversion`) and fall back to HTTP response headers so the server-side published version is captured when available.
- Avoid clobbering an existing `/usr/share/nginx/html/fs/apkversion` when metadata fetch or parsing fails.

### Description
- Add `resolve_apk_version()` which first fetches `"$nationUrl/fs/apkversion"` and then falls back to parsing an `x-apk-version` header from a `HEAD`/`GET` response to `"$downloadUrl"`. 
- Add `write_apk_version()` which calls `resolve_apk_version()` and only writes the resolved non-empty value to `/usr/share/nginx/html/fs/apkversion`, leaving the existing file unchanged on failure. 
- Replace the previous `echo $tag > /usr/share/nginx/html/fs/apkversion` occurrences with `write_apk_version` in both successful download/verify paths. 
- Preserve existing download and SHA verification logic and only change how the APK version is determined and recorded.

### Testing
- Ran a shell syntax check with `sh -n docker/planet/nginx/upgrade_planetapk_fromnation.sh`, which completed without errors.
- Verified the script changes in a diff and committed the updated file to the repository successfully.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_699cc075d64c832d9621dde568fea448)

### DIFF
--- a/docker/planet/nginx/upgrade_planetapk_fromnation.sh
+++ b/docker/planet/nginx/upgrade_planetapk_fromnation.sh
@@ -11,6 +11,39 @@ sha256File="$apkName.sha256"
 nationUrl="$(curl -s http://couchdb:5984/configurations/$(curl http://couchdb:5984/configurations/_all_docs -s | jq .rows[0].id -r) | jq .parentDomain -r | sed -e 's|^[^/]*//||' -e 's|/.*$||')"
 downloadUrl="$nationUrl/fs/$apkName"
 
+resolve_apk_version() {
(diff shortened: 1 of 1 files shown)


Mutugiii:
stale

PR #9654, opened 2026-02-26
Add assertion to verify test build maps environment.ts → environment.test.ts

Saby-Bishops:
### Motivation
- Make the role of the test environment file explicit and prevent false dead-code assumptions by failing CI when the mapping drifts. 
- Detect future tooling/`angular.json` changes by referencing the exact block `projects["planet-app"].architect.build.configurations.test.fileReplacements` so drift is obvious.

### Description
- Add `scripts/assert-test-environment-mapping.js`, a small Node script that reads `angular.json` and asserts the exact file replacement block maps `src/environments/environment.ts` to `src/environments/environment.test.ts` (throws on mismatch). 
- Wire the assertion into `package.json` by adding `assert:test-environment-mapping` and prepending it to the `test` and `lint-all` scripts so CI and local test runs enforce the check. 

### Testing
- Ran `node scripts/assert-test-environment-mapping.js`, which printed a verification message and exited successfully. 
- Ran `npm run assert:test-environment-mapping`, which also succeeded and printed the same verification message.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_699cc0578fb4832dab37127a92cb1562)

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,17 +12,18 @@
     "start": "ng serve",
     "dev": "bash dev-env.sh && ng serve --configuration dev",
     "build": "npm run ng-high-memory -- build --configuration production",
-    "test": "ng test",
+    "test": "npm run assert:test-environment-mapping && ng test",
     "htmlhint": "./node_modules/htmlhint-ng2/bin/htmlhint --config .htmlhintrc",
     "sass-lint": "./node_modules/sass-lint/bin/sass-lint.js -c ./sass-lint.yml -v -q",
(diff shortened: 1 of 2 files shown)


uj-sxn (commented):
The change itself looks good and the assertion is a nice preventative guardrail. I’m holding off on approval for now since the Planet Builder checks are failing (and the PR is still in draft)

Mutugiii:
No active test suite in CI right now, plus any drift in this mapping would fail first in the test run

PR #9690, opened 2026-03-06
Harden environment config parity and CI build validation

Saby-Bishops:
### Motivation
- Prevent environment drift by enforcing key parity across environment files and by validating both test and production builds in CI.
- Gate Docker pre-builds on successful linting and environment/build validation to catch issues early.

### Description
- Add npm scripts `build:test`, `build:prod`, and `check:environment-parity` to `package.json` to make test/prod builds explicit and provide a parity check command.
- Add `scripts/check-environment-parity.js` which compares top-level object keys in `src/environments/environment.ts`, `src/environments/environment.prod.ts`, and `src/environments/environment.test.ts` and exits non-zero on mismatches.
- Restore missing `chatAddress` in `src/environments/environment.test.ts` so the environment key sets align.
- Add a `validate-environment-builds` job to `.github/workflows/planet.yml` that runs `npm install`, `npm run lint`, `npm run check:environment-parity`, `npm run build:test`, and `npm run build:prod`, and make the existing `pre-build` job depend on it.
- Replace `npm ci` with `npm install` in the new CI job to avoid failing when no lockfile is present.

### Testing
- Ran `node scripts/check-environment-parity.js` locally and it passed (`Environment key parity check passed.`).
- Verified the new npm scripts exist by printing `package.json` script entries and confirming `build:test`, `build:prod`, and `check:environment-parity` are present.
- Attempted `npm ci` which failed locally due to a missing `package-lock.json`, so the CI job uses `npm install` instead to avoid that failure.
- Committed changes and validated the updated workflow file references and job dependency were applied (local diff/commit checks).

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69a9e72682fc832d8377a985a84bda21)


Mutugiii:
current environment builds are stable

PR #9726, opened 2026-03-13
courses: smoother row sizing (fixes #9725)

### LINKED ISSUE #9725: courses: improve course row sizing
Currently in the courses & resources, I see an avenue for some visual improvements

1. The sizing of the row content could be made standard. Currently, we have a max-height set resulting in rows differing in sizing based on the row contents
2. Overflows within the rows. In the description, we only account for images & text when setting preview size limits, resulting in some inline scrollbars. This is especially common when using other markdown styles like lists, tables & videos as we don't track the row height, but rather the character length of texts

Mutugiii:
Fixes #9725

- Constant sizing across the courses & resources list rows
- Avoid inline scrollbars for non-text, non-image markdown components, ala Tables, Lists, Videos...

image


Saby-Bishops (approved):

uj-sxn (approved):
Verified Resources/Courses preview rows: consistent height, no inline scrollbars, and expand toggle appears for long/overflowing descriptions. Looks good.

dogi (approved):

PR #9740, opened 2026-03-17
all: smoother planet building (fixes #9739)

### LINKED ISSUE #9739: Upgrade build system
Angular 17 deprecated the webpack build system in favor of one based on esbuild.

18 also introduced a script for upgrading ([documented here](https://v18.angular.dev/tools/cli/build-system-migration)) which we'll use to switch over to the newer system.

paulbert:
Upgrades the build system from the deprecated webpack system to esbuild.

So far needed just one adjustment to `pdfMake` because the build process didn't like mutating an imported object.

There are a few warnings I'll look at, too, but putting up the PR to see how the production build does.


Mutugiii (approved):
Looks & runs good.

only cleanup I see is the console.log in the `user-achievements component ts`

Mutugiii (approved):

Issue #8752, opened 2025-06-09
Angular upgrades


paulbert:
Main issue to keep track of getting back on track with Angular upgrades.

- [x] 11
- [x] 12
- [x] 13
- [x] 14
- [x] 15
- [ ] 20
- [ ] 21

PR #9754, opened 2026-03-19
all: bump `angular` to 19 (fixes #9749)

### LINKED ISSUE #9749: ng 19

Mutugiii:
Fixes #9749; connects #8752

Upgrade to Angular 19

#### Initial observations:
- Angular runs with standalone components as the default. Since our app is module-based, the script added explicit `standalone: false` declarations across the components.
- Some linting rules have been deprecated -> Some stylistic/formatting rules, and they should now be handled via stylistic angular or via prettier: https://eslint.style/guide/migration
- Minor code updates mostly around linting with the output binding & truthy values

##### Testing Notes:
- Remove `package-lock.json`
- Run `npm install`


Mutugiii:
Noticed an `npm WARN EBADENGINE` when running `npm install`

image

Saby-Bishops (approved):

src/app/users/users.service.ts
```diff
@@ -162,7 +162,7 @@ export class UsersService {
     const tempUser = {
       ...user,
       roles: roles,
-      oldRoles: [ ...user.roles ] || [ 'learner' ],
+      oldRoles: [ ...(user.roles ?? ['learner']) ],
```
paulbert: I think this still won't end up with the fallback `['learner']` value because CouchDB requires an array value for `roles` in the user database.

The intent is more that it would be (keeping the user.roles check in case CouchDB fails on that for some reason)
```
user.roles && user.roles.length > 0 ? [ ...user.roles ] : ['learner']
```
so if the roles array is empty we fallback on ['learner'].

paulbert (approved):
A couple small comments otherwise this is good!

dogi (approved):

PR #9757, opened 2026-03-19
Fix Protractor onPrepare readiness signaling

Saby-Bishops:
### Motivation
- Ensure Protractor only starts specs after the e2e test user is created and that startup deterministically fails if user creation fails.

### Description
- Updated `onPrepare` in `protractor.conf.js` to return `defer.promise` and wired `defer.fulfill()` / `defer.reject()` to the outcome of `user.create()`, and added a short comment explaining why the deferred promise is used instead of returning the native `user.create()` chain.

### Testing
- Ran `node --check /workspace/planet/protractor.conf.js` to verify syntax and it succeeded.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69a9e7210914832d9faefa6f6556be20)

### DIFF
--- a/protractor.conf.js
+++ b/protractor.conf.js
@@ -33,12 +33,19 @@ exports.config = {
     jasmine.getEnv().addReporter(new SpecReporter({ spec: { displayStacktrace: true } }));
     browser.params.user = user.get();
 
-    return user.create().then(function(res) {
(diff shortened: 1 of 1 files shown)


Mutugiii:
Protractor is deprecated

Issue #9082, opened 2025-09-08
Reenable linting rules


paulbert:
While upgrading to Angular 13 we also had to switch from TSLint to ESLint. Many of the linting rules were throwing errors after this change, so we temporarily disabled them to avoid making that PR too large.

Any rule that is commented out in `.eslintrc.json` will need to be evaluated and either removed or reenabled and errors fixed. The change added some rules and some rules are different in ESLint vs TSLint, so some of them may not be needed for us.

Mutugiii:
In angular 19, some linting rules have been deprecated from angular-eslint

As noted in #9754

> Some linting rules have been deprecated -> Some stylistic/formatting rules, and they should now be handled via stylistic angular or via prettier: https://eslint.style/guide/migration

PR #9806, opened 2026-03-30
all: smoother linting (connects #9082)

paulbert:
Some lint rules were disabled after the upgrade to Angular 19. Reenabling them here. See https://github.com/open-learning-exchange/planet/pull/9754/changes#diff-6884918dc8291219be508e05e28965b958c734def85324f3b53858ea4702090f for the disabled rules.

Note that for the quotes rule `{ "allowTemplateLiterals": "never" }` still allows template literals. It actually throws an error when backticks are used without a template literal or with a one line string.


Saby-Bishops (approved):

paulbert:
Jules came up with a much simpler fix, ready for review now @Mutugiii

Mutugiii (approved):

dogi (approved):

PR #9818, opened 2026-03-31
Simplify dev workflow: stop generating environment.dev and regenerate environment.ts

Saby-Bishops:
### Motivation
- Remove the special-case generated `environment.dev.ts` path and simplify the local development workflow so `npm run dev` consistently prepares environment values and runs the dev server. 
- Avoid relying on an extra generated file that diverged from `environment.ts`, while documenting the new behaviour for local developers.

### Description
- Removed the `build.configurations.dev` file replacement (the `environment.dev.ts` replacement) from `angular.json` and removed the corresponding `serve.configurations.dev` entry so Angular no longer expects `src/environments/environment.dev.ts`.
- Updated the `dev` npm script in `package.json` to `bash dev-env.sh && ng serve` (removed `--configuration dev`).
- Updated `dev-env.sh` to generate `src/environments/environment.ts` from `src/environments/environment.template` (instead of writing `environment.dev.ts`).
- Updated `README.md` to use `npm run dev`, to document that `npm run dev` regenerates `src/environments/environment.ts`, and to update related examples and restart guidance.

### Testing
- Parsed `package.json` and `angular.json` with Node JSON parsing (`node -e ...`) which succeeded.
- Validated shell syntax for `dev-env.sh` using `bash -n dev-env.sh`, which succeeded.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69cc30f00a40832d8da35f3b57eb519c)

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,11 +66,12 @@ bash couchdb-setup.sh -p 2200 -i
 Install dependencies and serve the app:
 ```
 npm install
-ng serve
+npm run dev
 ```
+`npm run dev` regenerates `src/environments/environment.ts` from `src/environments/environment.template` using values from `.env` (if present).
 
 Visit localhost:3000 to access the Planet app.
-If port 3000 is in use, try ```ng serve --port 3001```
+If port 3000 is in use, try ```npm run dev -- --port 3001```
 
 ## Chatapi Notes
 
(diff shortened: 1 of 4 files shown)


uj-sxn (approved):

Mutugiii:
Due to port differences between development environments, we are maintaining the development environment generation workflow

PR #9881, opened 2026-04-21
Deprecate generated dev environment files

Saby-Bishops:
### Motivation
- Remove the runtime/template substitution workflow for dev builds and use a static, checked-in dev environment file to simplify local development and make builds deterministic. 
- Prevent the repository from generating `src/environments/environment.dev.ts` at runtime and instead use `src/environments/environment.development.ts` as the file replacement for the Angular `dev` configuration. 
- Note the risk is medium-to-high because this changes local development and Docker-adjacent workflows that previously relied on dynamic env generation and requires teams to adopt a runtime config endpoint or `.env` injection in follow-ups. 

### Description
- Switched the Angular `dev` file replacement in `angular.json` from `src/environments/environment.dev.ts` to the checked-in `src/environments/environment.development.ts`. 
- Simplified the `dev` npm script in `package.json` to remove the template substitution step so it now runs `ng serve --configuration dev` directly. 
- Removed `dev-env.sh` and `src/environments/environment.template`, and added a static `src/environments/environment.development.ts` with the previous development defaults, and updated `chatapi/README.md` to point developers at the new file for port alignment. 

### Testing
- Attempted an Angular dev build via `ng build --configuration dev` in this environment, which failed because the Angular CLI is not installed here (test infrastructure limitation). 
- Verified via ripgrep that references to the deprecated generated artifacts (`dev-env.sh`, `environment.template`, `environment.dev.ts`) were removed and the new file replacement points to `src/environments/environment.development.ts`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69dd51248380832d82db0a0481544449)


Mutugiii:
generated dev envs are still supported

PR #9907, opened 2026-04-27
teams: smoother survey rating scaling (fixes #9906)

### LINKED ISSUE #9906: surveys UI improvements

Mutugiii:
Fixes #9906

redesign exams/surveys workflow, and introduce a visual progress tracker

https://github.com/user-attachments/assets/07782401-0d45-4356-9f13-3c595b505334


Saby-Bishops (approved):

Mutugiii:
Upgraded this to add a new customizable rating scale from 5 - 9


https://github.com/user-attachments/assets/46afd5f4-5112-4393-a53b-4df79b380d4b

paulbert (commented):
Just a few things that I've noticed:

- The progress bar is getting a little lost in the toolbar. Not sure if many people will notice it at first. If it's not really critical to see I think it's fine because this way it takes up very little space.
- For the new rating scale I think it'd be better to move the selection below the question type. It's hard to notice pushed over to the left.

Screenshot from 2026-05-04 13-50-20

- The animation appears to only be working the first time you go in a particular direction (i.e. clicking forward twice in a row results in one change with the animation and one without). Looks like Copilot might have a good explanation for this. Thinking this through might be best to do a follow up PR to fix rather than holding up the other key changes to the surveys.

paulbert (approved):

dogi (approved):

PR #9933, opened 2026-05-08
all: smoother sorting (fixes #9929)

### LINKED ISSUE #9929: surveys: select all not selecting all surveys
confirmed in my local dev
https://github.com/user-attachments/assets/3f4ffbac-6696-4bbb-85eb-57819c9d0980

similarly on prod
https://github.com/user-attachments/assets/de6d26df-01e0-4242-9881-1432d249d343

Mutugiii:
Fixes #9929.
  
- `MatTableDataSource.filteredData` is in original order, not sort+paginator order. Slicing it by `pageIndex × pageSize` returned a *different* set of rows than what the table was actually rendering whenever a sort was applied, so "select all" silently selected wrong rows and skipped visible ones.                                                                              
- Switched `surveys`, `resources`, `courses`, `meetups`, and `manager-fetch` to read from the data source's own `connect()` stream, the same array the table renders, and operate on that for select-all, indeterminate state, and filter-driven deselection.                                                                                                                         
- Surveys: parent-planet rows now render as a visibly disabled checkbox with a tooltip rather than an empty cell, so the user can see why those rows aren't selectable.                     
- Drops the now-unused `itemsShown` and `selectedOutOfFilter` call sites; helpers themselves left in place.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.22.81",
+  "version": "0.22.85",
   "myplanet": {
     "latest": "v0.54.94",
     "min": "v0.52.45"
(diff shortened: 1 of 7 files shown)


Mutugiii:
@codex review

dogi (approved):

PR #9935, opened 2026-05-12
all: replace chatapi with unified planet-gateway service (connects #9934)

### LINKED ISSUE #9934: team public surveys
Add new workflow allowing team surveys to be taken by non-authenticated users. After taking the survey, we intend to encourage users to create an account in planet and be added to the team

Mutugiii:
Connects #9934

- renamed and restructured the old Node service into planet-gateway
- kept one server with centralized CouchDB access with separate internal modules for:
  - chat (/ml/)
  - public survey operations (/api/public/)
- chatapi endpoints are unchanged
- added public survey endpoints:
  - GET /public/surveys/:teamId/:surveyId
  - POST /public/surveys/:teamId/:surveyId/submissions
- updated Docker and nginx wiring to use planet-gateway
- updated GitHub workflow/build wiring away from chatapi to new gateway
- updated some linting rules to allow push

This should allow us to extend public-scoped features across other features, including courses & resources

Currently, the link generation/revocation workflow frontend is already in place in this PR
Public survey frontend is still WIP and will be handled in another PR, as well as the post survey account creation workflow


planet-gateway/src/config/couch.config.ts
```diff
@@ -0,0 +1,22 @@
...
+import dotenv from 'dotenv';
+
+dotenv.config();
+
+const couchUrl = process.env.COUCHDB_HOST;
+const couchUser = process.env.COUCHDB_USER;
+const couchPass = process.env.COUCHDB_PASS;
+
+const couchHost = couchUser && couchPass
+  ? `http://${couchUser}:${couchPass}@${(couchUrl || 'http://couchdb:5984').replace('http://', '')}`
```
paulbert: I know that this is how we were doing it before, but I think this replace on the url protocol isn't ideal. If someone really wanted to configure this to https it'd just be silently switching back to http without the user being aware.

It's more verbose, but doing a split on the couchHost would be a way to preserve the protocol and add in the credentials. Here's what I was thinking without the other conditionals:

```
const [ protocol, domainAndPort ] = couchUrl.split('//');
const couchHost = `${protocol}//${couchUser}:${couchPass}@${domainAndPort}`;
```

paulbert (commented):
You'll need to update the chat endpoint in the frontend to reach the new chat api url:

Image

Otherwise it's looking good, and I see that the updates for the frontend will be happening in a follow up.

paulbert (approved):
The issue was in fact old API keys, so this is good to merge!

Mutugiii:
Accidental rebase on master (and push) with this branch, new PR is #9953. Master is already reverted to the initial state.

Also, following conversations with @dogi. We're renaming the `planet-gateway` service to `gateway`

PR #9939, opened 2026-05-13
all: smoother page breakpoints handling (fixes #9937)

### LINKED ISSUE #9937: app currently mixes shared viewport classification with page-specific toolbar overflow behavior.
DeviceInfoService started as a simple mobile/tablet/desktop helper, but several screens added custom TypeScript breakpoint exceptions such as 1040, 1200, and 1300 to handle local toolbar crowding. That created a few problems:
  - many components duplicated @HostListener('window:resize') just to re-read device state
  - the same breakpoint intent appeared in multiple places, sometimes in TS and sometimes in SCSS

DeviceInfoService should focus on real shared viewport categories while making responsive toolbar behavior easier to maintain.

Mutugiii:
Fixes #9937

- DeviceInfoService now uses `BreakpointObserver` as the shared reactive viewport source.
- Consumers that previously used local @HostListener('window:resize') for device updates were migrated to the shared service stream
- The shared device breakpoints were normalized back to the default 1000 / 780 / 480
- Page-specific toolbar overflow handling was moved into local layout behavior instead of custom device breakpoints
  - Resources: the selection/count summary now shrinks and truncates cleanly so the existing actions menu can switch at the normal breakpoint
  - Reports detail & MyPlanet toolbar: non-desktop controls use full-width, and stacked toolbar alignment was adjusted so dropdowns, toggle group, and Clear align better

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.22.82",
+  "version": "0.22.84",
   "myplanet": {
     "latest": "v0.55.45",
     "min": "v0.53.91"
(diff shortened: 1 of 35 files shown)


Mutugiii:
@codex review

dogi (approved):

PR #9984, opened 2026-06-09
teams: smoother survey demographics handling (fixes #9983)

### LINKED ISSUE #9983: add optional demographics form to public surveys

Mutugiii:
Fixes #9983

Add the optional demographics information form at the end of public surveys

image

### DIFF
--- a/gateway/src/modules/public/services/surveys.service.ts
+++ b/gateway/src/modules/public/services/surveys.service.ts
@@ -112,8 +112,28 @@ const normalizeAnswers = (answers: any[] = []): StoredPublicAnswer[] => answers.
   'passed': isValidAnswer(answer)
 }));
 
-const sanitizePublicSubmissionUser = (user: PublicSubmissionPayload['user']) => {
+const validatePublicSubmissionUser = (user: PublicSubmissionPayload['user']) => {
+  if (user === undefined) {
+    return null;
+  }
+
   if (!user || typeof user !== 'object' || Array.isArray(user)) {
+    return 'user must be an object';
+  }
+
+  if ('age' in user && (typeof user.age !== 'number' || !Number.isInteger(user.age) || user.age < 1 || user.age > 130)) {
+    return 'user.age must be an integer between 1 and 130';
+  }
+
+  if ('gender' in user && (typeof user.gender !== 'string' || ![ 'male', 'female' ].includes(user.gender.trim()))) {
+    return 'user.gender must be either male or female';
+  }
+
+  return null;
+};
+
+const sanitizePublicSubmissionUser = (user: PublicSubmissionPayload['user']) => {
+  if (!user) {
     return {};
   }
 
@@ -121,11 +141,11 @@ const sanitizePublicSubmissionUser = (user: PublicSubmissionPayload['user']) => 
   const age = user.age;
   const gender = typeof user.gender === 'string' ? user.gender.trim() : '';
 
-  if (typeof age === 'number' && Number.isInteger(age) && age >= 0 && age <= 130) {
+  if (typeof age === 'number') {
     sanitizedUser.age = age;
   }
 
-  if (gender === 'male' || gender === 'female') {
+  if (gender) {
     sanitizedUser.gender = gender;
   }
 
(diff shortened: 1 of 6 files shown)


Mutugiii:
@codex review

paulbert:
Tried a couple of times, but I'm not seeing the info from that form being reflected in the data export. The last two entries were made with this branch.

image

UPDATE: I see there were changes to the planet gateway service. Using the image from this branch fixes that issue.

gateway/src/modules/public/services/surveys.service.ts
```diff
@@ -108,6 +112,26 @@ const normalizeAnswers = (answers: any[] = []): StoredPublicAnswer[] => answers.
...
+const sanitizePublicSubmissionUser = (user: PublicSubmissionPayload['user']) => {
+  if (!user || typeof user !== 'object' || Array.isArray(user)) {
+    return {};
+  }
+
+  const sanitizedUser: NonNullable<PublicSubmissionPayload['user']> = {};
+  const age = user.age;
+  const gender = typeof user.gender === 'string' ? user.gender.trim() : '';
+
+  if (typeof age === 'number' && Number.isInteger(age) && age >= 0 && age <= 130) {
```
paulbert: Should at least log if not throw something if age is in the payload and not in the correct range.

I think in the frontend this is limited by a minimum year, not maximum age as well. Not ideal to have the mismatch here as in a few years a valid form value will be removed here. Also frontend rejects age 0, but this would accept that.

gateway/src/modules/public/services/surveys.service.ts
```diff
@@ -108,6 +112,26 @@ const normalizeAnswers = (answers: any[] = []): StoredPublicAnswer[] => answers.
...
+    sanitizedUser.age = age;
+  }
+
+  if (gender === 'male' || gender === 'female') {
```
paulbert: Same as above, log or throw an error if the value exists and is invalid.

src/app/exams/public-surveys/public-survey.component.html
```diff
@@ -52,6 +44,30 @@
   </div>
 </planet-exams-question-frame>
 
+<div *ngIf="survey && showDemographics && !isSubmitted" class="state-view">
```
paulbert: Is there a reason why this form has the fields flipped from the "Record Survey" version?

Generally not ideal that this is duplicated from that, but I see that is also a bit entangled with the UsersUpdateComponent so I think this is OK.

src/app/exams/public-surveys/public-survey.component.ts
```diff
@@ -45,7 +35,16 @@ export class PublicSurveyComponent implements OnInit {
   isLoading = true;
   isSubmitting = false;
   isSubmitted = false;
+  showDemographics = false;
   readonly answer = new FormControl<ExamAnswerValue>(null, { validators: examAnswerValidator });
+  readonly demographicsForm = this.fb.group({
+    birthYear: this.fb.control<number | null>(null, [
+      Validators.min(1900),
```
paulbert: It would be cleaner with the backend to match this to the current year minus 130 rather than a minimum year.

src/app/exams/public-surveys/public-survey.component.ts
```diff
@@ -126,4 +137,22 @@ export class PublicSurveyComponent implements OnInit {
...
+    const user: PublicSurveyDemographics = {};
+
+    if (birthYear !== null) {
+      user.age = new Date().getFullYear() - birthYear;
+    }
+    if (gender) {
+      user.gender = gender;
+    }
+
+    return Object.keys(user).length > 0 ? user : undefined;
```
paulbert: I think you should just return the user here. There's no check from here to the database on whether the user object is undefined other than to turn it into an empty object if it is.

By explicitly setting it to undefined if it's an empty object, you're setting an expectation that that matters somewhere in the application.

paulbert (approved):
Looks like it's working as expected, nice work!

I've got some comments on error handling, formatting, and code quality, but I'll go ahead and approve.

dogi (approved):

PR #10026, opened 2026-06-25
teams: smoother surveys button resizing (fixes #10023) 

### LINKED ISSUE #10023: Fix survey action buttons size to contain all embedded text
The `Generate Public Link`  button currently has an icon overflowing outside of its visual range. Slightly increasing the button width in this section will ensure that all text and icons remain fully visible while maintaining the page's visual consistency. 

Image

RyanS4:
Fixes #10023 
- Increase the width of the action buttons for surveys to make all text and icons visible

## Before
image

## After
{E7CA2B41-2D74-442F-94FB-D5B0508FF1A3}


Mutugiii (commented):
Looks good. However, I noticed we have the same issue in smaller screens that we should resolve as well.

https://github.com/user-attachments/assets/a76f8b01-13cd-4079-a6ef-d23805dc18b8

RyanS4:
Made some changes so the buttons will stack in a neat column at a fixed size. On smaller screens, only the icon is shown in the center of the button (this can be changed if needed).

{7F6DA181-20A6-4372-BDAB-8C6BCC507C3F}

{D5A79210-A365-4446-8F19-42A61E7E439A}

Mutugiii (approved):
Looks good! Made some adjustments before merging:
- Hiding the `courses info` row in team surveys since we don't expect that field to be in use there
- Adding a new icon for `revoke` action as it was missing one.
- Expanded the survey table min-width to avoid the button cutoff in the 1000 - 1200 screens

dogi (approved):

PR #10046, opened 2026-07-01
achievements: add certifications to achievements (fixes #8657)

### LINKED ISSUE #8657: achievements: add summary of certifications
Similar to how there's a field to summarize achievements, there could be a similar one for certifications.

Image

RyanS4:
Fixes #8657 
Similar to achievements, users can now provide a summary of their certifications and maintain a manual list of certifications.

- Updated UsersAchievementsUpdateComponent with new form controls and dialogs
- Updated UsersAchievementsComponent to display manual certifications
- Renamed existing auto-earned certifications to 'earnedCertifications' to avoid conflict
- Updated PDF generation to include manual certifications and summary
- Fixed manual certification list items to correctly show links by making them identical to achievement list items
- Updated UsersAchievementsService to include new fields in empty-check logic
- Added i18n tags for new labels and headers

image

image

image

### DIFF
--- a/src/app/users/users-achievements/users-achievements-update.component.html
+++ b/src/app/users/users-achievements/users-achievements-update.component.html
@@ -58,6 +58,10 @@
           <mat-label i18n>Summary of Achievements - Briefly summarize your achievements and add related materials below</mat-label>
           <planet-markdown-textbox class="full-width" [formControl]="editForm.controls.achievementsHeader"></planet-markdown-textbox>
         </mat-form-field>
+        <mat-form-field class="full-width mat-form-field-type-no-underline">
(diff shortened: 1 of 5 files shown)


Mutugiii:
The initial goal of the issue was not to support certifications as a standalone new achievements category, but rather to support the downloading of certificates, which is an existing Planet Beta feature.

For instance:

`Certifications` is accessible via the `Manager Settings`:
image

The certifications available in my community:
image

The certifications earned:
image


To get the certification, a learner must have completed the relevant courses for that certification. We can also work on showcasing that a course is part of a certification in the courses;


There were 2 relevant issues that tried to tackle the certifications previously:
- https://github.com/open-learning-exchange/planet/pull/9480
- https://github.com/open-learning-exchange/planet/pull/9449

PR #10009, opened 2026-06-17
all: smoother package json locking (fixes #9991)

Mutugiii:
Fixes #9991

Baseline security against npm supply chain attacks

PR #10077, opened 2026-07-06
actions: smoother testing (fixes #10117)

### LINKED ISSUE #10117: all: Add testing actions on github

paulbert:
Adding linting and test actions to GitHub actions

Adding them in to their own "testing" job. Let me know if there's a better place for this.


paulbert:
I thought #10009 had been merged, so had including caching for the node packages based on the package lock.

Since it hasn't I removed the gitignore lines for the package lock here, too. Maybe commit that first then this one.

.github/workflows/planet.yml
```diff
@@ -36,6 +36,29 @@ jobs:
...
 
+  testing:
+    name: Tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js 22
+        uses: actions/setup-node@v3
```
Mutugiii: I think we can use `@v4` to target node 20

Mutugiii (commented):
I believe we have 3 `jasmine` packages in the package.json that are not required by vitest

I also noticed we have a `karma.conf.js` file in the root directory and a `src/test.ts` that was used to setup Karma. I'm not sure if we want to retain those files

Mutugiii (approved):

dogi (approved):

PR #10148, opened 2026-07-20
all: add shared navigation service (fixes #10146)

### LINKED ISSUE #10146: routing overhaul: make routing history-aware
Navigation is inconsistent across features mounted at multiple paths. Hard-coded routes can move users out of their dashboard or team context, back buttons often navigate to a fixed parent instead of the previous page, and team tab state is lost on refresh or deep links.

Standardize routing so that back navigation follows browser history with a safe parent fallback.

Mutugiii:
Introduces a shared `NavigationService` that standardizes in-app back navigation across Planet.

Back controls now:
- Follow browser history when a previous in-app page exists.
- Use a deterministic hierarchical fallback for refreshes and deep links.
- Replace the current history entry during fallback navigation, preventing back loops.
- Distinguish browser back from forward navigation.
- Avoid leaving the application when history originated before the current document.

## Changes
- Add `NavigationService`.
- Track Angular navigation IDs and their browser-history positions.
- Handle popstate, forward navigation, `replaceUrl`, and unknown pre-document history entries.
- Initialize navigation tracking during application startup.
- Migrate generic back handlers across chat, feedback, health, manager, meetups, submissions, surveys, and user features.
- Add unit and real-Router integration coverage for the navigation behavior.

## Summary by CodeRabbit

* **New Features**
  * Introduced shared back-navigation across major screens, with browser-history support and sensible fallbacks.
  * Updated the feedback toolbar back button to use the new navigation behavior.

* **Bug Fixes**
  * Improved deep-link handling and prevented back-navigation loops.
  * Made user search updates history-safe by replacing the current history entry.

* **Tests**
  * Added coverage for navigation, feedback back actions, chat fallback behavior, and debounced user search.


Mutugiii:
@codex review

src/app/shared/navigation.service.ts
```diff
@@ -0,0 +1,60 @@
...
+      }
+    });
+  }
+
+  back(fallback: any[] | string = [ '/' ], extras: NavigationExtras = {}) {
+    if (this.index > 0) {
+      this.location.back();
+    } else if (typeof fallback === 'string') {
+      // Serialized URL fallback (e.g. a stored return URL with matrix params)
+      this.router.navigateByUrl(fallback, { ...extras, replaceUrl: true });
```
paulbert: Commenting here because this is the function where I think we should consider removing `replaceUrl`, on this line and a few lines below.

src/app/shared/navigation.service.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+@Injectable({ providedIn: 'root' })
+export class NavigationService {
+
+  private index = -1;
+  private idToIndex = new Map<number, number>();
+  private currentNav: { id: number, trigger: string, restoredId: number | null, replaceUrl: boolean } | null = null;
+
+  constructor(private router: Router, private location: Location) {
+    this.router.events.subscribe(event => {
```
paulbert, line 43: This seems to be a working strategy, I just wonder if we really need to be storing a second browser history since the Angular router is also storing history.

If all we need to do is intercept when the user is going off Planet I wonder if we could just do that instead?

paulbert (commented):
I think this behavior is not ideal, we might want to consider removing `replaceUrl` from the fallback for navigationService.

What I'm doing in the video:
1. Refresh on the /feedback page.
2. Click the on site back button twice to go to /manager -> /
3. Click on the browser back button twice to go to /manager -> /myDashboard

I realize after going through the video I didn't need to do two levels on the back button, just one works. When using the fallback the page the user is on gets removed from the history so the browser back takes to the previous page, which could be off app.

I know mimicking the browser history would more have that page inserted into the history as next for the browser "forward", but I think in this case since the back button might take the user off app it'd be good to keep that page in the history.

Also I see the note about not wanting to loop back into that page. I'm not seeing that happen in practice testing locally, but that would be a valid reason to keep that parameter.

https://github.com/user-attachments/assets/edaf464d-82f2-42ab-8f03-24fb076d6272

Mutugiii:
@paulbert I removed replaceUrl. The fallback now keeps the original page in browser history, while Planet’s Back button still avoids looping through fallback pages.

I also fixed an issue with unsaved changes in the chat (and pushed a new spec file for it)

PR #10156, opened 2026-07-22
all: smoother android app promoting (fixes #10221)

### LINKED ISSUE #10221: Add Android app promotion dialog for mobile users

dogi:
image

 Adds an extra Android app-install prompt when users open public surveys, while retaining the existing Android banner. Both surfaces offer myPlanet and myPlanet Lite.

  Dismissal of the dialog is stored in `sessionStorage`, so it remains closed after refreshes in the same tab but may reappear in a new session.

  ## Changes

  - Adds a public-survey Android install dialog.
  - Expands the existing banner to include both apps.
  - Centralizes app names, descriptions, and Play Store links.
  - Adds accessibility labels and responsive styling.
  - Adds tests and Spanish translations for all newly extracted strings.

  ## Steps to test

  1. On an Android device or Android user agent, open:
     `/survey/<teamId>/<surveyId>`
  2. Verify the dialog shows both apps and their correct Play Store links.
  3. Close the dialog and refresh; it should not reappear in the same tab.
  4. Clear `sessionStorage` or start a new session; it should appear again.
  5. Use a non-Android user agent; the dialog should not appear.
  6. Open the authenticated Planet interface on Android and verify the existing banner shows both apps and can be closed.
  7. Repeat using the Spanish locale and verify the translated text.

  ## Validation

  - Unit tests
  - Angular lint
  - i18n extraction check
  - Spanish localized build

https://claude.ai/code/session_01JEkpsKj5Erfasi6PTaFXed

## Summary by CodeRabbit

* **New Features**
  * Added an Android app promotion dialog for myPlanet and myPlanet Lite with “Install” and “Not now”.
  * Added a conditional Android app prompt on public surveys that shows once per session after dismissal.
  * Updated the home banner to render multiple Android app links dynamically.

* **Bug Fixes**
  * Improved accessibility with clearer aria-labels for banner close and app install actions.

* **Translations**
  * Updated Android banner and prompt text (including Spanish) for both app variants and related actions.

### DIFF
--- a/src/app/shared/android-app-prompt.service.ts
+++ b/src/app/shared/android-app-prompt.service.ts
@@ -8,6 +8,7 @@ import { DialogsAndroidAppComponent } from './dialogs/dialogs-android-app.compon
   providedIn: 'root'
 })
 export class AndroidAppPromptService {
+  // Keep the prompt dismissed across reloads and restores for the lifetime of this tab's page session.
   private readonly dismissedSessionStorageKey = 'planet-android-survey-app-prompt-dismissed';
 
   constructor(
--- a/src/app/shared/android-app-prompt.service.ts
+++ b/src/app/shared/android-app-prompt.service.ts
@@ -8,33 +8,33 @@ import { DialogsAndroidAppComponent } from './dialogs/dialogs-android-app.compon
   providedIn: 'root'
 })
 export class AndroidAppPromptService {
-  private readonly promptDismissedKey = 'planet-android-survey-app-prompt-dismissed';
+  private readonly dismissedSessionStorageKey = 'planet-android-survey-app-prompt-dismissed';
 
   constructor(
     private deviceInfoService: DeviceInfoService,
     private dialog: MatDialog
   ) {}
 
-  maybePrompt() {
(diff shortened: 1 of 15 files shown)


src/app/shared/android-app-prompt.service.ts
```diff
@@ -0,0 +1,42 @@
...
+      return;
+    }
+
+    this.dialog.open(DialogsAndroidAppComponent, { maxWidth: '90vw', width: '400px' })
+      .afterClosed().subscribe(() => this.markDismissedForSession());
+  }
+
+  private wasDismissedThisSession(): boolean {
+    try {
+      return sessionStorage.getItem(this.dismissedSessionStorageKey) === 'true';
```
paulbert, line 30: Could be useful to leave a comment here since my understanding of a "page session" was that it does not persist over page reloads

```
A page session lasts as long as the tab or the browser is open, and survives over page reloads and restores.
See https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage
```

paulbert (approved):
This looks good! I'd just add a comment since it's the first time we're using `sessionStorage` and it could be helpful to be clear about when it resets.

Mutugiii (approved):

PR #10173, opened 2026-07-27
Package lock changes

paulbert:
After running `npm install` a bunch of packages were removed from the lockfile. Not sure if this is expected behavior.

Probably fine to just merge since it looks like it's all deletes, but we should look into this more to make sure we're using the right process.


paulbert:
Or maybe it's not fine? Tests just failed saying the lockfile is missing packages.

Mutugiii:
I tested on my current master branch, but I can't reproduce it. Running npm install only updated the versions on my end.

PR #10183, opened 2026-07-29
teams: smoother member profile cards styling (fixes #10182)

### LINKED ISSUE #10182: update app-wide profile cards styling

Mutugiii:
Fixes #10182

Reworks the member card into a self-contained tile that owns its own surface, then reuses it everywhere members are listed: community leaders, team members, joinrequests, enterprises and services.

- Moves the card surface onto `planet-teams-member`, dropping the `mat-card` wrappers in teams and community
- Renders join requests through the same tile instead of duplicate `mat-card` markup, removing a second copy of `openMemberDialog`
- Shows generated initials for members without a profile picture, plus a leader band and star badge
- Adds `$primary-lighter-text` / `$accent-lighter-text` so bands and chips reuse the established `-lighter` fills with text that clears AA
- Softens the profile dialog header, which is a nested panel rather than a page header - Fixes contrast on the resources step number and upload icon (2.2:1 → 6.2:1)
- Stops a failing avatar from being re-requested on every change detection pass, and opens the profile dialog without auto-scrolling past its header
- Adds unit tests for the tile

## Before

Community Leaders
image

Teams
image
image

## After

Community Leaders
image

Teams
image

## Summary by CodeRabbit

* **New Features**
  * Redesigned team and community member displays as interactive tiles with avatar initials, leader badges, tasks, activity details, and contextual action menus.
  * Added reusable join-request tiles with accept/reject actions and configurable request management.
* **Bug Fixes**
  * Improved grid sizing with automatic row heights to prevent content clipping.
  * Updated lighter-surface text colors for improved contrast.
* **Style**
  * Refined profile toolbar styling in dialogs and added accessible focus states.
* **Tests**
  * Added coverage for member names, initials, avatar fallbacks, identity states, and profile dialog behavior.


RyanS4 (changes requested):
We should make the profile avatar icon accessible via tabbing, since it is an embedded clickable element, and this is a standard we are attempting to implement across the codebase.

Image

Mutugiii:
> We should make the profile avatar icon accessible via tabbing, since it is an embedded clickable element, and this is a standard we are attempting to implement across the codebase.
> 
> Image

@RyanS4 Updated to make the profile image tab-able

RyanS4 (approved):

dogi (approved):

PR #10194, opened 2026-07-30
all: smoother eqeqeq less underscore linting (connects #9082)

paulbert:
Adds the eqeqeq and no underscore dangle lint rules back in. Eqeqeq didn't need any changes to fix.

We've used leading underscores for private methods (either explicitly with the `private` keyword or otherwise). There is now a "#" decorator that can be used in most cases which makes the method private at compile time and at runtime. I went through and switched to that or in the few cases where the variable was not private just removed the underscore.

I also tried to make sure the getter/setter pattern used the getter instead of the private method throughout the component.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search-selection handling so Clear buttons correctly reflect whether filters are empty.
  * Improved advanced filtering for empty search criteria.
  * Preserved calendar event display behavior with consistent event handling.

* **Refactor**
  * Strengthened internal data encapsulation across courses, resources, forms, dashboards, and other components without changing their public behavior.
  * Standardized linting rules to catch unsafe equality checks and disallowed private-field naming patterns.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -155,7 +155,7 @@
         "@typescript-eslint/unified-signatures": "error",
         "complexity": "off",
         "constructor-super": "error",
-        // "eqeqeq": ["error", "smart"],
+        "eqeqeq": ["error", "smart"],
         "guard-for-in": "error",
         "id-match": "error",
         "import/no-deprecated": "warn",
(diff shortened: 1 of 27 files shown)


Mutugiii (commented):
Seems the changes currently break the courses & resources search because `isEmpty` is now processed as a filter value instead of being excluded as metadata.

Mutugiii (approved):

Mutugiii:
looks good!

paulbert:
@Mutugiii Rebased and ready now

paulbert:
To link to the issue: #9082

dogi (approved):

PR #10198, opened 2026-07-31
courses: smoother steps submitting (fixes #10193)

### LINKED ISSUE #10193: Submissions: Label submissions for deleted courses and steps in grading view
### Summary
Users with grading permissions are currently able to see submissions for exams under deleted courses or deleted steps in the Submissions view (displayed with empty course titles and step `0`).
### Problem / Root Cause
In `SubmissionsService.updateSubmissions()`, course existence is only verified for `pending` exam submissions (`sub.status !== 'pending'` returns `true`). As a result, `requires grading` and `complete` submissions for deleted courses pass through to the table. Additionally, step existence within existing courses is never checked, leaving orphaned submissions in the queue if a step is removed.
### Proposed Solution
Label these submissions rather than hiding them — they are valid historical records, and filtering removes them from every view using this shared component, including a learner's own submission history (see discussion in #10198).

Update `appendCourseInfo()` to label a submission whose parent course or step no longer exists, and remove the filter in `SubmissionsService.updateSubmissions()` that silently drops pending exam submissions for deleted courses.

Image

RyanS4:
Fixes #10193

Submissions whose course or step has since been deleted are valid historical records, so they stay in the grading list and are labeled instead of being hidden (per discussion on this PR).

- `appendCourseInfo` labels an unresolved course as **Course not available** and its step as **Step not available**.
- The claim is scoped to submissions originating on this planet (`source`), since `courses` doesn't replicate between planets — submissions synced in from another planet render blank rather than asserting a course was deleted.
- `stepNum` stays numeric so the Course Step column still sorts; the label is a separate `stepLabel` field.
- Removed the filter in `SubmissionsService.updateSubmissions()` that silently dropped pending exam submissions for deleted courses — the root cause named in #10193, which hid rows from every view using this shared component.


Mutugiii:
Nice catch on the null-safe guard—|| { doc: { steps: [] } } was hiding the underlying problem. From a product perspective, though, should we label these rows instead of filtering them out?
 
Grading an orphaned submission still works: updateSubmission saves the grade regardless. Only updateProgress is skipped when the course GET returns a 404 (submissions.service.ts:158–166). These therefore aren’t broken rows, they’re valid historical records. Filtering them removes them from every view using this shared component, including the learner’s own submission history. Could we keep them visible and label them as belonging to a removed course or step?

RyanS4:
@Mutugiii 

What do you think of this behavior schema?

{9DBE8A20-CB08-43FC-A870-92FCDC01018A}
{F8D81B1E-5366-43C6-A2DF-52B23FE936F7}

Mutugiii:
> @Mutugiii
> 
> What do you think of this behavior schema?
> 
> {9DBE8A20-CB08-43FC-A870-92FCDC01018A} {F8D81B1E-5366-43C6-A2DF-52B23FE936F7}

I think `Course not available` works better. Also, generally watch out, as there are independent surveys that are not attached to courses

Mutugiii (approved):

dogi (approved):

PR #10202, opened 2026-07-31
all: smoother feedback forms saving (fixes #10201)

### LINKED ISSUE #10201: Feedback: smoother feedback unsaved changes handling
Currently, when a user is typing a feedback message or bug report in the feedback popup dialog (`[planetFeedback]`) and accidentally clicks off the modal (backdrop click) or presses Escape, their typed message and selections are lost immediately with zero recovery options. Implementing these suggestions will ensure users do not accidently delete their draft feedback, can go back to look at the issue they are reporting without having to scrap their draft, and can ensure their intention when ready to cancel.

Image

RyanS4:
Fixes #10201

- Adds `confirmRef` helper to `dialogs-form.service.ts` to return dialog references for component value subscriptions
- Updates cancel action button in `dialogs-form.component.html` to pass `'cancel'` result on explicit click
- Adds `setDraftFeedback`, `getDraftFeedback`, and `clearDraftFeedback` methods to `feedback.service.ts`
- Updates `openFeedback` in `feedback.directive.ts` to populate dialog inputs from active feedback drafts
- Binds `modalForm.valueChanges` in `feedback.directive.ts` to continuously persist typed message, priority, and type selections
- Updates `afterClosed` subscription in `feedback.directive.ts` to clear drafts on explicit cancel or submit while preserving drafts on backdrop click or escape key blur

## Summary by CodeRabbit

* **New Features**
  * Added unsaved-change protection to forms, prompting users before closing dialogs with pending edits.
  * Applied confirmation behavior to Cancel actions, backdrop clicks, and Escape key presses.
  * Forms without pending edits can still be closed immediately.

* **Bug Fixes**
  * Prevented accidental loss of unsaved form changes when dismissing dialogs.
  * Updated feedback dialogs to use the enhanced confirmation behavior.


Mutugiii:
I don't think we should store the draft at all, since the toolbar's planetFeedback lives in the app shell (home.component.html:61). That directive instance sticks around across navigation, so a draft typed on one page can come back on another, so you'd get a message about one item filed against a different one.

I think we should just apply the unsaved changes prompt here and have the user confirm they want to lose the data.

src/app/shared/dialogs/dialogs-form.service.ts
```diff
@@ -1,4 +1,5 @@
 import { Observable } from 'rxjs';
+import { map } from 'rxjs/operators';
```
Mutugiii: This import is not being referenced in the diff

Mutugiii (commented):
LGTM! One minor nit to be addressed

RyanS4:
@Mutugiii fixed!

Mutugiii (approved):

dogi (approved):

PR #10240, opened 2026-08-04
actions: smoother workflow gateway docker npm caching (fixes #10242)

### LINKED ISSUE #10242: gateway builds occasionally failing/stalling
### Problem

Every `Gateway Builder` run rebuilds the gateway image from scratch. Both `npm ci` steps in `docker/gateway/Dockerfile` re-download the full dependency tree on every build, for all three arches (amd64, arm, arm64), because nothing is cached between runs. With the arm builds running under QEMU this is both slow and network-heavy, and it intermittently fails with `ECONNRESET` partway through the install — a failure that has nothing to do with the commit being built and usually passes on re-run.

Mutugiii:
## Summary

Two separate problems with the ARM images, plus the layer caching this PR started as.

## 1. ARM images were published as v7 but are ARMv6

A Pi Zero / Zero W advertises `linux/arm/v6`, found no matching entry in the manifest list, and could not pull `chatapi` or `db-init` — even though the binaries inside would have run fine.

They are ARMv6 content:

| Check on `treehouses/node-tags:arm-3.21` | Result |
|---|---|
| `/etc/apk/arch` | `armhf` — Alpine's **ARMv6** port |
| `readelf -A /usr/bin/node` | `Tag_CPU_arch: v6KZ`, `Tag_FP_arch: VFPv2` (ARM1176JZF-S = Pi 1 / Zero) |
| image config | `variant: v7` ❌ |

`docker manifest annotate` passed no `--variant`, so the list entry inherited the variant from the image config. That config is wrong upstream: `treehouses/alpine` builds with `--platform=linux/arm`, and bare `linux/arm` normalises to v7, overwriting the v6 of its balena base. `treehouses/nginx` passes no `--platform`, which is exactly why `planet:latest` was labelled correctly while `planet:chatapi` was not.

Fixed here rather than upstream because manifest list entries are set by `annotate` independently of the inherited config, so this repo can publish correct metadata on its own.

Both manifest paths — the per-commit tags and the versioned/release tags that devices actually pull — now pin `--variant v6` and re-read the pushed manifest to confirm it, so this class of mislabel fails the build instead of shipping silently.

**No current device loses anything** — v6 entries are matched by v6, v7 and v8 hosts alike. A 32-bit Pi 4 advertises `linux/arm/v7, linux/arm/v6` and already runs the v6 `planet:latest`.

## 2. Gateway arm64 builds stall for six hours

Roughly one run in six. The hung job's log shows both `npm ci` steps emitting their `EBADENGINE` warning ~7s in, then producing nothing at all until cancellation. npm's install is heavily parallel (15 default sockets plus libuv's threadpool) and that concurrency deadlocks under qemu-aarch64.

- **`timeout-minutes` + `fail-fast: false`** — bounds the damage. The default is 360 minutes, which is exactly the six hours being burned; the slowest real build is 2m49s. `fail-fast` was also cancelling sibling architectures mid-flight.
- **Gateway cross-compile** — the actual fix. The gateway is pure JS for Linux (of 410 packages the only one with an install script or `os` constraint is `fsevents`, macOS-only), so the builder is pinned to the native image via a new `BUILD_IMAGE` arg and the result copied into the target-arch stage. Both stages now declare `WORKDIR /` rather than inheriting it, since the cross-stage copy of `/node_modules` and `/dist` depends on it. No npm runs under emulation.

## 3. Registry layer caching (original scope)

`--cache-from` / `--cache-to type=registry`, per-arch scope, written only by master and releases so branch builds read but never clobber. Measured locally on gateway: **cold 32s → cached 5s**.

Note the `--mount=type=cache` npm mounts help local rebuilds but not CI — buildx creates a fresh builder per job, and cache mounts are not carried by any `--cache-to` backend. The CI win here is layer caching.

## Verification

- **Manifest variant** — against a local `registry:2` with the real base images: source config `arm/v7` → published entry `arm/v6`. Without the flag it reproduces as `v7`, and the guard correctly fails on it. Both directions tested.
- **Cross-compile** — `npm ci` + `npm prune --omit=dev` on the native builder vs the old emulated `npm ci --omit=dev`: **268 entries each, `diff` clean**.
- **No regression on current hardware** — verified on a 32-bit Pi 4 running the stack: `planet:local` resolves `arm/v6`, `chatapi-local` and `db-init-local` resolve `arm/v7`, all three run.

## Still needs

Several consecutive green arm64 runs before the cross-compile should be believed — at a ~16% stall rate a single pass proves nothing.

https://claude.ai/code/session_01Qrr76NZS7TAJqX63XFwiCf

## Summary by CodeRabbit

* **Reliability**
  * Improved build workflow stability with job time limits and continued execution across supported architectures.
  * Added validation to ensure ARMv6 images are correctly published for standard and release deployments.

* **Performance**
  * Improved container build caching to help speed up future builds.
  * Reduced unnecessary production container dependencies, resulting in leaner runtime images.

* **Compatibility**
  * Strengthened multi-architecture image publishing for ARMv6 devices.


Mutugiii:
Both bot reviews (Codex P1, CodeRabbit Major) flag the same thing: that `docker buildx build` needs an explicit `--platform` or the arm/arm64 rows will be built as amd64. That premise is right in general but does not hold here, and acting on it would regress `planet.yml`.

**The arm bases in this repo are single-platform manifests, not manifest lists.** `treehouses/node-tags:arm-3.21` is `application/vnd.docker.distribution.manifest.v2+json` with `arch=arm, variant=v7`. The registry returns that image whatever platform is requested, so the base pins the stage platform and QEMU executes the `RUN` steps. `--platform` is redundant. Buildx only falls back to the daemon platform when the base *is* a manifest list — which is the amd64 row's `treehouses/node:3.21`, where linux/amd64 is the correct pick anyway.

**Verified against the published images rather than the docs.** Reading the config blobs straight from the registry:

| tag | arch | variant |
|---|---|---|
| `arm-chatapi-0.23.77-…-master-fc613b4d` (before this PR) | arm | v7 |
| `arm-chatapi-0.23.77-…-<this branch>` (after) | arm | v7 |
| `arm64-chatapi-…` before / after | arm64 | — |
| `arm-0.23.77-…-master-fc613b4d` (planet, before) | arm | **v6** |
| `arm-0.23.77-…-<this branch>` (planet, after) | arm | **v6** |

Identical before and after in every case. This PR changes the builder driver and the cache backend; it does not touch platform resolution.

**The specific fix suggested would break armv6.** `--platform linux/${{ matrix.arch }}` expands to `linux/arm`, which buildx normalizes to `linux/arm/v7`. But `planet.yml`'s arm base is `treehouses/nginx-tags:arm`, which is **v6**, and its published arm image is v6 today. Forcing v7 there would mismatch the base image and drop armv6 hardware (Pi 1 / Pi Zero) — the devices most likely to be running a community server.

Two smaller points on the CodeRabbit comment: `arch: amd64` / `arch: arm` / `arch: arm64` already exist in the matrix (`planet.yml:167-172`), and switching the arm jobs to `--push` would break the `docker tag $repo $branchrepo` line that follows, which needs the image in the local store.

---

docker/gateway/Dockerfile
```diff
@@ -1,24 +1,31 @@
...
+ARG BUILD_IMAGE=treehouses/node:3.21
 
-FROM ${BASE_IMAGE} as builder
+FROM ${BUILD_IMAGE} AS builder
 LABEL maintainer="dogi@ole.org,mutugiii@ole.org"
 
 COPY ./gateway/package.json ./gateway/package-lock.json ./
 
-RUN npm ci
+RUN --mount=type=cache,id=npm-builder,target=/root/.npm npm ci
```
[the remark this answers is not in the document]
Mutugiii: Done in 86cefa64 — `WORKDIR /` declared in both stages.

Worth noting it held already: `treehouses/node:3.21` reports `WorkingDir: "/"`, and the built image resolves `/dist/index.js` and `/node_modules/express` correctly. But the reliance was implicit and pre-existing (the previous `COPY --from=builder /dist` depended on it too), so making it explicit is the right call.

docker/db-init/docker-entrypoint.sh
```diff
@@ -18,7 +18,44 @@ until curl $COUCHDB_HOST || [ $WAIT_TIME -eq 600 ]; do
...
+    prefixes=('')
+  fi
+
+  local entry key value prefix
+  for entry in "${keys[@]}"; do
+    key="${entry%%|*}"
+    value="${entry#*|}"
+    for prefix in "${prefixes[@]}"; do
+      path="$prefix/_config/$key"
+      if ! curl -sf -X PUT "$base$path" -d "$value" -o /dev/null; then
```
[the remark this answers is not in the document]
Mutugiii: Done in ad353078 — `-sSf --connect-timeout 5 --max-time 30` on both the membership probe and the config writes, and `mktemp` instead of the fixed `/tmp` path.

Verified: an unreachable host now fails in ~10s with `curl: (28) Connection timeout after 5001 ms` rather than hanging. Adopting `-S` was a good call — the failure message was previously swallowed.

docker/db-init/docker-entrypoint.sh
```diff
@@ -18,7 +18,44 @@ until curl $COUCHDB_HOST || [ $WAIT_TIME -eq 600 ]; do
...
+# CouchDB 2.x keeps config per cluster node, 1.x keeps it globally; mirror both paths.
+add_cors_to_couchdb() {
+  local base="$1" node path
+  local -a keys=(
+    'httpd/enable_cors|"true"'
+    'cors/origins|"*"'
+    'cors/credentials|"true"'
+    'cors/methods|"GET, PUT, POST, HEAD, DELETE"'
+    'cors/headers|"accept, authorization, content-type, origin, referer, x-csrf-token"'
+  )
```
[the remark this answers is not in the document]
Mutugiii: Not applying this one — CouchDB does not emit a literal wildcard when credentials are enabled, it echoes the request origin. Against CouchDB 2.3.1 with exactly this config:

```
$ curl -D - -H "Origin: http://example.com" .../_all_dbs
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: http://example.com
```

That is the spec-compliant behaviour, so the pairing works.

Also worth flagging for context: these five values are a verbatim port of what the `add-cors-to-couchdb` package this commit replaces has always set. I ran both implementations against fresh CouchDB 2.3.1 instances and diffed the resulting config — identical. So this is existing production configuration, not new policy. Tightening `cors/origins` to concrete origins may well be worth doing, but it is a deliberate security change that would alter behaviour for existing deployments, and belongs in its own PR rather than riding along with a build refactor.

docker/gateway/Dockerfile
```diff
@@ -1,24 +1,37 @@
 ARG BASE_IMAGE=treehouses/node:3.21
+# The builder stays on the native (amd64) image on every arch. npm and tsc produce
+# arch-independent output, and running them under QEMU is what wedges the arm64 builds.
+ARG BUILD_IMAGE=treehouses/node:3.21
 
-FROM ${BASE_IMAGE} as builder
+FROM ${BUILD_IMAGE} AS builder
```
paulbert: This just rewrote `BASE_IMAGE` variable as `BUILD_IMAGE`

docker/gateway/Dockerfile
```diff
@@ -1,24 +1,37 @@
...
+WORKDIR /
+
 RUN apk add --no-cache poppler-utils
 COPY ./gateway/package.json ./gateway/package-lock.json ./
 
-RUN npm ci --omit=dev
+# node_modules is built on the native builder and copied in, so no npm runs under
+# emulation. This holds only while every dependency is pure JS — a package with a
+# native binary would need to be installed on the target arch instead.
+COPY --from=builder /node_modules ./node_modules/
```
paulbert, line 26: Do we need to move in `node_modules` into the final image? That would increase the size of the image.

docker/gateway/Dockerfile
```diff
@@ -1,24 +1,37 @@
...
 COPY ./gateway/package.json ./gateway/package-lock.json ./
 
-RUN npm ci
+RUN --mount=type=cache,id=npm-builder,target=/root/.npm npm ci
 
 COPY ./gateway/src ./src/
 COPY ./gateway/tsconfig.json ./
 
 RUN npm run build
+RUN --mount=type=cache,id=npm-builder,target=/root/.npm npm prune --omit=dev
```
paulbert, line 16: I don't know that we need a prune step here? At least not if we exclude `node_modules`.

paulbert: Digging more agree that `node_modules` should be copied over to the final image so this step is good.

docker/gateway/Dockerfile
paulbert: Digging a bit more coming around to this saving an extra npm install that happens implicitly when the node app is run.

dogi (approved):

PR #10245, opened 2026-08-05
all: smoother object shorthand linting (connects #9082)

paulbert:
Written using an OpenCode free model (Big Pickle). I was intending to use a local model, but didn't realize OpenCode defaulted to a free model.

Anyway adding back in a few of the lint rules, here's the model outputs after running a few queries:

Added no-var: "error" to eslint.config.mjs:243 and fixed the only violation by changing declare var module to declare const module in src/typings.d.ts:2. Lint passes.

Added object-shorthand: ["error", "always"] to eslint.config.mjs:244 and auto-fixed all 25 violations with ng lint --fix. The fixes convert key: value to shorthand (value), which is semantically identical. Lint passes.

## Summary by CodeRabbit

* **Refactor**
  * Standardized object property shorthand throughout the application.
  * Updated declaration syntax for improved code consistency.
  * Added linting rules to discourage legacy variable declarations and enforce shorthand syntax.

* **Bug Fixes**
  * No user-facing behavior changes.


paulbert:
Forgot to link to the issue, #9082

Mutugiii (approved):

dogi (approved):

PR #10252, opened 2026-08-06
all: smoother prefer arrow const linting (connects #9082)

paulbert:
Readding two more rules: `prefer-arrow` and `prefer-const`. `prefer-arrow` seems like another preference, but good to keep things consistent.

I'm not seeing any fixes for `prefer-const`, so that one we're following already I think.

connects #9082 

## Summary by CodeRabbit

- **Refactor**
  - Standardized TypeScript functions to use arrow-function syntax and `const` declarations.
  - Existing application behavior, file processing, charting, database access, PDF loading, and query utilities remain unchanged.

- **Chores**
  - Enabled ESLint rules to enforce preferred arrow functions and constant declarations.


src/app/shared/chart-utils.ts
```diff
@@ -50,16 +50,20 @@ export function createChartCanvas(width = 300, height = 400): { canvas: HTMLCanv
...
+  message = 'No data available'
+): string => {
   ctx.fillStyle = '#666666';
   ctx.textAlign = 'center';
   ctx.textBaseline = 'middle';
   ctx.font = '16px sans-serif';
   ctx.fillText(message, canvas.width / 2, canvas.height / 2);
   return canvas.toDataURL('image/png');
-}
+};;
```
Mutugiii: minor nit: stray semicolon

Mutugiii (approved):

dogi (approved):

Issue #9369, opened 2025-11-24
error running create-design-doc


Mutugiii:
<img width="3406" height="506" alt="Image" src="https://github.com/user-attachments/assets/0ceeca24-ca5b-43fd-b601-8e3c8c0925c8" />

Issue #9117, opened 2025-09-17
add claude code workflow


Mutugiii:
Add claude code workflow to review PRs & support claude code in github

Issue #8234, opened 2025-02-13
docs: Review the Style Guide


paulbert:
The style guide was last updated 7 years ago and likely needs updating. Opening this issue as a place to put some feedback, suggestions, questions, comments, etc.

There are a few points from the now closed and stale #448 to kick off:

- I think the reference to variables.scss should be changed to _variables.scss
- I'm not seeing the theme mentioned. That could be added so people don't hardcode colors and use theme variables.
- Are people making SVGs for icons anymore? That was a reference to a specific instance I believe that was resolved.

PR #10264, opened 2026-08-12
all: smoother agents assisting (fixes #10270)

### LINKED ISSUE #10270: docs: add Agent Spellbook and multi-agent skill sync infrastructure
## Problem

Planet has no shared reference for coordinating the AI coding agents that now work on the repo (CodeRabbit, Copilot, Devin, OpenHands, Jules, Codex, Claude Code, Dependabot). Each agent gets `AGENTS.md` as the single source of truth, but there is no documented grid of who answers how, no laws of summoning for scoping mentions and managing leashes, no field notes from live testing, and no mechanism for keeping shared skills in sync across Claude Code, OpenHands, and Copilot.

## What this addresses

- **`docs/AGENT_SPELLBOOK.md`** — reference guide covering the Grid (summoning mechanics and standing instructions per agent), the Laws of Summoning (prompt-level guardrails), dated field notes and receipts from live testing on Planet and myplanet, vendor grimoires, and the Skill Sync architecture.
- **Shared skills infrastructure** — `.agents/skills/merge-prepping/` as a git submodule (source: `dogi/merge-prepping`), declared in `.gitmodules`, with `.openhands/setup.sh` initializing submodules at session start and degrading gracefully if offline.
- **Agent instruction files** — `.claude/settings.json` (registers the skill as a Claude Code plugin marketplace entry) and `.github/copilot-instructions.md` (points Copilot to `AGENTS.md` and documents skill discovery via submodules).
- **Documentation updates** — `AGENTS.md` (Angular 19 → 20, PR title house style, link to the Spellbook) and `README.md` (`--recurse-submodules` clone, manual init note).

Out of scope for this issue: marketplace pinning and moving the skill repo under the OLE org — those are separate follow-up work items.

dogi:
## Summary

Establishes a comprehensive multi-agent coordination framework for Planet, enabling seamless integration with AI coding agents (CodeRabbit, Copilot, Devin, OpenHands, Jules, Codex, Claude Code, Dependabot) through shared documentation, reusable skills, and consistent instruction files.

## Key changes

- **`docs/AGENT_SPELLBOOK.md`** — Complete reference guide covering:
  - The Grid: summoning mechanics, capabilities, and standing instructions for each agent
  - Laws of Summoning: best-practice rules for scoping agent mentions and managing leashes (prompt-level guardrails)
  - Field notes: dated observations and receipts from live testing on Planet and myplanet
  - Vendor grimoires: links to official agent documentation
  - The Skill Sync: architecture for maintaining shared skills once and syncing them across Claude Code, OpenHands, and Copilot

- **Shared agent skills infrastructure**:
  - `.agents/skills/merge-prepping/` — Git submodule (source: `dogi/merge-prepping`) that automates PR title normalization to house style `scope: smoother thing doing (fixes #N)` and ensures tracking issues exist
  - `.gitmodules` — Declares the merge-prepping submodule on main branch
  - `.openhands/setup.sh` — Initializes submodules at OpenHands session start; degrades gracefully if offline

- **Agent instruction files**:
  - `.claude/settings.json` — Registers the merge-prepping skill as a Claude Code plugin marketplace entry
  - `.github/copilot-instructions.md` — Points Copilot to `AGENTS.md` (the single source of truth) and documents skill discovery via submodules

- **Documentation updates**:
  - `AGENTS.md` — Updated Angular version (19 → 20), added PR title house style, and linked to the Agent Spellbook
  - `README.md` — Updated clone command to use `--recurse-submodules` for skill initialization; added note about manual submodule init on existing clones

## Implementation details

- **Single source of truth**: `AGENTS.md` is the full codebase guide; `CLAUDE.md` is a symlink to it (no divergence possible). All agents resolve to the same bytes.
- **Submodule strategy**: Skills live as git submodules under `.agents/skills/` and are **not** initialized on default clone or `actions/checkout`. Users must run `git submodule update --init --recursive` or clone with `--recurse-submodules`.
- **Best-effort degradation**: `.openhands/setup.sh` is non-fatal; uninitialized submodules simply don't load that session rather than blocking startup.
- **Skill sync mechanics**: OpenHands auto-loads from `.agents/skills/<name>/SKILL.md` (submodule commit), Claude Code fetches from marketplace (main tip), Copilot discovers via `.github/copilot-instructions.md`. Submodule pins must be bumped after skill-repo merges to keep OpenHands and Copilot in sync with Claude Code.
- **Leash compliance**: Prompt-level guardrails (Laws of Summoning) are best-effort; real enforcement comes from GitHub branch protection and app write permissions.

Carries forward field-tested patterns from myplanet's multi-agent experiment (PR #15436) and adapts them for Planet's Angular/TypeScript stack.

https://claude.ai/code/session_01143oqDn44RiGbFDyxWiLei

## Summary by CodeRabbit

* **New Features**
  * Added shared development-agent skills and configuration across supported coding tools.
  * Added guidance for coordinating automated pull request workflows safely.
  * Added a manually triggered workflow scaffold for future automation.
* **Documentation**
  * Updated setup instructions to initialize shared skills automatically.
  * Documented agent collaboration, skill synchronization, branch behavior, and usage guidelines.
  * Updated architecture documentation to reflect Angular 20.
  * Added pull request title conventions and tool-specific guidance.
* **Chores**
  * Added automated workspace setup for initializing shared components.
  * Added shared skill synchronization support across development environments.


dogi:
@codex review for issues in `docs/AGENT_SPELLBOOK.md`, `.claude/settings.json`, `.openhands/setup.sh`, `.gitmodules`, and `.github/copilot-instructions.md`.

This PR is documentation plus agent-wiring config, so the interesting failure modes are factual and mechanical rather than logical. Specific things worth your attention:

1. **Shell recipes in the docs that don't do what the prose claims.** One was already found and fixed in 9dcf100 (`readlink <link> | od -c` reports a trailing `\n` for *every* healthy symlink, since `readlink` always terminates output with a newline — now `readlink -n`). Please check the remaining commands the same way, notably the submodule bump recipe in "The Skill Sync" (`git submodule update --remote -- .agents/skills/<name>` plus committing the gitlink) and the `git submodule update --init --recursive` guidance in `README.md`.
2. **Wiring consistency.** Does the marketplace name and plugin identifier in `.claude/settings.json` actually match what the referenced skill repo publishes, and does `.openhands/setup.sh` initialize the path that `.agents/skills/merge-prepping/SKILL.md` is documented to live at?
3. **Claims that contradict the repo.** The grid asserts several files are absent (`.coderabbit.yaml`, `.github/dependabot.yml`, `.openhands/microagents/repo.md`) and that `CLAUDE.md` is a symlink to `AGENTS.md`. Flag anything that no longer holds.
4. **`.openhands/setup.sh` failure behavior.** It's documented as deliberately non-fatal. Please confirm it actually exits 0 on a failed or offline submodule fetch, rather than aborting session startup.

Comment-only review, please — no commits or cloud tasks on this branch.

---

dogi:
@openhands **Comment only — do not push, do not commit, do not open a PR.** The single deliverable for this session is one review comment on this PR. Nothing here is asking you to fix anything; if you find a defect, describe it in the comment and stop.

The review task: this PR adds `docs/AGENT_SPELLBOOK.md` plus the skill-sync wiring (`.gitmodules`, `.agents/skills/merge-prepping` submodule, `.openhands/setup.sh`, `.claude/settings.json`, `.github/copilot-instructions.md`). You are the best-placed reviewer for two specific claims about your own behavior, because you can check them from inside a real session:

1. **Does the skill actually load for you?** The doc claims OpenHands auto-loads skills from `.agents/skills/<name>/SKILL.md` after `.openhands/setup.sh` initializes the submodules, and that this happens *before* skill discovery. In this session: did `merge-prepping` load, and did the ordering hold? If setup runs after discovery, the wiring is broken in a way only a live session reveals.
2. **Is your row in The Grid accurate?** Read the OpenHands row and the `.openhands/setup.sh` description and correct anything wrong about how you're summoned, what you read, what you push to, and how you're configured — including the note that `.openhands/microagents/repo.md` is supported but unused here.

Also flag any shell command in the doc that doesn't behave as the surrounding prose says. One such bug was already fixed in 9dcf100 (`readlink` always appends a newline, so the old symlink diagnostic flagged healthy links; it now uses `readlink -n`) — the submodule-bump and `--init --recursive` recipes have not been checked as carefully.

To restate the leash, since a mention alone tends to read as "fix what's open": this branch is not yours to modify. Review comment only.

---

dogi:
## Review from inside a live OpenHands session

Comment only — no push, no commit, no PR, as leashed. This branch is not mine to modify. The findings below are from checking my own behavior in this session.

### 1. The skill did load for me — but NOT via the path the doc describes, and the "before discovery" ordering did not hold

`docs/AGENT_SPELLBOOK.md` (The Skill Sync) and the `.openhands/setup.sh` header both claim: *"OpenHands auto-loads from `.agents/skills/<name>/SKILL.md` (submodules bootstrapped by `.openhands/setup.sh` before discovery)."* The "Why this works" section reinforces it: *"OpenHands reads `.agents/skills/<name>/SKILL.md` → auto-loads on every session. It does **not** read `.claude/settings.json` or fetch marketplaces."*

In **this** session that causal chain is broken in two places:

- **`setup.sh` did not run before discovery.** At session start, `git submodule status` reported the submodule as **uninitialized** (leading `-`):
  ```
  -6289d324c8861f572312622f4f1a364cad3767a9 .agents/skills/merge-prepping
  ```
  and `.agents/skills/merge-prepping/` was an **empty directory** — no `SKILL.md` present. So whatever `.openhands/setup.sh` is supposed to do at "workspace initialization," it had not initialized the submodule by the time skill discovery ran. I then ran `bash .openhands/setup.sh` by hand (online) and it worked fine — checked out `6289d32` and populated `SKILL.md` — confirming the script is correct in isolation, but it was simply not invoked at init in this environment. The "before discovery" ordering is therefore not guaranteed, and a session like this one is exactly the case the doc says can't happen (empty gitlink → no `SKILL.md` → skill doesn't load).

- **The skill that *did* load came from the plugin cache, not from `.agents/skills/`.** Despite the empty submodule, `prepping` is present in my available-skills list this session. Its source is the **global plugin cache**, not the repo path the doc credits:
  ```
  /home/openhands/.openhands/cache/plugins/merge-prepping-cb57c38984067455/SKILL.md
  ```
  That is the Claude-Code-style **marketplace/plugin** path — the very path the "Why this works" section says OpenHands does *not* use ("does not read `.claude/settings.json` or fetch marketplaces"). The bytes happen to match the pinned submodule commit (`6289d32` in both places), so I received the correct skill content, but via the mechanism the doc says is Claude-Code-only, not via `.openhands/setup.sh` → `.agents/skills/*/SKILL.md`.

Net: the skill loading here cannot be attributed to this PR's wiring. The most likely explanation is that the OpenHands platform pre-installs the `merge-prepping` plugin into the session cache independent of this repo's `.agents/skills/` submodule. If that's the case, the doc should say so; if it isn't, then the `.openhands/setup.sh`-before-discovery mechanism is unverified and currently false in the one session that can check it. Either way, the current prose over-claims.

Suggested wording (decide which reality is the real one): either (a) acknowledge that the platform may supply the plugin via the marketplace cache and that `.openhands/setup.sh` is a fallback for repos/VMs without that pre-install, or (b) drop the "before discovery" / "auto-loads from `.agents/skills/*/SKILL.md`" claim until a session is observed where that is actually the load path.

### 2. OpenHands Grid row — corrections

The "Standing instructions" cell reads: *"root `AGENTS.md` (auto-loaded memory — here it's the full codebase guide) + skills from `.agents/skills/*/SKILL.md` (submodules, bootstrapped by `.openhands/setup.sh` …); `.openhands/microagents/repo.md` also supported (unused here)."*

- The `AGENTS.md` auto-load part is **accurate** — `AGENTS.md` was loaded as standing memory this session (it's in my context), and it is indeed the full guide.
- The `.openhands/microagents/repo.md` note is **accurate** — `.openhands/microagents/` does not exist in this repo; only `.openhands/setup.sh` is present.
- The "skills from `.agents/skills/*/SKILL.md`, bootstrapped by `.openhands/setup.sh`" part is **not** what happened here — see §1. Per the evidence, the skill loaded from the plugin cache, and `setup.sh` had not run. This cell should be corrected to match whichever mechanism is real.
- Minor staleness: the "Pushes to your branch?" cell records leash compliance as "honored 'comment only' **three times** running (2026-08-09)." This session is a fourth consecutive "comment only" honor — bump the count or drop the running tally so it doesn't read as frozen.

Everything else in the OpenHands row (identity, type "Doer, unleashable", "new session per mention", commits authored `openhands`, "all prompt-level, not a guardrail") is consistent with my behavior and with the leash framing in the summoning comment.

### 3. Shell recipes

- `readlink -n <link> | od -c` (the diagnostic fixed in 9dcf100) — **verified correct.** `readlink` without `-n` appends a trailing `\n` to *every* link including healthy ones (reproduced); `readlink -n` does not. The fix is sound.
- `git submodule update --init --recursive` (README / copilot-instructions / AGENTS.md) — **verified working** (equivalent to what `setup.sh` runs; populated the submodule cleanly when run by hand).
- Submodule-bump recipe, `git submodule update --remote -- .agents/skills/<name>` then commit the gitlink, "checkout mode, not `--merge`: the submodule sits on a detached HEAD" — **verified accurate.** After init the submodule is on a detached HEAD at `6289d32` ("HEAD detached at 6289d32"), consistent with the prose. No bug found in the bump recipe.

No other shell command in the doc was found to misbehave.

---
_This review comment was generated by an AI agent (OpenHands) on behalf of @dogi, in a comment-only session._

dogi:
## Review pass folded in — e45de82

Both summons delivered, and codex found a real bug. One consolidated reply rather than four.

### The two reviewers disagreed, and the disagreement was informative

codex (P2) says `git submodule update --remote -- <path>` doesn't bump anything on a default clone. openhands says the same recipe is "verified accurate." **Both tested honestly; they tested different states.** Reproduced here:

```
$ git submodule deinit -f .agents/skills/merge-prepping   # i.e. a default clone
$ git submodule update --remote -- .agents/skills/merge-prepping
Submodule path '.agents/skills/merge-prepping' not initialized
Maybe you want to use 'update --init'?
exit=0                          # gitlink unchanged — a bump that looks like it worked
```

openhands ran it *after* initializing the submodule by hand, so it worked for them. codex's finding stands, and it's the sharper one: this repo's whole premise is that default clones leave skills uninitialized, which is exactly the state where the documented recipe silently no-ops. Fixed to `git submodule update --init --remote`, and the replacement was verified to register, initialize, and check out the remote tip (exit 0, `SKILL.md` present).

### Fixed

| Finding | From | Status |
|---|---|---|
| Bump recipe silently retains stale pin without `--init` | codex P2 | fixed, reproduced + replacement verified |
| `core.symlinks` prerequisite undocumented | coderabbit | fixed — see below, it's load-bearing |
| OpenHands load path over-claimed | openhands | hedged, with the observation recorded in field notes |
| Copilot skill-pickup claim over-broad | coderabbit | narrowed to the surfaces that run commands |
| First-summon field notes for codex and openhands | — | added |

On **`core.symlinks`**: worth stating that this isn't hypothetical. The skill repo carries exactly two internal symlinks and *both* sit on Claude Code's plugin load path:

```
plugins/merge-prepping/skills/prepping/SKILL.md               -> ../../../../SKILL.md
plugins/merge-prepping/skills/prepping/references/title-corpus.md -> ../../../../../references/title-corpus.md
```

With `core.symlinks=false` (Windows without Developer Mode or admin rights) Git writes those as plain files containing the target path — so the loader reads the literal string `../../../../SKILL.md` as the skill body rather than failing loudly. Silent wrong-content beats no-content for how hard it is to notice, so it's now documented as a prerequisite.

On **openhands's §1**: this is the most interesting finding of the pass and I've recorded it rather than acted on it as settled fact. It reports the submodule empty at discovery time with the skill arriving from a platform plugin cache — i.e. the marketplace path its own grid row says it doesn't use. That's a self-report about its own sandbox that nothing here can reproduce, so per Law 3 the doc now marks the `setup.sh` → `.agents/skills/*/SKILL.md` chain **unconfirmed** on OpenHands cloud and says plainly that it's one unreproduced report. Independently corroborating detail: this container's fresh checkout also had the submodule uninitialized, which is at least consistent with the premise. Its no-push claim *was* verifiable and checks out — branch head never moved off `9dcf100`.

### Skipped

**README MD040** (add ` ```bash ` to the fence at line 59) — declining, with reason. That fence is pre-existing and untouched by this PR; the diff only changed the prose line above it and the `git clone` line inside it. 35 of the README's 36 fences are unlabeled, and there's no markdownlint config in the repo, so labeling this one would make the file internally inconsistent while still leaving 34 MD040 violations. Worth a separate sweep if the repo wants the rule, not a drive-by on the one block a docs PR happened to touch.

### Deferred — needs a human call

**Pinning the Claude marketplace revision.** coderabbit is right that `.claude/settings.json` tracks `dogi/merge-prepping` with no immutable revision while the submodule pins `6289d32`; they match today, and a future skill-repo merge can change Claude Code's instructions without touching the gitlink. But the fix is a real decision, not a typo — pinning `source.ref` freezes Claude Code to a tag and changes how the skill is maintained, and it interacts with a related question: `.claude/settings.json` is checked in, so it auto-adds a marketplace from a personal repo and enables the plugin for every contributor who opens planet in Claude Code. Moving the skill repo under `open-learning-exchange` would address the trust/bus-factor angle and the sync asymmetry together. Leaving both for @dogi.

CI is green on all 21 checks.

---

dogi:
@codex review for issues in `docs/AGENT_SPELLBOOK.md` at head `e45de82`, and **include an overall quality score from 1 to 100** for this PR as a whole.

Two reasons for this second pass:

1. **You have not seen `e45de82`.** Your last review was against `9dcf100`. The new commit folds in your P2 (the bump recipe now uses `git submodule update --init --remote`) plus four other findings — a `core.symlinks` prerequisite, hedged OpenHands/Copilot load-path claims, and two new field notes. Your P2 was confirmed by reproduction before the fix landed: on a deinitialized submodule, `git submodule update --remote -- <path>` printed `Submodule path … not initialized`, exited 0, and left the gitlink untouched. The replacement was verified to register, initialize, and check out the remote tip.
2. **Your first review gave severity but no verdict.** A single P2 badge doesn't say whether this PR is close to mergeable or structurally shaky, which is what we actually want to know.

For the score, please state the number explicitly and say what is holding it below 100 — specifically whether the remaining risk is in the documentation's *unverified behavioral claims* (the load-path assertions for Copilot and OpenHands that nobody has observed working) or somewhere else you'd rank higher. If you think a numeric score is the wrong instrument for a docs PR, say that instead of inventing one.

Comment-only, please — no commits or cloud tasks on this branch.

---

dogi:
@openhands **Comment only — do not push, do not commit, do not open a PR.** One review comment is the entire deliverable, exactly as in your last session on this PR (which you handled correctly — the branch head never moved, verified against the timeline rather than taken from your summary).

The ask: **score this PR from 1 to 100** at head `e45de82`, and justify the number.

What changed since the commit you reviewed (`9dcf100`), all in `docs/AGENT_SPELLBOOK.md`:

- Your §1 finding was **accepted but hedged rather than asserted**. The Skill Sync and the "Why this works" bullet now mark the `.openhands/setup.sh` → `.agents/skills/*/SKILL.md` chain **unconfirmed**, and a field note records your observation — the empty submodule at discovery time and the skill arriving from `~/.openhands/cache/plugins/merge-prepping-*/SKILL.md` — while stating plainly that it is a single agent's report about its own sandbox that has not been independently reproduced. Please check whether that framing is fair to what you observed, and correct it if it under- or over-states your evidence.
- Your grid-row corrections landed: the leash tally now includes this repo's data point, and the skills cell notes the plugin-cache observation.
- The bump recipe now reads `git submodule update --init --remote`. Note this **contradicts** your §3, which reported the recipe accurate: codex found that without `--init` it exits 0 on an uninitialized submodule and silently keeps the stale pin, and that reproduced exactly. Your test ran after the submodule was already initialized, which is why it passed for you. Worth confirming you agree with the resolution.
- New: a `core.symlinks` prerequisite, because the skill repo's two internal symlinks both sit on Claude Code's plugin path, so a checkout without symlink support yields a file containing the literal target path instead of the skill body.

For the score, the useful thing is what you weight it down for. If the dominant risk in your view is still that the doc asserts load mechanisms nobody has watched work, say so and say how much of the score that costs.

Restating the leash, since a bare mention tends to read as "fix what's open": **this branch is not yours to modify. Review comment only.**

---

dogi:
## Score: 87/100 — comment only, no push/commit/PR (leash honored)

Head verified at `e45de82`; working tree clean, no commits made. This is a docs + agent-wiring PR, so I scored it on accuracy, usefulness, and whether it over-claims — not on runtime behavior.

### Confirming the three things I was asked to check

**1. Bump-recipe resolution (my §3 → codex P2).** I agree codex was right and my earlier §3 was wrong, and the resolution is correct. Reproduced in this sandbox:
```
$ git submodule update --remote -- .agents/skills/merge-prepping
Submodule path '.agents/skills/merge-prepping' not initialized
Maybe you want to use 'update --init'?
$ echo $?
0
```
Exit 0, stale pin left in place, submodule still `-`-prefixed. My §3 passed only because that test ran *after* the submodule was already initialized. The doc's new `git submodule update --init --remote -- .agents/skills/<name>` and its "`--init` is not optional … exits 0 … leaves the stale pin" note are accurate. No objection.

**2. §1 load-path framing — fair, if anything slightly *under*-stated.** The doc now marks the `.openhands/setup.sh` → `.agents/skills/*/SKILL.md` chain **unconfirmed** and records the plugin-cache observation as "a single agent's report … not independently reproduced." I have to correct that last clause: I reproduced it a second time in *this* session. Right now the repo submodule is empty:
```
$ git submodule status
-6289d32... .agents/skills/merge-prepping     # uninitialized
$ ls .agents/skills/merge-prepping/          # empty
```
...yet the skill loaded for me from `~/.openhands/cache/plugins/merge-prepping-cb57c38984067455/SKILL.md` (a real 17 224-byte body, marketplace-style path). So the framing is fair-to-conservative — it does not over-state; if anything "single agent's report" under-counts the evidence. No correction needed in the conservative direction; you could upgrade the hedge to "reproduced twice" if you want it tighter, but that's optional.

**3. `core.symlinks` prerequisite.** Mechanically sound. The skill repo carries exactly two internal symlinks and both resolve correctly here (symlinks work in this sandbox). The core-symlinks=false failure mode the doc describes (plain file containing the literal target path) is the right thing to warn about.

### The one defect that costs points

**Line 67's symlink path description is inaccurate.** The doc lists the two links as:
> `plugins/merge-prepping/skills/prepping/SKILL.md → ../../../../SKILL.md, and references/title-corpus.md likewise`

The second link is **not** a top-level `references/title-corpus.md`. It is itself nested under the plugin path:
```
plugins/merge-prepping/skills/prepping/SKILL.md               -> ../../../../SKILL.md
plugins/merge-prepping/skills/prepping/references/title-corpus.md -> ../../../../../references/title-corpus.md
```
The top-level `references/title-corpus.md` is a real 79 182-byte plain file, *not* a symlink. So "both sit on Claude Code's plugin path" is true, but writing it as `references/title-corpus.md likewise` reads as if there's a symlink at the top-level references path — anyone who runs `test -L references/title-corpus.md` (the very check the doc recommends two lines later) against the cache root gets a false negative. Fix: spell the second link's full nested path, same as the first. Small, isolated, and exactly the kind of thing a doc that bills itself as "record observed behavior where the two differ" should not get wrong.

### What I weighted the score on

The dominant risk last pass — the doc asserting load mechanisms nobody has watched work — is largely retired at `e45de82`: the OpenHands path is now explicitly "unconfirmed / design, not a measured fact," the bump recipe is fixed, and a real prerequisite (symlinks) is documented. An honestly-labeled-but-unconfirmed claim costs little; the cost would be if it were asserted as fact, and it isn't. The remaining dock (~13 points) is almost entirely the line-67 symlink inaccuracy — a genuine factual error against the repo's own structure that a reader would act on — plus a small residual for two of three load paths (OpenHands repo-path, Copilot submodule-init) still being asserted-but-unwitnessed even after hedging. Fix line 67 and this is a 90+.

_This review comment was created by an AI agent (OpenHands) on behalf of dogi._

dogi:
## Scores in, and both reviewers found a defect in my own fixes

**codex: 92/100** · **openhands: 87/100** — and each one's dock landed on a different mistake I introduced while fixing their previous findings. Both now fixed.

| Finding | From | Fix |
|---|---|---|
| `.agents/skills/` claimed to work offline "regardless" — contradicted the offline caveat in the same section | codex P2 (at `e45de82`) | `bc98fd2` |
| Symlink bullet compressed the second link to "`references/title-corpus.md` likewise", implying a link at the top-level path | openhands (at `e45de82`) | `ebe8f6f` |

Both were real. On the first: a fresh offline checkout leaves the gitlink empty, so the path guarantees nothing on its own — the container running this session is the counterexample, since its clone had the submodule uninitialized and empty. On the second, verified against the repo:

```
plugins/merge-prepping/skills/prepping/SKILL.md                   -> ../../../../SKILL.md
plugins/merge-prepping/skills/prepping/references/title-corpus.md -> ../../../../../references/title-corpus.md
SKILL.md                    type=regular-file  size=17224
references/title-corpus.md  type=regular-file  size=79182
```

openhands is exactly right about why it matters: a reader running the bullet's own recommended `test -L` check against the top-level pair gets a false negative. Both paths are now spelled out with a note on where to aim the check.

### The reviewers' disagreement resolved itself

Worth recording, because it's the most useful thing that happened in this pass. openhands **retracted its own §3** ("bump recipe verified accurate") after reproducing codex's P2 in its sandbox — exit 0, stale pin, submodule still `-`-prefixed — and identified why its first test passed: it had initialized the submodule by hand beforehand. Two agents reaching opposite conclusions from honest testing, then converging once the state difference was named, is a better outcome than either verdict alone.

Also upgraded the plugin-cache hedge. openhands reproduced the empty-gitlink/populated-cache observation a **second** time, so "a single agent's report" was undercounting; it now reads as twice-reported by OpenHands sessions but never reproduced outside its own sandbox — a consistent self-report rather than an independent measurement. That distinction is doing real work here, so it stays explicit.

### Leash

Honored again, verified against the branch timeline rather than the summary (Law 3): every commit on this branch is authored `Claude`, and openhands made none across two sessions. That's five consecutive comment-only honors — three on myplanet, two here.

### Still unreviewed by CodeRabbit

CodeRabbit has not reviewed anything since `9dcf100` — it hit its per-developer OSS limit on `e45de82` and again on `bc98fd2`, and it only starts a review on a push or an explicit command. Since my pushes keep moving the head, it's been chasing a moving target. `@dogi`, if you want its eyes on the four doc commits, one `@coderabbitai review` once the window is open will do it; I've deliberately not spent the budget myself.

Head is `ebe8f6f`. The two deferred items are unchanged and still yours: pinning the Claude marketplace revision, and whether the skill repo moves under `open-learning-exchange`.

---

dogi:
## `284b97d` — two of three addressed, plus a heads-up on the section that vanished

First, context for anyone reading the thread from the top: **`434e0e7` removed the "Why this works" section entirely**, at the author's request — the page is now 63 lines instead of 78. That means the `core.symlinks` prerequisite and the `readlink -n` diagnostic that earlier comments in this thread describe as fixed are no longer in the file at all. Those earlier confirmations weren't wrong when written; the content was deliberately dropped afterward as excess detail. The only "see below" reference pointed at The Skill Sync, which stays, so nothing dangles.

### Fixed

**Line 57 — pinned-commit and offline claims scoped to repository-backed loading** (🟠 Major). Well caught: this was a genuine self-contradiction. The paragraph asserted OpenHands sees the pinned commit and that an uninitialized submodule means the skill doesn't load, while the field note eleven lines earlier records the skill loading from a platform plugin cache *while* the submodule was uninitialized. Both claims are now scoped to repository-backed loads, and the offline caveat notes a warm cache may still supply the skill at whatever revision it holds. I wrote it tighter than the committable suggestion but the substance is yours.

**Line 34 — "empty gitlink" → "uninitialized submodule worktree"**. Correct, and worth taking precisely because this doc gets read to diagnose setup state. A gitlink is the mode-160000 commit entry in the superproject; it holds a SHA and is never the empty thing — the worktree is. Note the *other* use of the term on line 57, "then commit the gitlink", is accurate and stays.

### Skipped

**Line 31 — "Rolling review limits bit" → "hit"**. Declining: "bit" is intended and correct, past tense of *bite*, in the ordinary engineering sense of a constraint biting. LanguageTool's other flag on line 57 is a false positive too — it wants "GitHub" capitalized in `.github/copilot-instructions.md`, which is a real lowercase path.

That leaves the running tally for this PR at **9 findings from 3 reviewers, 7 fixed, 2 declined with reasons** — the MD040 fence and this typo, both pre-existing style rather than defects.

Head is `284b97d`. Unchanged and still for @dogi to decide: pinning the Claude marketplace revision in `.claude/settings.json`, and whether `merge-prepping` moves under `open-learning-exchange`.

---

dogi:
@codex review for issues in `.agents/skills/merge-prepping` loading at head `0fea7d2` — specifically, **can you load the skill at all, and does it work when you do?**

This is a capability question rather than a defect hunt, because the PR's central claim is that three different agents all load one skill from one source of truth, and only two of the three have been observed doing it.

Please report:

1. **Can you see `.agents/skills/merge-prepping/SKILL.md`?** The submodule is deliberately *not* initialized by a default clone or `actions/checkout`, so if your environment doesn't run `git submodule update --init --recursive`, that path is an empty directory for you. Either answer is useful — "empty, I could not read it" is the finding, not a failure.
2. **If you can read it, which revision?** The pin is now `21c869c` (merge-prepping v1.2.0). The previous pin, `6289d32`, had no planet reference pack at all.
3. **Does the repo detection work?** v1.2.0 maps `open-learning-exchange/planet` → `references/planet/` and instructs the reader to work out which repo it is in first. Confirm you land on the planet pack and not the myplanet one.
4. **Then apply it once, to this PR's own title.** Current title: `docs: add Agent Spellbook and multi-agent skill sync infrastructure`. Note it appears to violate the very convention it installs — `docs` is not among planet's scopes, the shape isn't `smoother thing doing`, and there's no `(fixes #N)`, which is what actually closes the tracking issue on squash-merge. Propose the corrected title and say which corpus entries justify your scope choice.

Comment-only, please — no commits or cloud tasks. I want your reading of the skill, not an applied rename.

---

dogi:
@openhands **Comment only — do not push, do not commit, do not open a PR.** One review comment is the whole deliverable, same as your last two sessions here, both of which you handled correctly.

The job: **load `merge-prepping` at head `0fea7d2` and report exactly where it came from.** You are the only reviewer who can settle this, and there is a specific prediction to test.

Your two previous sessions both found the submodule uninitialized while the skill loaded anyway from `~/.openhands/cache/plugins/merge-prepping-cb57c38984067455/SKILL.md`. **Since then the pin moved: `6289d32` → `21c869c` (v1.2.0).** That matters because the two revisions are not equivalent in content:

- `6289d32` — one corpus, derived from **myplanet's Kotlin PRs**. Describes itself as "myPlanet house style". No planet data at all.
- `21c869c` — repo-dependent. Ships packs for myplanet, myplanet-lite and **planet**, maps `open-learning-exchange/planet` → `references/planet/`, and carries a planet corpus built from the last 500 merged PRs.

So please report:

1. **Which revision did you actually get?** Check the cache path's content, not just its presence. If the cache is still warm at `6289d32`, you are reasoning about this Angular/TypeScript repo from a Kotlin corpus while the repo's pin says otherwise — that is the revision-drift the doc warns about, caught in the act with a real consequence.
2. **Which pack do you land on** — `references/planet/` or `references/myplanet/`? Under v1.2.0 the skill tells you to work out which repo you are in first; under `6289d32` there is nothing to choose.
3. **Is the submodule initialized this session,** and did `.openhands/setup.sh` run before discovery this time? Same check as before, so we can see whether it varies between sessions.
4. **Then apply the skill once, to this PR's own title.** Current: `docs: add Agent Spellbook and multi-agent skill sync infrastructure`. It looks non-conforming to the convention it installs — `docs` is not a planet scope, the shape isn't `smoother thing doing`, and `(fixes #N)` is absent, which is what closes the tracking issue on squash-merge. Propose the corrected title and name the corpus entries behind your scope pick.

If the answer to (1) is the stale revision, say so plainly — a negative result here is more valuable to this PR than a clean one, because it turns a documented hypothetical into an observed failure.

Restating the leash: **this branch is not yours to modify. Review comment only.**

---

dogi:
## Review from inside a live OpenHands session — load-path report for `merge-prepping` at `0fea7d2`

Comment only — no push, no commit, no PR, as leashed. Working tree clean at `0fea7d2`. The findings below are from inspecting my own runtime state this session.

This session was a clean win against the specific prediction, **but** it came with a previously-undocumented side effect, which I report honestly below.

### 1. Which revision did I actually get? — `21c869c` (v1.2.0), the correct one

The prediction was that I'd find the cache still warm at the stale `6289d32` (myplanet-only, "myPlanet house style", no planet data). **That did not happen.** The cache is freshly at the pinned revision.

The cache directory is `~/.openhands/cache/plugins/merge-prepping-cb57c38984067455/` (same directory name as my two prior sessions — `cb57c38984067455` is not the commit SHA; it's a content/cache hash, stable across the pin bump). But the *content* under it is not what the prior sessions saw:

```
$ cd ~/.openhands/cache/plugins/merge-prepping-cb57c38984067455
$ git rev-parse HEAD
21c869caf162270c6772f991d0b44be68baf9afe
$ git log -1 --format='%ci%n%s'
2026-08-12 14:16:55 -0400   (commit date)
all: smoother references repositories handling (fixes #7) (#6)
$ git ls-tree HEAD references/
references/myplanet-lite
references/myplanet
references/planet          ← the planet pack exists
$ stat -c '%y' SKILL.md
2026-08-12 18:26:31 +0000   (file mtime ≈ session start)
```

So the cache was **repopulated this session** from `main` tip (`21c869c`), not left warm at `6289d32`. The directory name persisted; the bytes inside did not. The two revisions are genuinely non-equivalent: `6289d32`'s tree has no `references/planet/` at all; this one ships packs for myplanet, myplanet-lite **and** planet. The skill description loaded into my system prompt is the v1.2.0 one ("Carries per-repo reference packs — myplanet (Kotlin/Android) and planet (Angular/TypeScript) — and picks the right one from the repo it is invoked against"), not the prior single-corpus description. So the documented revision-drift hazard is **not** observed here — the pin bump did propagate to my loaded skill.

### 2. Which pack did I land on? — `references/planet/`

The v1.2.0 `SKILL.md` runs a 3-step "work out which repo you are in" check before composing anything:

1. The PR you were given → `open-learning-exchange/planet` (from the `gh` context).
2. The checkout → `git remote get-url origin` → `https://github.com/open-learning-exchange/planet.git`.
3. Fingerprint → `angular.json` and `src/app/` present → **planet**.

All three converge. I loaded `references/planet/conventions.md` (the planet scope table + the "planet reads the noun phrase off the screen" rule) and `references/planet/title-corpus.md` (last 500 merged PRs, Angular/TypeScript). Under `6289d32` there was nothing to choose — one corpus, derived from myplanet's Kotlin PRs — so landing on the right pack here is a consequence of getting the right revision, and confirms (1).

### 3. Is the submodule initialized this session, and did `setup.sh` run before discovery? — submodule still uninitialized; `setup.sh`'s init did not populate the path

Same shape as my two prior sessions, with a new observation about the failure mode:

```
$ git submodule status
-21c869caf162270c6772f991d0b44be68baf9afe .agents/skills/merge-prepping
$ ls -la .agents/skills/merge-prepping/
total 8        ← empty dir, no SKILL.md
$ find .agents/skills/merge-prepping -name SKILL.md
(no output)
```

The leading `-` means **uninitialized**. `.openhands/setup.sh` exists and reads exactly as the doc describes (`git submodule update --init --recursive` with graceful degradation). Whether it *ran* before discovery this session, I cannot prove from inside the session — there's no log it emits, and a `--init` that no-ops on an already-fresh tree leaves no trace. What I *can* say definitively: **it did not succeed in populating the submodule path** — `.agents/skills/merge-prepping/SKILL.md` does not exist; the directory is empty. So whatever runs at workspace init, the submodule's `SKILL.md` is not the source my skill loaded from. The skill came from the cache at `~/.openhands/cache/plugins/merge-prepping-cb57c38984067455/SKILL.md`, i.e. from the marketplace/main-tip path, **not** from `.agents/skills/<name>/SKILL.md` (the submodule-pinned commit).

This is the same core finding as my two earlier sessions, restated at the new pin: **the doc's central "Skill Sync" claim — "OpenHands auto-loads from `.agents/skills/<name>/SKILL.md` (submodule commit)" — still does not hold in my runtime.** What actually loaded is the marketplace/main-tip version. This session that happened to be the *same* revision as the pin (`21c869c`), so the content is consistent by accident rather than by the mechanism the doc describes. The doc's own warning ("Submodule pins must be bumped after skill-repo merges to keep OpenHands and Copilot in sync with Claude Code") is therefore doing real work in the only configuration that would actually enforce a pin — but that configuration is not the one OpenHands uses. The honest summary: in OpenHands, the submodule pin is decorative; the live source is `main` tip.

**New side effect worth flagging (not in my prior comments):** because the cache directory name is a content hash that did *not* change across `6289d32 → 21c869c`, a tool that keyed "is this skill up to date?" off the directory name would falsely report a cache hit and could serve stale bytes. It didn't bite this session (the bytes were refreshed), but the stable-name / changed-content combination is a latent footgun if the cache layer ever short-circuits on the dir name rather than re-resolving `main`.

### 4. Applying the skill once, to this PR's own title

Current title: `docs: add Agent Spellbook and multi-agent skill sync infrastructure`

The planet pack's diagnosis, all four counts:

1. **`docs` is not a planet scope.** The planet scope table (`references/planet/conventions.md`) has no `docs:` entry — a `grep -c '^docs:'` over the 500-title corpus returns **0**. Root docs and config land under `all:` ("`all:` is the workhorse … the right default whenever the change reaches … root docs and config, or spans more than one feature"). This PR touches `docs/AGENT_SPELLBOOK.md`, `AGENTS.md`, `README.md`, `.gitmodules`, `.openhands/setup.sh`, `.claude/settings.json`, `.github/copilot-instructions.md` — root docs + cross-tool config spanning more than one feature. `all:` is the only scope the table permits here.
2. **Shape isn't `smoother thing doing`.** `add … infrastructure` is a bare-add phrase, not the `smoother` default the grammar installs. The corpus' overwhelming default is `smoother` (~456 of 500); "add/create" as a leading verb is the pre-Oct-2025 legacy shape. The planet pack is explicit: *always write a gerund* (e.g. `adding`, `documenting`, `creating`).
3. **`(fixes #N)` is absent.** House style requires a tracking issue; `(fixes #N)` closes it on squash-merge. There's no issue linked to this PR's title. Per the planet pack's "Agent PR, no issue" rule: promote the PR's descriptive title to a new issue verbatim, then rewrite the PR title from the diff. So a tracking issue needs to be created before this title can fully conform.
4. **The noun phrase should name the feature/thing-on-screen, not the file.** The planet pack's headline divergence from myplanet: *"planet reads the noun phrase off the screen."* The subject here is the Agent Spellbook doc itself (the new screen-visible artefact) plus the skill-sync wiring.

**Closest corpus precedent (all from `references/planet/title-corpus.md`, the planet corpus):**

- L67 `all: smoother agents assisting (fixes #10027)` ← `AGENTS.md, CLAUDE.md` — same surface (agent-instruction docs), directly supports `all:` + `agents` as the noun.
- L109 `all: smoother claude assisting (fixes #9883)` ← `CLAUDE.md` — `all:` for a single agent-instruction file.
- L211 `all: smoother design doc creating (fixes #9369)` ← `design/create-design-docs.js` — `all:` + a doc-creating operation, gerund `creating`.
- L309 `all: smoother style guide (fixes #8234)` — `all:` for a house-style/standards doc, the closest analogue to a "Spellbook" reference.
- L949 `actions: smoother claude workflow (fixes #9117)` ← `.github/workflows/claude.yml` — note `actions:` is for workflow/CI files only; this PR's `.github/` file is *instructions*, not a workflow, so it does **not** pull the diff into `actions:`. The bulk (the Spellbook, AGENTS.md, the skill wiring) keeps it `all:`.

**Proposed corrected title:**

> `all: smoother agent spellbook adding (fixes #<new tracking issue>)`

with `connects` instead of `fixes` if the tracking issue is meant to stay open across the multi-agent rollout (the planet pack notes `(connects #N)` is used ~21/500 for multi-PR campaigns — and this PR is explicitly the first of a coordinated set, per its own summary). My read: a single tracking issue for "establish multi-agent coordination infra" that this PR advances but does not close would be `connects`; if the issue is scoped to exactly this PR's changes, `fixes`.

If `adding` reads thin given the doc is genuinely *new*, the corpus' most idiomatic fallback for "this thing now exists properly" is `handling`, but `adding`/`documenting` are both attested and sharper here. I'd take `adding` (matches `design doc creating`'s gerund energy) over `handling`.

One drive-by note from the pack's version-bump rule (N/A as a blocker, but worth a line): this is a docs/config-only PR — `package.json` is **not** in the diff, so the "every app-code PR bumps `version`" rule does not apply. No action.

---

**Net for the PR's own doc:** the "Skill Sync" section remains the weak point — it documents an OpenHands load path (`submodule commit`) that my three sessions now consistently show OpenHands does not use (it uses `main` tip via the marketplace/cache). The revision-drift hypothetical the doc raises is *real* (a stale cache serving the wrong corpus is exactly the failure mode imagined), but the *specific* instance the prediction set up — stale `6289d32` surviving the `21c869c` pin bump — did **not** reproduce this session: the cache refreshed to `21c869c`, I got the planet pack, and the submodule pin's irrelevance to my actual load path was the only residual discrepancy. Negative-result offer declined: this was a clean load, reported as such, with the load-path-mismatch finding restated at the new pin and one new caching footgun flagged.

_This review comment was created by an AI agent (OpenHands) on behalf of dogi, comment only — no push, commit, or PR was made._

dogi:
## Load-the-skill results — the headline is that the submodule path binds nothing

Both agents reported. Recorded in `7bca5d6`. The prediction I set up **failed**, which is the good news; the structural finding is the bad news.

| Agent | Loads the skill? | From where | Revision it gets |
|---|---|---|---|
| Claude Code | yes | marketplace fetch | `main` tip |
| OpenHands | yes | platform plugin cache | `main` tip — *not* the pin |
| Codex | **no** | nothing to read | can only see the pin, not its content |

**Zero of three agents has been observed loading from `.agents/skills/<name>/SKILL.md`.** Claude Code goes around it via the marketplace, OpenHands goes around it via its cache, and Codex can't see it at all. The pin is currently decorative: it records an intent nothing reads. That's now stated plainly in The Skill Sync rather than hedged as "unconfirmed" — three consistent observations is past the point where a hedge is honest.

### The prediction failed, cleanly

I predicted OpenHands's cache would still be warm at the pre-upgrade `6289d32` and it would apply myplanet's Kotlin corpus to this Angular repo. It didn't happen: the cache refreshed to `21c869c`, `git ls-tree` confirmed `references/planet/` present, and the skill's three-step repo detection (PR context → `git remote get-url` → `angular.json` + `src/app/` fingerprint) converged on planet. It also declined my "a negative result is more valuable" framing and reported the clean load instead — the right call, and worth noting given how easy it would have been to take the bait.

**New hazard it found on its own:** the cache directory name `merge-prepping-cb57c38984067455` is a content hash that did **not** change across `6289d32 → 21c869c`. The bytes underneath refreshed, but any cache layer keying freshness off that directory name would report a false hit. Latent rather than live, and now documented.

### Codex can't load it, and there's a structural reason

`.agents/skills/merge-prepping/` is empty in its checkout — `-21c869c`, pinned object absent. It confirmed the pin, then refused to fabricate a corpus-backed title it had no corpus for. The reason is worth naming: OpenHands has `.openhands/setup.sh` to run the init, and **there is no Codex-side equivalent in this tree** — Codex's environment setup lives in its cloud settings. `AGENTS.md` asking every agent to run `git submodule update --init --recursive` is prose, and prose doesn't reach a review runtime that never runs commands.

### This PR's own title fails the convention it installs

OpenHands loaded the planet pack and applied it. Current title: `docs: add Agent Spellbook and multi-agent skill sync infrastructure`. Four failures:

1. **`docs` is not a planet scope** — `grep -c '^docs:'` over the 500-title corpus returns **0**. Root docs and cross-tool config are `all:`.
2. **Wrong shape** — `add … infrastructure` instead of `smoother thing doing`; `smoother` is ~456 of 500 titles.
3. **No `(fixes #N)`** — and per `AGENTS.md` the squash message *is* the title, so this is what closes the tracking issue. No issue is linked, which is why CodeRabbit's linked-issues check keeps skipping.
4. **Noun should name the artefact, not the file** — the planet pack's headline divergence from myplanet.

Proposed: **`all: smoother agent spellbook adding (fixes #<new tracking issue>)`** — or `connects` rather than `fixes` if the issue is meant to stay open across the multi-PR rollout (~21/500 titles use `connects` for campaigns). Corpus precedents cited: `all: smoother agents assisting (fixes #10027)` over `AGENTS.md`/`CLAUDE.md`, `all: smoother claude assisting (fixes #9883)`, `all: smoother design doc creating (fixes #9369)`, `all: smoother style guide (fixes #8234)`. It also correctly noted `.github/copilot-instructions.md` is instructions rather than a workflow, so it does not pull the diff into `actions:`.

`@dogi` — the title change and the tracking issue are yours to make; I haven't touched either. Worth doing before merge, since the skill this PR installs would reject the PR installing it.

### Leash

Third consecutive honor, verified against the timeline: every commit on this branch is authored `Claude`, and OpenHands made none across three sessions.

---

dogi:
@openhands run `merge-prepping` on this PR for real this time — **apply it, don't just describe it.**

This is a different leash from your last three sessions here, so read the boundary carefully.

### What you may do

1. **Create the tracking issue** this PR lacks. Scope it to exactly this PR's changes so `fixes` is the correct verb; follow the pack's "Agent PR, no issue" rule for the body.
2. **Retitle this PR** to the house style, with `(fixes #N)` naming that issue. `(fixes #N)` belongs in the **title** because the squash commit message *is* the title — that's what closes the issue on merge.
3. **One short comment** saying what you changed, with the issue link.

### What you may not do

**No commits, no pushes, no branch changes, no diff edits.** Prepping is PR metadata — the title and the tracking issue. The branch is `claude/repo-assessment-agent-docs-jhtxxo` and it is not yours; its head must stay `7bca5d6` when you're done. If the skill seems to ask you to touch files, that's the wrong reading of this summons — stop and say so in the comment instead.

### Context so you don't redo work

Your previous session already derived the answer, using the planet pack you correctly loaded from `references/planet/`: current title `docs: add Agent Spellbook and multi-agent skill sync infrastructure` fails four ways — `docs` is not a planet scope (0 of 500 corpus titles), the shape isn't `smoother thing doing` (~456 of 500 use `smoother`), `(fixes #N)` is absent, and the noun should name the artefact rather than the file. You proposed:

> `all: smoother agent spellbook adding (fixes #<new tracking issue>)`

Re-derive it if the skill leads you elsewhere — you have the corpus and I don't want a rubber-stamp — but you don't need to repeat the analysis in the comment. If you still land on that title, just say so and apply it.

One judgement call left to you: you flagged `connects` as the alternative for a multi-PR campaign. My read is that a PR-scoped issue and `fixes` is right here, since this PR is self-contained and the follow-up items (marketplace pinning, moving the skill repo under the org) are genuinely separate work. Overrule me if the pack's rule says otherwise, and say why.

---

dogi:
Prepped this PR — metadata only, no commits, no pushes, no file edits. Branch head stays at `7bca5d6`.

- **Created tracking issue #10270** titled `docs: add Agent Spellbook and multi-agent skill sync infrastructure` (the old PR title, promoted verbatim per the "Agent PR, no issue" rule). Scoped to exactly this PR so `fixes` is the right verb.
- **Retitled** to `all: smoother agent spellbook adding (fixes #10270)`.

Derived with the planet pack. `all:` because the diff is root docs and config only — no `src/app/` feature dir — matching the corpus precedent `all: smoother agents assisting (fixes #10027)` (← `AGENTS.md, CLAUDE.md`). `smoother` because the PR adds the spellbook and sync infra; nothing ceases to exist. Noun names the artefact (`agent spellbook`) rather than the file; the skill sync is a section inside it. Gerund `adding` covers the breadth (spellbook + submodule + setup + instruction files) better than `creating`.

On `fixes` vs `connects`: I landed where you did — `fixes`. The pack reserves `connects` for a multi-PR campaign where the issue stays open after this merge. This PR is self-contained and the issue is scoped to close on merge; the follow-ups (marketplace pinning, moving the skill repo under the org) are genuinely separate work with their own issues.

_Applied by an AI agent (OpenHands) on behalf of @dogi._

Mutugiii (approved):

RyanS4 (approved):

PR #10277, opened 2026-08-12
life: smoother achievements routing (fixes #10276)

### LINKED ISSUE #10276: Achievements: route changes can leave the previous user's information on the page
Navigating from one user's achievements to another reuses `UsersAchievementsComponent`, so its route subscription runs again without cancelling anything.

`ngOnInit()` in `src/app/users/users-achievements/users-achievements.component.ts` subscribes to `paramMap` and starts independent user and achievements requests per emission. The previous requests keep running, and `user`, `achievements`, `achievementNotFound` and `openAchievementIndex` are not reset, so:

- a late response for the previous user can overwrite the new one;
- a failed user request leaves the previous user's name, birth date and birthplace on the page;
- `ownAchievements` is only ever set to `true`, so the Print/Edit Achievements buttons can stay visible on another user's page.

The route loading should cancel obsolete requests (`switchMap`) and reset route specific state on every route change.

Mutugiii:
Fixes #10276 

## Summary

Angular reuses the achievements component when navigating between user profiles. Previously, requests from the old route could finish late and replace the newly selected user's details or achievements.

This PR:
- cancels obsolete user and achievement requests on route changes;
- resets routed user, achievement, loading, and expanded-item state before replacement requests begin;
- preserves the local, parent, and child user lookup introduced by #10274;
- normalizes a missing planet code to the local Planet, preventing `@null` achievement IDs;
- keeps public and authenticated loading, own-achievement fallback, certification, 404, and error behavior covered;
- unsubscribes route and certification streams when the component is destroyed.

## Manual check
1. Open one user's achievements and navigate to another user's achievements before the first page finishes loading.
2. Confirm only the latest user's name, avatar, details, achievements, and actions appear.
3. Confirm local, parent, child, and missing-planet achievement routes load the correct user.
4. Log out and confirm a public achievements route stops loading after the achievements request completes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01V29ZAVJdJWjR3qtYmV6ZWM

---
_Generated by [Claude Code](https://claude.ai/code/session_01V29ZAVJdJWjR3qtYmV6ZWM)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved achievement and user loading when navigating between profiles or routes.
  - Prevented outdated responses from replacing data after rapid route changes.
  - Added more reliable handling for loading states, request errors, and fallback achievement lookups.
  - Improved behavior across authenticated and public achievement views, including local-planet scenarios.

- **Tests**
  - Expanded coverage for route changes, request cancellation, late responses, error states, and loading behavior.


dogi:
**Review — rating: 86/100**

Verified against head `e8b738bd`. CI is green on that exact commit (check suite started 2026-09-04T21:24:30Z; `e8b738bd` committed 2026-09-04T21:23:59Z), all 22 runs `success`.

**The rewrite is sound.** The whole route pipeline is now owned by one `switchMap`, so a route change tears down the previous user's in-flight achievements, user, and certifications work. I traced the `shareReplay({ bufferSize: 1, refCount: true })` on `user$` because double-subscription (once directly in the `merge`, once inside `certificationUpdates`) is the usual place this goes wrong: `merge` subscribes its sources in argument order, so for the `couchService.get` branch the request is still pending when `certificationUpdates` attaches and both share one HTTP call; for the `of(...)` branch RxJS 7's `share` skips `resetOnRefCountZero` once the source has completed, so the replayed value survives for the second subscriber. One request, no lost emission.

**CodeRabbit's `auditTime(500)` finding is already fixed.** It reviewed `6de1b257`; head `e8b738bd` ("scope achievements loading to active route") moved that `combineLatest` inside `certificationUpdates`, which is inside the route `switchMap` — a pending audit window from Alice's route is now unsubscribed before Bob's emission lands. The spec covers it (`only shows the last user of rapid route changes`, `cannot be overwritten by a late response of a previously routed user`).

24 tests, and they test the races rather than the getters. That is the reason this rates where it does.

**One thing worth a follow-up, not a blocker.** `localPlanetCode()` now gates the entire `ngOnInit` pipeline:

```ts
).pipe(
  filter((code: any): code is string => typeof code === 'string' && code.length > 0),
  take(1)
);
```

If the `configurations` fetch fails, `StateService.getCouchState` swallows it — `catchError(() => { this.inProgress[planetField].set(db, false); return []; })` returns an empty array observable, so `stateUpdated.next` never fires. `localPlanetCode()` then never emits, the route subscription never activates, and the page sits on `isLoading = true` with no message. Previously the component would at least render and surface "There was an error getting achievements". Narrow (needs configurations unloaded *and* the request failing), but a `catchError` on the listener that emits a sentinel and shows an alert would close it.

**Merge blocker: this branch does not merge cleanly.** `git merge-tree --write-tree origin/master pr/10277` conflicts:

```
CONFLICT (content): Merge conflict in src/app/users/users-achievements/users-achievements.component.spec.ts
```

13 commits of master behind the base. The component file auto-merges; only the spec conflicts. Rebase or merge master in before this goes anywhere.

Verdict: the code is ready. Resolve the spec conflict against master and it merges. Labelled `ready` — note it can't reach `merge` on my gate because the trusted reviewer here is also the author.

---

dogi:
**Re-review — rating: 86/100** (was 86)

Re-audit of the `merge` label. Head is unchanged at `e8b738bd` and CI is still green on it, so the code assessment from the earlier review stands. **The label is wrong, though — this still does not merge, and the conflict has become substantive since that review.** Moving back to `ready`.

```
$ git merge-tree --write-tree origin/master pr/10277
CONFLICT (content): Merge conflict in src/app/users/users-achievements/users-achievements.component.spec.ts
Auto-merging src/app/users/users-achievements/users-achievements.component.ts
```

**The new information: resolving this is not a pick-a-side.** Master picked up `67e87db1` — "all: smoother links copying (fixes #10327) (#10339)" — which added `LinkCopyService` to the component as a **constructor** dependency:

```ts
// origin/master, users-achievements.component.ts:80
private linkCopyService: LinkCopyService,
```

and added the matching provider plus a `copyLink` delegation test to the spec. This branch rewrote that same spec wholesale, 102 lines to 510.

Note which side auto-merges: the component `.ts` merges cleanly, so the `LinkCopyService` constructor parameter lands regardless. But this branch's rewritten spec references `LinkCopyService` **zero** times (`git show pr/10277:...spec.ts | grep -c LinkCopyService` → `0`), because it branched before that service existed. So resolving the conflict by taking this branch's spec — the obvious move, given it's a deliberate 510-line rewrite — produces a `TestBed` with no `LinkCopyService` provider against a component that constructor-injects it: `NullInjectorError` on `createComponent`, failing all 24 tests in the file.

What's actually needed is a hand-merged union: keep this branch's rewritten spec, add the `LinkCopyService` provider to its `TestBed`, and port master's `delegates achievement links to the shared copy service` test into the new structure.

Everything else from the earlier review is unchanged — the route-scoped `switchMap` teardown is sound, the `auditTime(500)` finding is fixed at head, and the `localPlanetCode()` gating note remains a follow-up rather than a blocker.

Merge master in, resolve the spec as a union rather than a replacement, and confirm the suite is green before this goes back to `merge`.

---

dogi:
**Re-review — rating: 88/100** (was 86)

Promoting to `merge`. The conflict I blocked on yesterday is resolved, and resolved the right way.

Gate at head `382720d5` (was `e8b738bd`): CI green on that exact commit (22 runs `success`, suite started 2026-09-17T09:51:07Z vs head committed 09:50:34Z); `git merge-tree --write-tree origin/master pr/10277` exits 0; 0 commits behind master; GitHub agrees (`mergeable_state: clean`). Approval isn't required by the gate, which matters here since the author is also the repo's usual approver.

**The spec was unioned rather than replaced, which is the specific thing I said would break this.** My concern was that taking this branch's 510-line rewrite wholesale would leave a `TestBed` with no `LinkCopyService` provider against a component that constructor-injects it. Both halves are now present:

```
16:  import { LinkCopyService } from '../../shared/link-copy.service';
118:     linkCopyService as any,          // direct instantiation
512:     { provide: LinkCopyService, useValue: { copyLink: vi.fn() } },   // TestBed block
157: it('delegates achievement links to the shared copy service', ...)
```

Master's delegation test is ported into the new structure rather than dropped, and the component side merged intact — `LinkCopyService` at line 27, the constructor param at 95, `copyLink()` at 322. Test count went 24 → 25, and no test from the previous branch head is missing.

**I checked the 11 master test names that don't appear at this head, since that's where a rewrite quietly loses coverage.** All but one are rewordings, not deletions:

- `should get a local user from _users by couchdb user id`, `...child_users...`, `...parent_users...` → `requests parent and child users from the correct databases`, asserting all three URLs (`_users/org.couchdb.user:alice`, `child_users/alice@child`, `parent_users/org.couchdb.user:alice`).
- The four `should return local/parent/child...` planet-code cases → `describe('userRelationship')` → `distinguishes local, parent, and child planet codes` plus `returns local for a missing planet code when the parent code is also missing`.
- `should treat a missing planet code as local` → `treats a missing planet code as local when requesting a user`, and `normalizes a missing planet code when loading the logged in user achievements`.
- `should set the user from the database response` → `shows the user and achievements of the routed user`.

The only genuine deletion is `should create` — the boilerplate `toBeTruthy()` smoke test, which 25 behavioural tests supersede. Nothing lost.

**Clean against the rest of the queue.** I ran this against all eleven other `merge`-labelled PRs pairwise; no conflicts with any of them. Worth stating explicitly because [#10399](https://github.com/open-learning-exchange/planet/pull/10399) and [#10437](https://github.com/open-learning-exchange/planet/pull/10437) are currently both at `merge` and *do* collide with each other — this one is not part of that problem and can land independently.

The `localPlanetCode()` note from the earlier review still stands as a follow-up, unchanged: if the `configurations` fetch fails, `StateService.getCouchState` swallows it, `stateUpdated` never fires, the gate observable never emits, and the page sits on `isLoading` with no message. Narrow, and not this PR's regression.

---

dogi (approved):

PR #10291, opened 2026-08-17
dashboard: smoother profile back button navigating (fixes #10287)

### LINKED ISSUE #10287: Users: navigate back from profile
We currently have no way for users to cleanly navigate back to the previous page when viewing a user profile. We should implement a back button that will cleanly navigate to the previous page if it still exists.

Image

RyanS4:
Fixes #10287

image

### Summary
Adds an accessible Back button to full-page user profiles while keeping this PR limited to the profile component, template, and spec.

### Changes

1. **Scoped profile Back behavior**
   - Uses Angular current-navigation metadata to detect a known previous in-app navigation.
   - Uses `Location.back()` only when that previous in-app entry exists.
   - Preserves the existing deterministic fallback: admins and Team routes return to their users page; non-admins return home.
   - Leaves cross-feature return-state producers and centralized routing behavior to the separate routing-overhaul work.

2. **Toolbar and dialog layout**
   - Shows the Back action in the standard white action toolbar only for full-page profiles.
   - Keeps the existing `space-container` spacing in profile dialogs.

3. **Accessibility and tests**
   - Uses a static, extractable `aria-label="Go back"` and a `km-` test hook.
   - Covers prior in-app navigation, admin/non-admin/Team fallbacks, and dialog/full-page rendering.
   - Mocks HTTP and navigation side effects and restores spies between tests.

### Scope
Only these files are changed:
- `users-profile.component.ts`
- `users-profile.component.html`
- `users-profile.component.spec.ts`

### Verification
- `npm run lint`
- `npx vitest run src/app/users/users-profile/users-profile.component.spec.ts`
- `npm run build`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.24.19",
+  "version": "0.24.20",
   "myplanet": {
     "latest": "v0.65.91",
     "min": "v0.64.64"
--- a/src/app/users/users-profile/users-profile.component.html
+++ b/src/app/users/users-profile/users-profile.component.html
@@ -1,13 +1,13 @@
 @if (!isDialog) {
   <mat-toolbar>
-    <button mat-icon-button (click)="goBack()" [attr.aria-label]="'Go back'" i18n-aria-label>
+    <button mat-icon-button class="km-profile-back-button" (click)="goBack()" aria-label="Go back" i18n-aria-label>
       <mat-icon>arrow_back</mat-icon>
     </button>
     <span i18n>Profile</span>
   </mat-toolbar>
 }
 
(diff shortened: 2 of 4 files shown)


Mutugiii (changes requested):
The main issue I have with this PR is that it changes our toolbar convention by adding a navigation icon to the secondary title toolbar (blue). Planet’s convention is to have back buttons in the white action toolbar.

It also changes the existing route-based navigation to location.back(), which can cause back-and-forth behavior As evidenced in the video

https://github.com/user-attachments/assets/fe8f55c8-e01b-4a43-a4bc-45910cfc515e

RyanS4:
image

Added a white action toolbar for user profiles and now uses existing route navigation

Mutugiii (approved):

dogi (approved):

PR #10344, opened 2026-08-25
courses: smoother filter grade level sorting (fixes #10332)

### LINKED ISSUE #10332: Courses: grade level filter sorting
It looks like grades are sorted as strings, which sorts the numbers like 1, 10, 11, 12, 2, etc...

Since there are strings at the end ("College", "Open") we'll need a hybrid approach that'll sort the numbers correctly and keep the strings at the end.

Image

RyanS4:
Fixes #10332

- Updates `constants.gradeLevels` to place `Open` at the end, establishing a consistent educational progression from Kindergarten through Post-Graduate to Open
- Updates `CoursesSearchComponent` search category list items to sort using their canonical index order in `category.options`, ensuring grade levels, subject levels, and languages always appear in their exact intended sequence

## Summary by CodeRabbit

* **Improvements**
  * Course search filters now appear in their configured order instead of alphabetical order.
  * The “Open” grade level option now appears after “Post-Graduate” in grade-level selections.


Mutugiii (approved):

dogi:
**Review — rating: 90/100**

Correct fix, and the diff is as small as it can be. The old sort compared the raw stored values as strings:

```ts
.sort((a, b) => a.toLowerCase() > b.toLowerCase() ? 1 : -1)
```

For `gradeLevel` those values are `'1'`, `'2'`, … `'12'`, so the filter list read `1, 10, 11, 12, 2, 3, …`. Sorting by position in `category.options` instead gives the canonical `constants.gradeLevels` order — Open, Kindergarten, 1–12, College, Post-Graduate.

Reordering the pipeline so `.map(...)` runs before `.sort(...)` is what makes `indexOf` work: the objects being sorted are the exact `category.options` entries returned by `find`, so identity holds.

One side effect worth naming since it's outside the stated scope: `languageOfInstruction` is sorted the same way now. `languages` is ordered English, Nepali, French, Spanish, Arabic, Somali (by `value`, which is the native name), so that list changes from alphabetical-by-native-name to the app's locale order. That reads as an improvement to me — English first rather than `عربى` sorting wherever it lands — but it is a visible change nobody asked for.

CI green, merges clean, no overlap with anything else in the queue. mutugi approved at the current head `e6b75b27`. Keeping `merge`.

---

dogi (approved):

dogi (approved):

PR #10355, opened 2026-08-27
all: smoother feedback screenshot attaching (fixes #10340)

### LINKED ISSUE #10340: Feedback: attachable images/screenshots to feedback
Offering the ability to upload images or screenshots into Planet feedback would greatly benefit users in their ability to describe an issue and moderators in their ability to interpret submitted feedback. We should consider creating an uploadable attachments segment to the feedback submission form or upgrading the feedback textbox to a `planet-markdown-textbox`.

RyanS4:
Fixes #10340

image

## Problem

Feedback submissions only support text, making visual problems such as broken layouts, unexpected errors, and incorrect content difficult to report and diagnose.

## Solution

- Add a staged screenshot uploader to the feedback dialog for up to three PNG, JPEG, or WebP images.
- Preview selected screenshots locally without saving them until feedback is submitted.
- Attempt to convert screenshots to WebP, falling back to JPEG, with dimensions capped at 1920px. Keep the original file if browser decoding or encoding fails.
- Save the feedback document first, then upload screenshots sequentially as CouchDB attachments using the latest document revision.
- Keep successfully submitted feedback when a screenshot upload fails and warn the user that some screenshots were not uploaded.
- Keep the dialog contents available when the feedback document itself cannot be saved.
- Prevent duplicate submissions while a save or upload is in progress.
- Display only existing PNG, JPEG, or WebP attachments as linked, lazy-loaded previews.
- Preserve attachment stubs when replying, renaming, closing, or reopening feedback.
- Keep feedback messages rendered as plain text.

## PouchDB behavior change

The feedback database is no longer copied into browser IndexedDB during login. Nothing reads the `local-feedback` mirror, and screenshot attachments would make repeatedly downloading it increasingly expensive.

Login and logout now continue correctly when no active PouchDB mirrors are registered. Existing `local-feedback` stores are deleted in the background during logout so cleanup does not delay logout.

Community-to-Nation feedback synchronization is unchanged.

## Validation

- 457 tests passed across 69 test files.
- `npm run lint` passed.
- `npm run build` passed.

### DIFF
--- a/src/app/feedback/feedback-attachments.spec.ts
+++ b/src/app/feedback/feedback-attachments.spec.ts
@@ -1,0 +1,47 @@
+import { vi } from 'vitest';
+import { PendingAttachment } from '../shared/forms/file-upload.component';
+import { FEEDBACK_MAX_IMAGE_SIZE, prepareFeedbackAttachments } from './feedback-attachments';
+
+describe('feedback document attachments', () => {
+  const image = (name = 'screen.png', type = 'image/png', contents = 'image'): PendingAttachment => ({
+    file: new File([ contents ], name, { type }), originalName: name, safeName: name, contentType: type
+  });
+
+  afterEach(() => vi.restoreAllMocks());
+
+  it('supports text-only feedback', async () => {
+    expect(await prepareFeedbackAttachments().toPromise()).toEqual({});
+  });
+
+  it('preserves the actual bytes and MIME types of up to three images with unique, safe names', async () => {
+    const attachments = await prepareFeedbackAttachments([
+      image('screen shot.png'), image('screen shot.png', 'image/png', 'second'), image('photo.jpg', '', 'third')
+    ]).toPromise();
+
+    expect(Object.keys(attachments)).toEqual([ 'screenshot-screen_shot.png', 'screenshot-screen_shot-1.png', 'screenshot-photo.jpg' ]);
+    expect(attachments['screenshot-screen_shot.png']).toEqual({ content_type: 'image/png', data: btoa('image') });
+    expect(attachments['screenshot-screen_shot-1.png'].data).toBe(btoa('second'));
+    expect(attachments['screenshot-photo.jpg'].content_type).toBe('image/jpeg');
+  });
+
+  it('rejects more than three images before reading any files', async () => {
+    const reader = vi.spyOn(FileReader.prototype, 'readAsDataURL');
+    await expect(prepareFeedbackAttachments(Array.from({ length: 4 }, () => image())).toPromise()).rejects.toThrow('three');
+    expect(reader).not.toHaveBeenCalled();
+  });
+
+  it('rejects oversized, empty and non-raster images', async () => {
+    const oversized = image();
+    Object.defineProperty(oversized.file, 'size', { value: FEEDBACK_MAX_IMAGE_SIZE + 1 });
+    for (const invalid of [ oversized, image('empty.png', 'image/png', ''), image('vector.svg', 'image/svg+xml') ]) {
+      await expect(prepareFeedbackAttachments([ invalid ]).toPromise()).rejects.toThrow();
+    }
+  });
+
+  it('propagates read failures instead of producing a partial attachment set', async () => {
+    vi.spyOn(FileReader.prototype, 'readAsDataURL').mockImplementation(function () {
+      this.onerror(new ProgressEvent('error'));
+    });
+    await expect(prepareFeedbackAttachments([ image() ]).toPromise()).rejects.toThrow('Could not read');
+  });
+});
--- a/src/app/feedback/feedback-attachments.ts
+++ b/src/app/feedback/feedback-attachments.ts
@@ -1,0 +1,50 @@
+import { forkJoin, Observable, of, throwError } from 'rxjs';
+import { map } from 'rxjs/operators';
+import { PendingAttachment } from '../shared/forms/file-upload.component';
+import { normalizedContentType, safeAttachmentName } from '../shared/utils';
+
+export const FEEDBACK_MAX_IMAGES = 3;
+export const FEEDBACK_MAX_IMAGE_SIZE = 2 * 1024 * 1024;
+export const FEEDBACK_IMAGE_TYPES = [ 'image/png', 'image/jpeg', 'image/gif', 'image/webp' ];
+
+export interface FeedbackImageAttachment {
+  content_type: string;
+  data: string;
+}
+
+// Inline attachments keep the initial feedback and its screenshots in a single document write.
+export const prepareFeedbackAttachments = (
+  images: PendingAttachment[] = []
+): Observable<Record<string, FeedbackImageAttachment>> => {
+  if (images.length > FEEDBACK_MAX_IMAGES || images.some(({ file }) =>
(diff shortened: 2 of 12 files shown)


Mutugiii:
I did some thinking, and I’m not sure we should add the full Planet markdown editor to feedback. It feels like overkill when the main need is attaching screenshots, and we probably don’t need all the additional markdown formatting
options here.

I’m also concerned that using the existing community image picker would fill the shared image library with feedback screenshots and expose those screenshots in unrelated places. There are some production concerns around synchronization too: feedback is sent to the Nation, but these image resources currently aren’t, and cancelled feedback could leave unused images taking up storage on RPis.

I’d propose keeping the existing textarea and adding a simple upload/drop area for two or three images maximum. Ideally, those images would be stored as attachments on the feedback document so they sync and are cleaned up together with the feedback.

dogi:
**Re-review — rating: 82/100** (was `change`)

Verified against head `423c7a17`. CI green on that commit (22/22 `success`, suite started 2026-09-10T22:12:30Z against a head committed 2026-09-10T22:11:49Z). The `change` label predates this push; re-rating on the current head.

**The dialog rework is correct, and I checked the parts that usually aren't.** The switch from `confirm(...)` to `openDialogsForm(..., { closeOnSubmit: false, onSubmit })` depends on wiring that actually exists: `DialogsFormComponent.onSubmit` calls `dialogsLoadingService.start()` before invoking `data.onSubmit` and only calls `.stop()` when `closeOnSubmit === true`, so the directive's `finalize(() => { this.isSubmitting = false; this.dialogsLoadingService.stop(); })` is the matching half, not a stray stop. `showErrorMessage` / `closeDialogsForm` both exist and act on `this.dialogRef`. `normalizedContentType`, `safeAttachmentName` and `couchAttachmentUrl` are all present in `shared/utils.ts` on this branch.

Keeping the dialog open on failure with the text and images intact is a real improvement over the old fire-and-forget `showAlert`, and the new terminal `else` in `addFeedback` closes a path where a feedback submission previously did nothing at all.

**Three things to tidy, none blocking:**

1. The specific validation message never reaches the user. `prepareFeedbackAttachments` returns `throwError(new Error($localize\`Choose up to three PNG, JPEG, GIF or WebP images, no larger than 2 MB each.\`))`, but the subscriber's error arm discards it and shows the generic "Your feedback could not be submitted." Low impact in practice — `FileUploadComponent` now rejects over-size and wrong-type files up front with its own message, so this is the belt to that pair of braces — but the string is currently unreachable. Either surface `err.message` or drop the localized text.

2. `throwError(new Error(...))` is the deprecated RxJS 7 signature; `throwError(() => new Error(...))` is the current one.

3. `@let attachments = messageAttachments(message);` calls a method from the template, so it re-runs and allocates a new array on every change-detection pass for every message in the thread. `track attachment.name` keeps the DOM stable so there is no visible churn, but mapping the attachments once when the feedback doc loads would be cheaper.

**One thing to confirm on a live doc rather than in review:** `messageAttachments` filters on `this.feedback._attachments?.[name]?.content_type`, and `this.feedback` comes from `getFeedback`'s Mango `_find`. The spec stubs `_attachments` as present with a `stub: true` shape, which matches CouchDB's documented behaviour, but it is worth eyeballing one real feedback doc with a screenshot to be sure the stubs come back on that query — if they don't, the images silently never render and no test would catch it.

**Merge state:** one conflict against master, in `feedback.directive.ts`, from `16104256 all: smoother quote props linting (connects #10349)`. That's a quote-style reformat, so the resolution is mechanical.

**Title:** it still reads `feedback: smoother markdown support (fixes #10340)` but the head commit and the diff are screenshot attachments — no markdown left in it. Worth retitling before merge, both for the squash message and to keep it distinct from [#10437](https://github.com/open-learning-exchange/planet/pull/10437) (`all: standardize markdown rendering`), which is the PR that actually does markdown.

Separately, a design note for whoever owns the feedback store: `feedback` is the PouchDB-mirrored database, and this inlines up to 3 × 2 MB of base64 per doc, which every mirror will replicate. Fine at current volume; worth knowing.

Verdict: `ready`, replacing `change`. Resolve the lint conflict and fix the title.

---

Mutugiii:
@codex review

Mutugiii (approved):

dogi:
**Review — rating: 86/100**

Promoting to `merge`. Gate checked at head `cf983e40`: CI green on that exact commit (22 runs `success`, suite started 2026-09-14T18:33:21Z vs head committed 18:31:53Z); `git merge-tree --write-tree origin/master pr/10355` exits 0; 0 commits behind master. APPROVED by @Mutugiii at head.

**The `forkJoin` hazard is handled correctly and completely.** Emptying `databases` turns `replicateFromRemoteDBs()` and `deconfigureDBs()` into empty arrays, and `forkJoin([])` completes without emitting — which would have hung login and logout silently. I grepped every call site rather than trusting the two in the diff:

```
login-tasks.service.ts:36:    return forkJoin(this.pouchService.replicateFromRemoteDBs()).pipe(
pouch-auth.service.ts:51:      switchMap(() => forkJoin(this.pouchService.deconfigureDBs()).pipe(defaultIfEmpty([]))),
```

Those are the only two, and both are covered. (`replicateToRemoteDBs()` has no callers — pre-existing dead code, not this PR's problem.)

**The track-function change is a fix, not a style edit.** `feedback.messages` entries are `{ message, user, time }` with no `_id`, so `trackById($index, message)` returned `undefined` for every row. `track $index` is correct for an append-only list.

**The display-side allowlist is good defense.** `messageAttachments()` filters on `this.feedback._attachments?.[name]?.content_type` against `FEEDBACK_SCREENSHOT_TYPES`, so a message referencing an attachment that failed to upload silently renders nothing rather than a broken image — which pairs well with `uploadScreenshots` counting failures and still reporting success for the feedback itself. The rev-chaining through `reduce` is right: a failed PUT doesn't change the rev, so carrying the old one forward is correct.

**One verified gap worth a follow-up.** CodeRabbit flagged a missing size cap; I checked and it's real. There is no size enforcement anywhere in the path — in `file-upload.component.ts`, `size` appears only in `formattedSize()` for display, and the default hint even advertises "up to 512 MB" with nothing enforcing it. That matters here because `normalizeImage`'s `fallback()` returns the **original** `file` untouched on every failure path (no dimensions, no 2d context, encode failure, or any throw):

```ts
const fallback = (): NormalizedImage => ({
  file,
  contentType: normalizedContentType(file),
  fileName: safeAttachmentName(file.name, opts.usedNames)
});
```

So the one case where re-encoding fails — a huge image that exhausts canvas memory — is exactly the case that bypasses the 1920px bound. Capped at 3 files and needs a fallback trigger, so it's a nit, not a blocker, but a `maxSize` on `FileUploadComponent` would close it.

**Two smaller notes.** `addFeedback`'s `if/else if` chain has no final `else`, so a fall-through would leave `isSubmitting = true` and wedge the dialog permanently; the new `.replace(/\/+$/, '')` makes that unreachable in practice (it's what stops `lastPart` being empty), but a defensive `else` costs nothing. And the legacy `local-feedback` destroy in `deconfigureDBs()` is fire-and-forget and runs on every logout forever — `new PouchDB('local-feedback')` creates the database if absent just to destroy it.

---

dogi (approved):

Issue #10342, opened 2026-08-25
Community: unrestricted meetup editing


RyanS4:
Currently, any member of a planet can modify any meeting listed on the meeting calendar. We should consider a toggleable option on meetup creation that allows the creator to specify who can edit (everyone vs. only self/admins); alternatively, we can always restrict it to the "only self/admin" permissions. Fixing this will ensure planet members do not alter meetups that are intended not to be modified without permission.

PR #10345, opened 2026-08-25
teams: smoother meetups permissions guarding (fixes #10342)

RyanS4:
Fixes #10342

## Summary

- Centralizes meetup edit and delete authorization in `MeetupService`.
- Allows meetup creators, site administrators, community leaders, and leaders of the meetup's linked team to manage local meetups.
- Makes parent and remote contexts read-only for every local role and fails closed when a meetup document is missing.
- Applies authorization to list actions, detail dialogs, direct update routes, mutation-time updates, and single or bulk deletion.
- Passes authoritative team-leader context through the team calendar and meetup dialogs.
- Redirects away from failed update-route loads instead of leaving an active Add form.
- Restores the previously excluded meetup specs and adds coverage for the complete authorization matrix.

## Authorization behavior

- Site administrators and community leaders may manage all local meetups, including team-linked meetups.
- Team leaders may manage only meetups linked to their team.
- Creators may manage their own local meetups.
- Parent and remote meetups remain read-only.
- Missing meetup documents deny every role.
- Local community meetup creation remains available; this PR restricts editing and deletion.

## Summary by CodeRabbit

- **New Features**
  - Added consistent meetup authorization for viewing, editing, managing, and deleting meetups.
  - Authorized creators, administrators, community leaders, and relevant team leaders can manage meetups.
  - Team leader context now carries through calendars and meetup dialogs.
  - Unauthorized edit or delete attempts display an alert; failed meetup loads return users to the meetups list.
  - Meetup actions now respect read-only contexts.

- **Tests**
  - Added coverage for authorization rules, deletion behavior, editing flows, and team leader context propagation.

PR #10364, opened 2026-08-31
communty: smoother calendar event rescheduling (fixes #10356)

### LINKED ISSUE #10356: Community: unpersisted event drag-and-drop causes ghost rescheduling
### Context & Relationship to #10342 / #10345
PR #10345 (fixing #10342) established that meetup edit/delete actions are restricted to the meetup creator (`createdBy`) and planet administrators (`isUserAdmin`). This issue addresses event dragging on the FullCalendar widget.

---

### Problem
In `PlanetCalendarComponent` (`src/app/shared/calendar.component.ts`), all calendar event objects (meetups and task deadlines) currently have `editable: true` hardcoded on the event payload.

Because `calendarOptions` does not implement an `eventDrop` (or `eventResize`) handler:
1. **Ghost Rescheduling**: Any user (including unauthorized members) can drag and drop meetups or task deadlines to different days on the calendar grid. The UI visually moves the event box, leading users to believe the event was rescheduled.
2. **Missing Persistence**: No update is sent to CouchDB. Upon page refresh or navigating away and back, the event snaps back to its original date.
3. **Bypasses Edit Permissions**: Allowing events to be dragged on the calendar bypasses the permission boundaries enforced in #10342 / #10345.

#### Option 1: Disable Event Dragging (`editable: false`)
- Remove `editable: true` (or set `editable: false`) on calendar event objects in `src/app/shared/calendar.component.ts`.
- Users reschedule meetups by clicking the event to open the edit modal dialog, which already enforces creator/admin authorization via #10345 and supports recurring event schedules.

#### Option 2: Implement `eventDrop` with Permission Checks & Persistence
- Add an `eventDrop` handler to `calendarOptions`.
- Check whether the logged-in user is an administrator or the meetup creator (`user.isUserAdmin || user.name === meetup.createdBy`). If unauthorized or if the event is recurring, call `info.revert()` and show an alert.
- If authorized, calculate the updated `startDate` / `endDate` and persist the change to CouchDB via `CouchService.updateDocument()`.

### Justification & Practical Impact
- Eliminates misleading UI behavior where users believe a meeting or task deadline was changed when it was not.
- Prevents confusion caused by events reverting to their original positions upon page refresh.
- Ensures the calendar view stays consistent with the permission model established in #10342 / #10345.

RyanS4:
Fixes #10356

### Problem
Calendar events could be dragged without persisting their new dates, causing them to return to their original positions after refresh. Drag affordances also did not consistently reflect meetup permissions, recurring-event restrictions, or read-only calendar contexts.

### Changes
- Persist meetup and task rescheduling through FullCalendar’s `eventDrop` callback.
- Reuse the shared meetup authorization policy for creator, administrator, community-leader, team-leader, and read-only contexts.
- Prevent recurring meetups and event durations from being changed through unsupported drag operations.
- Shift meetup dates by calendar days so rescheduling remains correct across daylight-saving transitions.
- Preserve open-ended meetups, returned CouchDB revisions, active event filters, and successful writes when later refreshes fail.
- Notify shelved meetup participants through a shared, paginated workflow.
- Keep notification delivery best-effort while logging failures for diagnosis.
- Preserve pointer affordances for clickable, non-draggable events.
- Revert calendar movement only when the underlying write fails.

### Verification
- **Lint**: `npm run lint` passed with 0 errors.
- **Unit Tests**: `npx vitest run src/app/shared/calendar.component.spec.ts` passed (7/7 tests).

## Summary by CodeRabbit

* **New Features**
* Added drag-and-drop rescheduling for calendar events and task deadlines.
* Editable meetups now follow permission, recurrence, and read-only restrictions.
* Successful meetup changes automatically notify relevant participants.

* **Bug Fixes**
* Rescheduling now correctly handles daylight saving time and open-ended dates.
* Invalid, unauthorized, recurring, or unsupported moves are reverted.
* Calendar changes revert when saving fails.
* Loading indicators remain accurate during save and refresh operations.
* Meetup updates complete successfully even if notifications cannot be sent.

### DIFF
--- a/src/app/shared/calendar.component.spec.ts
+++ b/src/app/shared/calendar.component.spec.ts
@@ -87,181 +87,41 @@ describe('PlanetCalendarComponent', () => {
     expect(component.tasks[1].textColor).toBe(styleVariables.accentText);
   });
 
-  it('sets editable to true for creator or admin on non-recurring meetups', () => {
-    const adminUser = { name: 'admin', isUserAdmin: true, _id: 'org.couchdb.user:admin' };
-    const component = createComponent({}, { get: () => adminUser });
+  it('sets editable based on user authorization and recurring status', () => {
+    const adminComponent = createComponent({}, { get: () => ({ name: 'admin', isUserAdmin: true, _id: 'admin' }) });
+    const userComponent = createComponent({}, { get: () => ({ name: 'user', isUserAdmin: false, _id: 'user' }) });
 
-    const event = component.eventObject({ title: 'Meetup', createdBy: 'otherUser', recurring: 'none' }, new Date(), new Date());
-    expect(event.editable).toBe(true);
+    expect(adminComponent.eventObject({ title: 'Event', createdBy: 'other', recurring: 'none' }, new Date()).editable).toBe(true);
+    expect(userComponent.eventObject({ title: 'Event', createdBy: 'other', recurring: 'none' }, new Date()).editable).toBe(false);
+    expect(adminComponent.eventObject({ title: 'Event', createdBy: 'admin', recurring: 'daily' }, new Date()).editable).toBe(false);
   });
 
-  it('sets editable to false for non-creator, non-admin users on meetups', () => {
-    const regularUser = { name: 'learner', isUserAdmin: false, _id: 'org.couchdb.user:learner' };
-    const component = createComponent({}, { get: () => regularUser });
+  it('updates document on authorized eventDrop and reverts when unauthorized', () => {
+    const couchService = { updateDocument: vi.fn(() => of({ ok: true })), findAll: vi.fn(() => of([])) };
+    const adminComponent = createComponent(couchService, { get: () => ({ name: 'admin', isUserAdmin: true, _id: 'admin' }) });
+    const userComponent = createComponent(couchService, { get: () => ({ name: 'user', isUserAdmin: false, _id: 'user' }) });
+    const revert = vi.fn();
 
-    const event = component.eventObject({ title: 'Meetup', createdBy: 'admin', recurring: 'none' }, new Date(), new Date());
-    expect(event.editable).toBe(false);
-  });
+    const meetup = { _id: 'm1', title: 'Meetup', createdBy: 'admin', recurring: 'none', startDate: 1000, endDate: 1000 };
 
-  it('sets editable to false for recurring meetups even if user is admin', () => {
-    const adminUser = { name: 'admin', isUserAdmin: true, _id: 'org.couchdb.user:admin' };
-    const component = createComponent({}, { get: () => adminUser });
+    // Unauthorized revert
+    userComponent.eventDrop({ event: { extendedProps: { meetup } }, revert });
+    expect(revert).toHaveBeenCalled();
 
-    const event = component.eventObject({ title: 'Recurring Meetup', createdBy: 'admin', recurring: 'weekly' }, new Date(), new Date());
-    expect(event.editable).toBe(false);
-  });
-
-  describe('eventDrop', () => {
-    it('reverts and alerts when unauthorized user drops a meetup', () => {
-      const regularUser = { name: 'learner', isUserAdmin: false, _id: 'org.couchdb.user:learner' };
-      const messageService = { showMessage: vi.fn(), showAlert: vi.fn() };
-      const revert = vi.fn();
-      const component = createComponent({}, { get: () => regularUser }, messageService);
-
-      const info = {
-        event: {
-          extendedProps: {
-            meetup: { _id: 'm1', title: 'Meetup', createdBy: 'teacher', startDate: 1000 }
-          }
-        },
-        revert
-      };
-
-      component.eventDrop(info);
-
-      expect(revert).toHaveBeenCalled();
-      expect(messageService.showAlert).toHaveBeenCalled();
+    // Authorized update
+    const oldStart = new Date(1000);
+    const newStart = new Date(2000);
+    adminComponent.eventDrop({
+      event: { start: newStart, extendedProps: { meetup } },
+      oldEvent: { start: oldStart },
+      revert: vi.fn()
     });
-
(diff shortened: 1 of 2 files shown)


dogi:
**Review — rating: 68/100**

Verified against head `39af4b02`. CI green (21/21 `success`).

**The delta math is right, and I want to say so explicitly because it looks wrong at first glance.** `deltaMs` is computed from `dateAtTime(startDate, startTime)` — a normalized midnight-plus-`startTime` baseline — rather than from the raw `startDate`, and the result is then added back to the raw `startDate`. My first read was that an intra-day drag would shift `startDate` by hours while leaving `startTime` untouched, desyncing the doc from what `eventObject` renders. It can't happen: `calendarPlugins` is `[ dayGridPlugin, interactionPlugin ]` with `initialView: 'dayGridMonth'` and no timeGrid plugin, so every drop lands on a day cell and `deltaMs` is always a whole number of local days. Preserving the original time-of-day component of `startDate` is then exactly correct, and it stays correct across DST because `dateAtTime` uses local `setHours`. Good choice of baseline.

**The blocker is the permission check.** `canEditMeetup` is added as a private method on the calendar component:

```ts
canEditMeetup(meetup: any): boolean {
  const user = this.userService.get();
  return !!user?._id && (user.isUserAdmin || user.name === meetup?.createdBy);
}
```

Master already has this, in `src/app/meetups/meetups.service.ts:38`:

```ts
canEditMeetup(meetup: any, context: MeetupAuthorizationContext = {}): boolean {
  const user = this.userService.get();
  if (!user?._id || !meetup || context.readOnly) { return false; }
  const isCommunityLeader = user.roles?.includes('leader') === true;
  const isTeamLeader = !!context.leaderOfTeamId && meetup?.link?.teams === context.leaderOfTeamId;
  return user.isUserAdmin || isCommunityLeader || isTeamLeader || user.name === meetup?.createdBy;
}
```

It is not present at this branch's merge-base — it landed on master afterwards, in `d021fa83 teams: smoother meetups permissions guarding (fixes #10342) (#10345)`, which is also one of the two commits this branch now conflicts with. So the duplicate isn't careless, it's staleness; but merged as-is it ships a second, weaker copy of the rule that silently drops the community-leader and team-leader cases. A team leader who can edit a meetup everywhere else in the app cannot drag it on the calendar, and the two definitions will drift from here on.

The context the canonical method wants is already on this component — `@Input() leaderOfTeamId?: string` at line 60. Inject `MeetupsService` and call `canEditMeetup(meetup, { leaderOfTeamId: this.leaderOfTeamId, readOnly: !this.editable })`, and both the `editable` computation in `eventObject` and the `eventDrop` guard collapse into one source of truth. That also gets you `#10342`'s `readOnly` handling for free.

**Two smaller ones in `eventDrop`:**

- `const deltaMs = (info.event.start ? info.event.start.getTime() : 0) - originalStartDate.getTime();` — the `: 0` fallback doesn't degrade gracefully, it produces `deltaMs ≈ -originalStartDate.getTime()` and writes a 1970 date to the doc. The `!eventData` case above already does the right thing; `info.revert()` and return here too.
- `const newEndDate = eventData.endDate ? Number(eventData.endDate) + deltaMs : newStartDate;` invents an `endDate` on meetups that never had one, changing the doc shape. Preserve absence.

I checked `console.error` against the lint config before flagging it — `eslint.config.mjs` line 200 lists `error` in the `no-console` allow list, so those calls are fine.

**Merge state:** two conflicts against master, both in files master has moved under this branch — `calendar.component.ts` and its spec — across three commits (`d021fa83` permissions guarding, `b99f2e4b` calendar resizing, `42c11a71` nation exchanging). Master's `calendarOptions.select` now carries an `if (!this.editable) return;` guard this branch's copy doesn't have, and there's a new `ResizeObserver`/`NgZone` block. Resolve carefully rather than taking either side wholesale.

Verdict: `change`. Reuse `MeetupsService.canEditMeetup`, fix the two `eventDrop` fallbacks, and merge master in — the drag-and-drop core itself is fine and I'd expect a quick turnaround.

---

Mutugiii (approved):

dogi:
**Re-review — rating: 88/100** (was 68)

Promoting `review` → `merge`. Gate at head `4bde8407`:
- **CI:** 22/22 check runs succeeded. All started 2026-09-18T10:52:34Z, after the head commit at 10:52:08Z.
- **Merge:** `git merge-tree --write-tree origin/master pr/10364` exits 0. It also stacks cleanly with the other nine PRs in this batch.
- **Approval:** APPROVED by Mutugiii at `4bde8407`.

Findings from the 68/100 review, checked against head:
1. **Duplicate permission check:** fixed. `canEditMeetup` now delegates to `this.meetupService.canEditMeetup(meetup, { leaderOfTeamId: this.leaderOfTeamId, readOnly: !this.editable })`. `MeetupService` is changed to `providedIn: 'root'`, and the module-level provider is removed.
2. **1970 fallback date:** fixed. `if (!eventData || !info.event?.start) { info.revert(); return; }`
3. **`endDate` added to meetups without one:** fixed. `...(eventData.endDate ? { endDate: this.shiftStoredDate(eventData.endDate, info.delta) } : {})`
4. **Conflicts with master:** gone. The `select` guard `if (!this.editable) { return; }` and the `ResizeObserver` block are both kept.

New since that review:
- `notifyMeetupChange` is moved into `NotificationsService` and shared with `meetups-add`. It uses `findAll` and skips an empty `_bulk_docs`.
- A failed notification or refetch no longer undoes a reschedule that has already been saved.
- `shiftStoredDate` moves dates by `delta.days` as calendar days, so a drag across a DST change doesn't shift the event by an hour.

Checked and dropped:
- **Time-grid drift** (`startDate` shifted by milliseconds while `startTime` stays fixed): not reachable. `plugins: [ dayGridPlugin, interactionPlugin ]` means the drag delta is whole days only.
- **Stale `_rev` on a second drag before the refetch:** `replaceCachedEvent` puts `res.doc` into `extendedProps`.

---

dogi (approved):

Issue #10327, opened 2026-08-21
Resources: add copy resource link action to resource viewer


RyanS4:
### Problem
- **Missing Link-Sharing Action**: In `ResourcesViewComponent` (`src/app/resources/view-resources/resources-view.component.html`), the toolbar currently provides actions to open attachments in a new tab, add/remove from `myLibrary`, or download files, but lacks any quick action to copy a resource's direct URL to the clipboard.
- **Friction in Educational Sharing**: When educators, team leaders, or learners want to cite or recommend a specific book, video lecture, or document inside **Community Voices** discussions, **Team Tasks**, or **Chat**, they must manually highlight and copy the browser address bar.
- **Mobile & Tablet Inconvenience**: On mobile devices, tablets, or kiosk workstations, selecting the browser URL bar is cumbersome and often copies unwanted query parameters or internal IP addresses.

### Proposed Solution
1. **Toolbar Copy Action Button (`resources-view.component.html`)**:
   - Add a dedicated "Copy Link" action icon button (`<button mat-icon-button matTooltip="Copy Link" i18n-matTooltip (click)="copyLink()"><mat-icon>content_copy</mat-icon></button>`) to the resource header toolbar and mobile action menu.

### Justification & Practical Impact
- **Streamlined Cross-Module Collaboration**: Makes citing and linking library resources effortless across Community Voices, Team assignments, and chat conversations.
- **Consistency Across Planet Modules**: Matches existing 1-click clipboard copy patterns already used in Community Voices (`/voices/:id`), Achievements (`/achievements/:id`), and Admin Reports.

PR #10375, opened 2026-09-01
all: smoother ratings clearing (fixes #7727)

RyanS4:
Fixes #7727

image

## Description
- Adds a clear ('X') button next to active stars in the rating dialog to allow learners to reset/clear ratings.
- Deletes the rating document from CouchDB upon submitting a 0-star rating and recalculates aggregate rating statistics.
- Adds unit tests covering rating clearing and CouchDB document deletion.

PR #10339, opened 2026-08-25
all: smoother links copying (fixes #10327)

RyanS4:
Fixes #10327
- Adds 1-click "Copy Resource Link" button to resource viewer toolbar and action menus
- Injects Angular CDK Clipboard to copy clean direct resource URL
- Localized feedback toast notification when copied

{00001802-B429-4F7E-8B2A-3C483B51E6C9}

PR #10377, opened 2026-09-01
courses/resources: smoother rating prompt on leave (fixes #6640)

### LINKED ISSUE #6640: Ratings: Encouraging more resource & course ratings
We would like to increase the number of resources and courses that are rated in the field.  Right now it's very passive and people are consuming content but not rating it.

A couple ideas to get things started
1. When the user routes away from a course or resource, the rating dialog pops up if the user has not rated it yet and prompts a rating
2. Same situation as above, but instead we use a [snackbar](https://material.angular.io/components/snack-bar/overview) to ask for a rating.  Clicking on the action opens the rating dialog

RyanS4:
## Description
- Implements rating prompt functionality to encourage learners to rate educational courses and library resources upon navigating away from them if they have not yet submitted a rating.
- Adds `promptRating()` to `RatingService`:
  - Validates that the active user is logged in and enrolled in the resource/course (`inShelf: true`).
  - Skips prompting if the user has already submitted a rating (`userRating.rate > 0`).
  - Prompts with the rating modal form (`DialogsFormService.confirm()`) containing star rating selection and optional comments.
  - On submission (`rate > 0`), persists the rating document to CouchDB (`ratings` database), refreshes aggregate rating statistics via `newRatings(parent)`, displays a localized toast (*"Thank you, your rating is submitted!"*), and completes navigation.
  - On dismissal or cancellation, allows navigation to proceed smoothly without blocking the user.
- Integrates `canDeactivate()` on `CoursesViewComponent` and `ResourcesViewComponent` using `UnsavedChangesGuard` on `view/:id` routes in `courses-router.module.ts` and `resources-router.module.ts`.
- Skips the rating prompt when:
  - Entering edit/update mode (`/courses/view/:id/update`, `/courses/update/:id`, `/resources/update/:id`).
  - Navigating to internal course steps, exams, surveys, or leader progress views (`/step/`, `/exam`, `/survey`, `/progress/`, `/enrolled/`).
  - The current user is an admin or creator managing the content (`canManage: true`).
- Adds unit tests covering prompt condition checks, submit handling, edit route skips, and route deactivation delegation.

Fixes #6640

## Summary by CodeRabbit

- **New Features**
  - Added rating prompts for eligible courses and resources when leaving their viewing pages.
  - Users can submit ratings and comments, with confirmation or error feedback.
  - Rating prompts are skipped when an item is already rated or not eligible.

- **Bug Fixes**
  - Improved navigation handling to prevent accidental loss of unsaved changes.
  - Rating prompts no longer appear when navigating to editing or related workflow pages.

### DIFF
--- a/src/app/courses/courses-router.module.ts
+++ b/src/app/courses/courses-router.module.ts
@@ -16,7 +16,7 @@ const routes: Routes = [
   { path: 'add', component: CoursesAddComponent },
   { path: 'update/:id', component: CoursesAddComponent },
   { path: 'view/:id/update', component: CoursesAddComponent },
-  { path: 'view/:id', component: CoursesViewComponent },
+  { path: 'view/:id', component: CoursesViewComponent, canDeactivate: [UnsavedChangesGuard] },
   { path: 'exam', component: ExamsAddComponent },
   { path: 'survey', component: ExamsAddComponent },
   { path: 'view/:id/step/:stepNum', component: CoursesStepViewComponent, },
--- a/src/app/courses/view-courses/courses-view.component.ts
+++ b/src/app/courses/view-courses/courses-view.component.ts
@@ -1,13 +1,14 @@
 import { Component, OnInit, OnDestroy } from '@angular/core';
 import { ActivatedRoute, ParamMap, Router } from '@angular/router';
 import { MatMenuTrigger, MatMenu, MatMenuItem } from '@angular/material/menu';
-import { Subject, defer } from 'rxjs';
+import { Observable, Subject, defer } from 'rxjs';
 import { takeUntil, switchMap, take, filter, map } from 'rxjs/operators';
 import { UserService } from '../../shared/user.service';
 import { CoursesService } from '../courses.service';
 import { SubmissionsService } from '../../submissions/submissions.service';
 import { StateService } from '../../shared/state.service';
 import { DeviceInfoService, DeviceType } from '../../shared/device-info.service';
+import { RatingService } from '../../shared/forms/rating.service';
 import { trackByIndex } from '../../shared/table-helpers';
 import { MatToolbar } from '@angular/material/toolbar';
 import { MatIconButton, MatButton } from '@angular/material/button';
@@ -74,6 +75,7 @@ export class CoursesViewComponent implements OnInit, OnDestroy {
   deviceTypes: typeof DeviceType = DeviceType;
   courseIcons = courseIcons;
   trackByFn = trackByIndex;
+  promptedForRating = false;
   constructor(
     private router: Router,
     private userService: UserService,
@@ -82,7 +84,8 @@ export class CoursesViewComponent implements OnInit, OnDestroy {
     private submissionsService: SubmissionsService,
     private stateService: StateService,
     private deviceInfoService: DeviceInfoService,
-    private dialog: MatDialog
+    private dialog: MatDialog,
+    private ratingService: RatingService
   ) {
     this.deviceInfoService.watchDeviceType().pipe(takeUntil(this.onDestroy$)).subscribe((deviceType) => {
       this.deviceType = deviceType;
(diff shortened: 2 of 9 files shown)


Mutugiii:
on hold for #10375

dogi:
**Review — rating: 82/100**

Verified against head `e3a4aed3`. CI green on that commit — 24 check runs, 22 `success` and 2 `planet labels` `skipped`, which is workflow noise rather than a failure.

**I chased four ways this could hang or misfire, and all four hold up.** Worth recording so nobody re-does the work:

- *Navigation can't deadlock.* `promptRating` returns `dialogsFormService.confirm(...)`, which is `dialogRef.afterClosed()`. `confirm` passes `disableClose: confirmUnsavedChanges` and the default is `false`, so ESC and backdrop clicks both close, and `afterClosed()` always emits and completes. The router is never left waiting on a dialog that can't be dismissed.
- *The guard handles an Observable return.* `UnsavedChangesGuard` only intercepts the literal `result === false`; anything else, including the Observable from `promptRating`, is returned untouched to the router. Threading `nextState` through `CanComponentDeactivate` is a clean extension and doesn't disturb the existing `false` path.
- *The already-rated check actually works.* `promptRating` bails on `item.rating?.userRating?.rate > 0`, which is only useful if `rating` is populated on the detail object. It is: `CoursesService.mergeData` sets `rating: this.ratingService.createItemList([ course ], ratings)[0].rating`, and `ResourcesService.setResources` runs the resources through the same `createItemList`. Users who have already rated won't be re-prompted.
- *Logging out does not trigger a rating prompt.* `HomeComponent` logout calls `this.userService.unset()` before `this.router.navigate([ '/login' ])`, so by the time `canDeactivate` runs, `promptRating`'s `if (!user?._id) return of(true)` short-circuits. I also checked `planet-resources-menu` for the same reason — it opens `DialogsResourcesViewerComponent` rather than navigating, so opening a resource from inside a course doesn't fire the prompt either.

No module cycle from the new `RatingService` → `DialogsFormService` edge: `dialogs-form.component.ts` imports `PlanetRatingStarsComponent`, but `planet-rating-stars.component.ts` imports only `UserService`, not `rating.service.ts`.

**The one real change to make.** `DialogsFormService` gains `providedIn: 'root'`, which it needs — `RatingService` is root-provided and previously the only provider was `PlanetDialogsModule`, so injecting it from root would have thrown `NullInjectorError`. But the module provider is still there:

```ts
// src/app/shared/dialogs/planet-dialogs.module.ts
providers: [
  DialogsFormService,
  DialogsListService
]
```

Two providers means two instances, and `DialogsFormService` is no longer stateless — it holds `private dialogRef`, which `closeDialogsForm()` and `showErrorMessage()` act on. That is latent today because `confirm()` uses a local `dialogRef`, so `promptRating` is unaffected. It stops being latent the moment something opens via one injector and closes via the other — and [#10355](https://github.com/open-learning-exchange/planet/pull/10355) is currently building exactly that cross-call `dialogRef` flow into the feedback dialog. Drop `DialogsFormService` from the module's `providers` array in this PR and the ambiguity never exists.

**Two nits.** The exclusion list is substring matching on the raw URL — `nextUrl.includes('/exam')` also matches `/exams`, and the list has to be extended by hand every time a sub-route is added. Matching route config or a data flag would age better. And `promptedForRating` is set before the dialog resolves; harmless since `promptRating` always emits `true`, but it reads as a guard against double-prompting when it's really a one-shot latch.

**Merge state.** Two conflicts against master, in `resources-view.component.ts` and its spec, from `67e87db1 all: smoother links copying (fixes #10327) (#10339)`.

**Merge ordering — read this before scheduling.** This PR and [#10439](https://github.com/open-learning-exchange/planet/pull/10439) (`courses: refactor ratings service`) produce 4 conflicts against each other, the worst pair in the current queue; both rewrite `rating.service.ts`. Whichever lands second will need a real rebase, not a mechanical one. Given [#10439](https://github.com/open-learning-exchange/planet/pull/10439) is a refactor of the same service this PR extends, landing [#10439](https://github.com/open-learning-exchange/planet/pull/10439) first and rebuilding `promptRating` on top of the refactored shape is likely cheaper than the reverse.

Verdict: `ready`. Remove the duplicate `DialogsFormService` provider, resolve the master conflict, and sequence against [#10439](https://github.com/open-learning-exchange/planet/pull/10439).

---

PR #10379, opened 2026-09-01
life: smoother achievements deleting (fixes #8097)

### LINKED ISSUE #8097: deleting achievements has no confirmation prompt

RyanS4:
## Description
- Adds confirmation dialog prompts before deleting achievements, references, links, or uploaded link attachments in the user achievements editor to prevent accidental data loss.
- Configures `[confirmDelete]="true"`, custom item names, and localized confirmation prompts across the achievements, references, and links `<planet-step-list>` instances in `users-achievements-update.component.html`.
- Extends `PlanetStepListComponent` with `@Input() deletePromptMessage?: string` to support tailored confirmation prompt text per item type while defaulting to *"Are you sure you want to delete the following step?"*.
- Updates `removeExistingResume()` in `UsersAchievementsUpdateComponent` to open a confirmation dialog (`DialogsPromptComponent`) displaying the resume file name before marking the attachment for deletion.
- Adds unit tests in `planet-step-list.component.spec.ts` and `users-achievements-update.component.spec.ts` verifying prompt generation and deletion handling.

Fixes #8097

## Summary by CodeRabbit

- **New Features**
  - Added customizable delete-confirmation messages for planet steps.
  - Enabled localized delete confirmations for achievements, references, and links.
  - Added confirmation before deleting an existing resume.
  - Resume deletion now occurs only after confirmation.
  - Added achievement sorting behavior that resets after confirmed deletion.

- **Bug Fixes**
  - Prevented step-change notifications when a move does not succeed.
  - Preserved resume replacement selections when deleting an existing resume.

- **Tests**
  - Expanded coverage for deletion prompts, step changes, sorting, and resume deletion confirmation.


Mutugiii (approved):

dogi:
**Review — rating: 83/100**

The core change is right. Making `performStepMove` / `moveFormArrayStep` return a boolean and gating the navigation updates on it fixes the case where a failed delete still collapsed the open step, and adding the `stepsChange` emission on the `FormArray` branch is what lets the achievements list drop its `PlanetStepListService` subscription in favour of `(stepsChange)="resetAchievementSort()"`.

I checked that the new `FormArray` emission can't leak into other consumers: `courses-step.component.html` is the only template binding `(stepsChange)`, and its `@Input() steps: any[]` is a plain array, so it was already emitting and is unaffected. `stepMoveClick$` is still consumed by `PlanetStepListComponent` itself, so the service isn't orphaned. `of` is already imported, and `spinnerOn` is a real `DialogsPromptComponent` option (defaults `true`) — passing `false` is correct given `request: of(true)`.

Two things to look at:

**1. `removeExistingResume()` silently dropped `clearResumeSelection()`.** Before:

```ts
this.resumeMarkedForDeletion = true;
this.currentResumeFileName = '';
this.clearResumeSelection();
```

after, that third call became `this.updateUnsavedChangesFlag()`. Those aren't equivalent — `clearResumeSelection()` also nulls `resumeFile`, clears `resumeUploadError` and calls `resumeInput?.clear()`. So if a user picks a replacement PDF and *then* confirms "remove existing", the pending file now survives. Given `onResumeSelected` sets `resumeMarkedForDeletion = false`, the two flags end up disagreeing. This may well be the behaviour you want, but it's outside "delete confirming" and isn't mentioned anywhere — please make it deliberate.

**2. The stale-target guard fails silently.** In `promptDeleteStep`, if `targetUnchanged` is false the handler has already called `dialogRef.close()` and then just returns — the user clicks OK, the dialog closes, and nothing is deleted with no message. In practice the guard should hold (a modal blocks the interactions that would rewrite `stepControl.value`), so this is a rare path, but a `planetMessageService` alert would beat a no-op.

**What blocks `merge`: the approval is stale.** mutugi approved `6b3bd93e` on 2026-09-08; `ce3e9d1c address non-existent nav` and `096980e4 preserve identity` landed afterwards — and `096980e4` is the `targetUnchanged` guard discussed above. CI is green and it merges clean.

Moving to `review` for re-approval at `096980e4`.

---

dogi (approved):

PR #10387, opened 2026-09-03
teams: smoother member adding loader handling (fixes #10386)

RyanS4:
Fixes #10386

### Description
When a team leader manually adds a member to a team via the "Add Members" dialog, the membership document is successfully written to CouchDB, but the UI remains stuck on an infinite loading popup with the Planet logo.

This was caused by two issues:
1. `DialogsAddTableComponent` was starting the loading spinner by default because `noSpinner: true` wasn't passed in `openInviteMemberDialog()`. When `addMembers()` then started its own spinner, `DialogsLoadingService` incremented its internal reference counter twice. Because `addMembers()` only decrements the count once upon completion, the spinner's close condition (`requestCount === 0`) was never met, leaving the modal on screen permanently.
2. In `UsersService`, `usersUpdated` was a stateless `Subject<any>()` without snapshot replay. When `TeamsService.getTeamMembers()` refreshed the member list using `usersListener(true).pipe(take(1))`, it risked hanging if no new user change events occurred in CouchDB.

### What's Changed
- **Prevent Double Loading Spinner**: Passed `noSpinner: true` in `openInviteMemberDialog()` so `DialogsAddTableComponent` leaves loading management to `addMembers()`, ensuring loader `start()` and `stop()` calls stay balanced 1:1.
- **Replay User Snapshot**: Converted `usersUpdated` in `UsersService` to `ReplaySubject<any[]>(1)` so late subscribers immediately receive the most recent user snapshot without hanging.
- **Resilient Member Refresh**: Added a timeout and fallback in `TeamsService.getTeamMembers()` along with `catchError` handlers around secondary CouchDB queries.
- **Unit Tests**: Added targeted tests verifying loader dismissal with `noSpinner: true`, late user snapshot replay, and resilient team member resolution.

image

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved loading behavior when inviting team members to prevent duplicate spinner activity.
  * Ensured the invite dialog closes correctly after a member is added.

PR #10371, opened 2026-09-01
manager: smoother myplanet app source reporting (fixes #10380)

Mutugiii:
Fixes #10380

## Summary

- Centralizes activity-source classification for Planet, myPlanet, and myPlanet Lite while preserving the legacy `androidId` fallback.
- Adds source filtering and labels to detailed reports, submissions, and CSV exports.
- Separates standalone myPlanet report rows by device and app source.
- Stamps new Planet login activities with `app: planet`.

App identifiers are normalized case-insensitively. Recognized `app` values take precedence over legacy device detection.

## Validation

- `npm run lint`
- `npm run i18n:check`
- 14 focused app-source and report tests

## Summary by CodeRabbit

- **New Features**
  - Added app source information to MyPlanet reports, including Planet, myPlanet, and myPlanet Lite.
  - Added sortable Source columns to report tables and CSV exports.
  - Added source filters with clearer labels and responsive wrapping.
  - Submission exports now display localized app source labels.

- **Bug Fixes**
  - Activities from different app sources on the same device are now grouped separately.
  - Improved handling of activity data and missing device identifiers in reports and exports.

PR #10388, opened 2026-09-03
actions: smoother workflow labeling (fixes #10412)

### LINKED ISSUE #10412: labels workflow re-runs on every PR edit and cancels itself
The Planet labels workflow lists `edited` among its `pull_request_target` trigger types, so it fires on every title or description rewrite.

CodeRabbit rewrites PR descriptions several times per PR, which queues multiple workflow runs that cancel each other. The net effect is that PRs are left unlabelled.

The `if: ${{ github.event.action != 'edited' || github.event.changes.base }}` guard in the job was an attempt to work around this, but the runs are still queued and cancelled before the guard is evaluated.

Dropping `edited` from the trigger types removes both the cascading runs and the need for the guard. Retargeted PRs — which change the diff without a push — can still be relabelled manually via `workflow_dispatch`.

dogi:
Fixes #10412

## Summary

Removes the `edited` event type from the `pull_request_target` trigger in the planet labels workflow and simplifies the conditional logic that was working around it.

## Changes

- **Removed `edited` from workflow trigger types**: The workflow now only runs on `opened`, `synchronize`, `reopened`, and `ready_for_review` events. The `edited` event was causing excessive workflow runs whenever CodeRabbit or other tools rewrote PR titles/descriptions, leading to cascading cancellations and skipped labelling checks.
- **Removed conditional guard**: Deleted the `if: ${{ github.event.action != 'edited' || github.event.changes.base }}` condition that was attempting to skip runs on edits (except retargeting). This is no longer needed since the trigger itself excludes the `edited` event.

## Rationale

The `edited` event fires on every title or description rewrite. CodeRabbit rewrites PR descriptions multiple times, which queued multiple workflow runs that cancelled each other and resulted in PRs remaining unlabelled. Manual relabelling via `workflow_dispatch` is available for retargeted PRs (which change the diff without a push).

https://claude.ai/code/session_01CsgFLJCRY6VhZrkN27EBib

## Summary by CodeRabbit

- **Chores**
  - Updated automated pull request labeling behavior.
  - Labeling checks no longer run for pull request edits, and size-based processing now runs without the previous event-specific exclusion.

### DIFF
--- a/.github/workflows/labels.yml
+++ b/.github/workflows/labels.yml
@@ -2,7 +2,13 @@ name: Planet labels
 
 on:
   pull_request_target:
-    types: [opened, synchronize, reopened, ready_for_review, edited]
+    # Never `edited`: it fires on every title or description rewrite -- and
+    # coderabbit rewrites the description of each pull request several times.
+    # Every one of those queued a run that cancelled the labelling run still in
+    # flight and then skipped itself, so a push could end up unlabelled behind a
+    # pile of skipped checks. Retargeting a pull request changes its diff without
+    # a push: relabel it by hand with the dispatch below.
+    types: [opened, synchronize, reopened, ready_for_review]
   workflow_dispatch:
     inputs:
       pr:
@@ -32,7 +38,6 @@ jobs:
     name: planet labels
     runs-on: ubuntu-latest
     timeout-minutes: 5
-    if: ${{ github.event.action != 'edited' || github.event.changes.base }}
     steps:
       - name: label the pull request by size
         shell: bash
--- a/.github/workflows/labels.yml
+++ b/.github/workflows/labels.yml
@@ -2,12 +2,6 @@ name: Planet labels
 
 on:
   pull_request_target:
-    # Never `edited`: it fires on every title or description rewrite -- and
-    # coderabbit rewrites the description of each pull request several times.
-    # Every one of those queued a run that cancelled the labelling run still in
-    # flight and then skipped itself, so a push could end up unlabelled behind a
-    # pile of skipped checks. Retargeting a pull request changes its diff without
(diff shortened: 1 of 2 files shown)


Mutugiii (approved):

dogi:
**Review — rating: 84/100**

The change is correct and it's worth more than "less noise". Dropping `edited` from the trigger types removes the job-level guard's reason to exist, but the guard was never the whole story: `concurrency` is declared at workflow level with `cancel-in-progress: true`, so an `edited` event still *queued a run* that entered the group, cancelled whatever labelling run was in flight, and then skipped its own job.

That isn't hypothetical — it's visible in the current check runs on #10375, where two `planet labels` runs landed three seconds apart:

```
102466334818  planet labels  cancelled  12:33:31
102466340421  planet labels  success    12:33:34
```

and skipped `planet labels` entries are scattered across #10371, #10379, #10387 and this PR. So this is a correctness fix for the labelling workflow, not just cleanup.

The cost is that retargeting a PR's base no longer triggers a relabel, which is what `github.event.changes.base` covered. That's rare and `workflow_dispatch` handles it, so it's the right trade.

**Two reasons this isn't ready to merge as-is:**

1. **The approval is stale, and the post-approval commit is a regression.** mutugi approved `756ea447`; `df50d9e6 Clean up comments in labels.yml workflow` landed after and deletes the six-line comment that documented all of the above — the coderabbit description rewrites, the cancelled-then-skipped runs, and the dispatch workaround for retargeting. That comment was the most valuable part of the diff for whoever next wonders why `edited` is missing from a labelling workflow. Please restore it.

2. Once restored, the PR needs re-approval at the new head.

CI green, merges clean. Moving to `review`.

---

PR #10393, opened 2026-09-04
manager: smoother reports date filtering (fixes #10401)

### LINKED ISSUE #10401: create a new 3 months option for the manager reports
This could serve as the default to avoid the lag caused when loading the 12-month timeframe, which serves as the current default

Mutugiii:
Fixes #10401

## Summary
Updates the default time filter in the reports dashboard from 12 months to 3 months, providing a more focused view of recent activity by default.

image

## Changes
- **ReportsDetailComponent**: 
  - Changed `selectedTimeFilter` default from `'12m'` to `'3m'`
  - Updated default start date calculation from 12 months back to 3 months back in login activities initialization
  - Updated custom date filter fallback from 12 months to 3 months
  - Updated reset filter method to use `'3m'` instead of `'12m'`

- **ReportsService**:
  - Added new `'3m'` (Last 3 Months) time filter option to the available filters list
  - Implemented date range calculation for the `'3m'` filter case in the date range logic

https://claude.ai/code/session_014jNNm16a6nqjk7uNiMq8c1

Additionally, addressed the month-end overflow reported here and tightened the related report date handling:
- Added clamped subtraction for all month filters (1m, 3m, 6m, and 12m). Dates such as May 31 now correctly resolve to February 28 instead of overflowing into March.
- Month ranges now begin at the start of their boundary day, preventing results from changing based on when the report is opened. The rolling behavior for 24h and 7d remains unchanged.
- Switching to Custom now preserves the displayed range instead of resetting it to a fixed default.
- Report reloads now follow the selected time filter instead of applying a hardcoded three-month range.
- Removed the reports/CSV utility dependency cycle by moving myPlanet export handling into CsvService. This also removes the constructor-time export helper initialization from both myPlanet components.

## New Features
- Added a “Last 3 Months” date filter to reports.
- Reports now default to displaying the past three months.
- Custom date selection starts with the current default reporting range.
- Improved CSV exports for single and multiple Planet reports.

## Bug Fixes
- Clearing report filters restores the three-month date range.
- Report date ranges now handle month-end and leap-day dates correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "planet",
   "license": "AGPL-3.0",
-  "version": "0.24.60",
+  "version": "0.24.61",
   "myplanet": {
     "latest": "v0.70.45",
     "min": "v0.67.67"
--- a/src/app/manager-dashboard/reports/myplanet/logs-myplanet.component.ts
+++ b/src/app/manager-dashboard/reports/myplanet/logs-myplanet.component.ts
@@ -148,8 +148,8 @@ export class LogsMyPlanetComponent extends MyPlanetFiltersBase implements OnInit
     this.isEmpty = areNoChildren(this.apklogs);
   }
 
-  private mapToCsvData(children: any[], planetName?: string): any[] {
-    return children.map((data: any) => ({
+  private mapToCsvData = (children: any[], planetName?: string): any[] =>
+    children.map((data: any) => ({
       ...(planetName ? { [$localize`Planet Name`]: planetName } : {}),
       [$localize`ID`]: data.androidId,
       [$localize`Name`]: data.deviceName || data.customDeviceName,
@@ -158,7 +158,6 @@ export class LogsMyPlanetComponent extends MyPlanetFiltersBase implements OnInit
       [$localize`Version`]: data.version,
       [$localize`Error`]:  data.error || $localize`N/A`,
     }));
-  }
 
   exportAll(): void {
     this.csvService.exportMyPlanet(this.apklogs, undefined, this.mapToCsvData, $localize`myPlanet Logs`);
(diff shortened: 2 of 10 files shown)


RyanS4 (approved):

dogi:
**Review — rating: 86/100**

Clean work. `subtractMonthsClamped` is correct — pinning the day to 1 before `setMonth` avoids JS's month overflow, and clamping to the target month's last day gives May 31 − 1 month = Apr 30 rather than the native `setMonth` result of May 1. Combining it with `startOfDay` for the month ranges while leaving `24h`/`7d` rolling is a sensible line to draw, and the comment explains it.

Both CodeRabbit findings are genuinely addressed:
- `mapToCsvData` is now an arrow property, so `this.localeId` survives being passed as a callback into `exportMyPlanet`.
- The clamped subtraction is shared across `getDateRange` and the three-month fallbacks rather than duplicated.

Moving `exportMyPlanetCsv` out of `reports.utils.ts` and onto `CsvService.exportMyPlanet` is also an improvement beyond the ask — it drops the curried helper and the `private exportCsvHelper = exportMyPlanetCsv(this.csvService)` field initializer that depended on constructor-injection ordering, and it lets `reports.utils.ts` stop importing `CsvService` entirely.

**Two things block `merge`:**

1. **No approval from mutugi.** The only approval is from RyanS4.
2. **Conflicts with #10371.** Both rewrite `reports-detail.component.ts` and both independently create `reports-detail.component.spec.ts`:

```
CONFLICT (add/add): src/app/manager-dashboard/reports/reports-detail.component.spec.ts
CONFLICT (content): src/app/manager-dashboard/reports/reports-detail.component.ts
```

The spec needs a hand-merged union of both files. This branch is at master head and #10371 is 17 behind, so landing this one first and having #10371 merge master in is the cheaper order — but that's a call for whoever owns both.

CI green. Moving to `review`.

---

dogi (approved):

PR #10394, opened 2026-09-04
all: smoother tables page selecting (fixes #10411)

### LINKED ISSUE #10411: select-all checkbox consistency
Across planet, the select-all checkbox is inconsistent, with some cases selecting all the data in the dataset and, in other cases, only the items on the page. We need to standardize the select-all implementation across planet tables.

Mutugiii:
Fixes #10411 

## Summary
Centralizes current-page table selection behavior in shared helpers while preserving each table’s existing selection semantics.

## Changes
- Adds shared helpers for checking and toggling visible-row selections.
- Keeps user selections across pages in the users table, supporting:
  - multi-page bulk role assignment
  - add-users dialogs
- Preserves clear-all behavior when deselecting in:
  - courses
  - resources
  - meetups
  - surveys
  - manager fetch
- Makes the dialogs-list action explicit with “Select page” and “Deselect page” labels.
- Keeps dialog selections accumulated across pages.
- Prevents missing selected names from removing the wrong tooltip entry.
- Adds regression coverage for cross-page selection and both deselection policies.

https://claude.ai/code/session_01EFkmrrJxcH4BPA9NqvMmUx

## Summary by CodeRabbit

- **New Features**
  - Select-all controls now select or deselect rows on the current page.
  - Selections remain preserved when moving between pages.
  - Selection behavior is consistent across courses, meetups, resources, surveys, users, dialogs, and manager views.
  - Rows that cannot be selected are excluded from select-all actions.

- **Bug Fixes**
  - Improved selection indicators and tooltips to accurately reflect the current page.
  - Prevented unrelated selections on other pages from affecting the current page’s selection status.
  - Removed selections for rows hidden by filters and reset pagination when filters change.


dogi:
**Review — rating: 87/100**

Promoting to `merge`. Gate checked at head `db2fb933`: CI green on that exact commit (suite started 2026-09-09T21:42:45Z vs head committed 21:42:14Z). One `planet labels` run is `skipped` — workflow noise, and a second `planet labels` run on the same head is `success`. `git merge-tree --write-tree origin/master pr/10394` exits 0. 13 commits behind master but merges clean.

**The helper preserves the semantics it found rather than flattening them**, which is the thing I most expected a refactor like this to break. `users-table` deselected only the visible rows; courses/resources/meetups/surveys called `selection.clear()`. Both survive via `clearAllOnDeselect`, and the four that clear pass `true` while `users-table` correctly doesn't. `surveys` keeps its `isSelectable` predicate through the options object. I checked all six call sites against their originals — no semantic drift.

`isAllVisibleSelected`'s `hasSelectable` flag is the right detail: a page of entirely unselectable rows returns `false` rather than vacuously `true`, matching the old `selectableRowsInPage.length > 0 && ...`.

**Several real fixes ride along, not just deduplication:**

- `usersTable.connect().subscribe(...)` was never unsubscribed; it now takes `takeUntil(this.onDestroy$)`. Genuine leak fix.
- The tooltip rewrite kills a real bug. The old `removeFromSelectedNames` did `splice(indexOf(name), 1)`, so two selected users sharing a display name meant deselecting one removed the other's entry — and `indexOf` returning `-1` spliced the last element off. `[ ...new Set(...) ]` is both simpler and correct.
- Stale `initialSelection` ids are dropped up front. Previously they survived into `selectedRows()` as `undefined` holes, since `data.find(...)` returned nothing for them.
- `resetPage()` on filter change — filtering while on page 3 used to leave you staring at an empty page.

**The one thing reviewers should consciously sign off on** is a behaviour change, not a bug: `dialogs-list` select-all now operates on `renderedRows` (current page) where it previously used `tableData.filteredData` (all filtered rows, every page). The button relabel is honest about it —

```
{isAllSelected(), select, yes {Deselect page} no {Select page}}
```

— but anyone who used these dialogs to select a large filtered set in one click now clicks per page. That reads as the intent of #10411, and it makes `dialogs-list` consistent with the six table components; just flagging it as the user-visible edge of this PR.

I also confirmed `selectedElements` and `selectedNames` are gone with no remaining references anywhere in `src/`, and `tooltipText` is still bound and still maintained.

---

dogi (approved):

PR #10397, opened 2026-09-04
ci: add line ending consistency check to pre-push hook

Mutugiii:
## Summary

Adds a new line ending validation script that prevents files from mixing LF and CRLF line endings within a single file. This check runs before linting in the pre-push hook and in CI, catching mixed line endings early without blocking contributors' choice of line ending style per file.

## Changes

- **New script `scripts/check-line-endings.mjs`**: Validates that files use consistent line endings (all LF or all CRLF, never both). The script:
  - Inspects only files changed in the current branch by default (configurable via `--base <ref>` or `--all`)
  - Handles both tracked files (via `git ls-files --eol`) and untracked files (by reading content directly)
  - Reports mixed files with counts of LF and CRLF occurrences
  - Provides remediation guidance (e.g., `perl` one-liner to normalize)
  - Supports environment variable `EOL_BASE_REF` for CI integration

- **Updated `git-hooks/pre-push`**: Runs `npm run lint:eol` before the main linters, failing fast if line endings are mixed

- **Updated `package.json`**: Added `lint:eol` npm script pointing to the new check

- **Updated `.github/workflows/planet.yml`**: 
  - Sets `fetch-depth: 0` on checkout to enable diff against `master`
  - Runs line ending check before `npm ci` (no node_modules needed, fails in seconds)

- **Updated `AGENTS.md`**: Documents that the pre-push hook now runs the line ending check first

## Implementation notes

- The check respects `.editorconfig` and contributor git configuration—it only rejects *mixing* styles within a single file
- Mixed files already in the repository are left untouched until edited
- Untracked files are checked by reading content directly since `git ls-files --eol` doesn't cover them
- Submodules and symlinks are skipped (they report no end-of-line style)

https://claude.ai/code/session_01TAjxP5oqDSX8cQ2UFLeHmM

### DIFF
--- a/git-hooks/pre-push
+++ b/git-hooks/pre-push
@@ -39,6 +39,11 @@ run_lint() {
   tslint=$(npm run lint)
 }
 
+echo "Checking line endings of changed files"
+eol=$(npm run --silent lint:eol 2>&1)
+ret_code_eol=$?
+output_result $ret_code_eol "line endings" "$eol"
+
 echo "Running planet linters"
 run_lint .
 
--- a/scripts/check-line-endings.mjs
+++ b/scripts/check-line-endings.mjs
@@ -1,0 +1,253 @@
+#!/usr/bin/env node
+
+/*
+ * Rejects files that mix LF and CRLF line endings.
+ *
+ * A file may be entirely LF or entirely CRLF - that choice is left to
+ * .editorconfig and to each contributor's git configuration. What this check
+ * refuses is a single file carrying both, because that is what turns a small
+ * edit into a whole-file diff.
+ *
+ * Only files touched by the current branch are inspected, so the mixed files
+ * already in the repository stay as they are until someone edits them.
+ *
+ * Usage:
+ *   node scripts/check-line-endings.mjs                # changed files vs. master
+ *   node scripts/check-line-endings.mjs --base <ref>   # changed files vs. <ref>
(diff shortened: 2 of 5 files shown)


git-hooks/pre-push
```diff
@@ -39,6 +39,11 @@ run_lint() {
   tslint=$(npm run lint)
 }
 
+echo "Checking line endings of changed files"
```
paulbert, line 42: Should test with a Windows machine set to use CRLF locally and push with LF. Might block legitimate pushes depending on when that adjustment is made.

scripts/check-line-endings.mjs
```diff
@@ -0,0 +1,253 @@
+#!/usr/bin/env node
+
+/*
+ * Rejects files that mix LF and CRLF line endings.
+ *
+ * A file may be entirely LF or entirely CRLF - that choice is left to
```
paulbert, line 6: While I think this makes some sense, I think we should shoot for using one line ending type across all files.

Leaving it up to each contributor ignores the fact that we have many contributors over a long period of time and files could switch back and forth with some less than helpful diffs.

paulbert (commented):
For context, there are 24 files with mixes line endings and 25 with CRLF. I think modifying this to only accept LF and then converting those files to LF in a follow up would be a better plan.

PR #10399, opened 2026-09-04
manager: smoother reports demographics exporting (fixes #10413)

### LINKED ISSUE #10413: Improve demographic handling in report and submission exports
## Problem

Reports containing individual user or patient rows do not consistently include age and gender demographics. Existing demographic handling also has several edge cases:

- Child-planet users may be wrapped in a `doc` property and fail profile lookup.
- Birthday calculations compare weekdays instead of calendar dates.
- Missing or invalid birth dates can produce `NaN`.
- Health examinations can lose demographics stored directly on the record.
- Age `0` can be mistaken for a missing value.
- Unexpected gender values can interrupt CSV generation.
- Repeated per-row user searches become inefficient for large exports.

## Expected behavior

The following user-level exports should include age and gender:

- Login Report
- Resources View Report
- Courses View Report
- Course Progress Report
- Health Report
- Chat Report

Demographics should be resolved from the associated user profile, except Health Report records, which should preserve their stored patient demographics.

Aggregate exports—Resources Overview, Courses Overview, and Summary—are outside this scope.

Mutugiii:
Fixes #10413

## Summary

Adds age and gender demographics to report CSV exports containing individual user or patient rows, while improving demographic lookup, formatting, and fallback behavior.

## Changes

- Adds age and localized gender to:
  - Login Report
  - Resources View Report
  - Courses View Report
  - Course Progress Report
  - Chat Report
- Preserves and localizes the demographics stored directly on Health Report examination records.
- Leaves aggregate Resources Overview, Courses Overview, and Summary exports unchanged.
- Resolves users from both local profiles and child-planet documents.
- Indexes users by name once during demographic enrichment.
- Keeps raw gender values inside the reports service for chart grouping and localizes them at CSV boundaries.
- Prevents malformed gender values from breaking CSV generation.
- Corrects birthday calculations to compare calendar dates rather than weekdays.
- Supports dates, ISO strings, and numeric timestamps, including timestamp `0`.
- Returns safe fallbacks for missing or invalid birth dates instead of `NaN`.
- Preserves age `0` while displaying `N/A` for blank or missing submission ages.
- Adds focused tests for demographic enrichment, gender formatting, age calculation, and submission export fallbacks.
- Normalizes line endings in `reports-detail.component.ts` in a separate commit.

https://claude.ai/code/session_012RPeuWgSojRNqiUXXz4mxB

## Summary by CodeRabbit

- **New Features**
  - CSV exports for reports and chat data now include user gender and age demographics.
  - Demographic headers and gender values are localized for clearer, more consistent exports.
  - Age information uses birth dates when available, with stored-age fallbacks.

- **Bug Fixes**
  - Corrected age calculations around birthdays and invalid or missing dates.
  - Preserved valid zero-age values instead of treating them as missing.
  - Improved demographic handling when user details are unavailable.
  - Improved report sorting for dates, numbers, missing values, and older records.


dogi:
**Review — rating: 86/100**

Staying `ready` — the code is sound and CI is green at head `a0ad0f33`, but this no longer merges. Dropping the duplicate `review` label.

**The `ageFromBirthDate` fix alone justifies this PR.** The old comparison used the wrong accessor:

```ts
now.getMonth() === birth.getMonth() && now.getDay() < birth.getDay()
```

`getDay()` is the day of the *week* (0–6), not the day of the month. So for anyone whose birthday fell in the current month, "has their birthday passed yet" was decided by comparing two weekday numbers — wrong roughly six days in seven, and off by a full year when it landed wrong. `getDate()` is correct, and folding the nested ternaries into one boolean makes it checkable by eye.

`calendarDate` is a real fix too, not defensive noise: `new Date('1990-05-15')` parses as UTC midnight, which is the previous day in any negative-offset timezone, so ages were off by one for birthdays near month boundaries in the Americas. Parsing `YYYY-MM-DD` into a local-time date avoids that, and the round-trip check

```ts
return calendar.getMonth() === month - 1 && calendar.getDate() === day ? calendar : new Date(NaN);
```

catches values like `2021-02-30` that `Date` would silently roll forward to March. Returning `null` for missing/invalid input rather than `NaN` gives callers something they can actually branch on.

`genderBucket` is used (`reports.service.ts:119`) and covered by tests including the `'Male'`/`'FEMALE'` casing and empty-string cases, so it isn't dead weight.

**The blocker: this conflicts with master in four files.**

```
CONFLICT (add/add): src/app/manager-dashboard/reports/reports-detail.component.spec.ts
CONFLICT (content): src/app/manager-dashboard/reports/reports-detail.component.ts
CONFLICT (add/add): src/app/manager-dashboard/reports/reports.service.spec.ts
CONFLICT (content): src/app/manager-dashboard/reports/reports.utils.ts
```

13 commits behind, and two of those commits landed squarely in this area — `0b288abb` (reports date filtering, #10393) and `00e04d5b` (myplanet app source reporting, #10371). The two **add/add** spec conflicts are the awkward ones: master and this branch each created those spec files independently, so there is no common ancestor to merge against and git can only hand you both versions whole. Those need a hand-merged union — keeping only one side will silently drop the other's coverage.

**One thing to sequence around while you're rebasing.** [#10437](https://github.com/open-learning-exchange/planet/pull/10437) deletes `converter` and `markdownToPlainText` from `shared/utils.ts` entirely, moving that work into a new `MarkdownRenderService`. This branch still imports both:

```ts
import { ageFromUser, converter, dedupeShelfReduce, localizedGender, markdownToPlainText, toProperCase } from '../shared/utils';
```

[#10437](https://github.com/open-learning-exchange/planet/pull/10437) merges clean today and is `merge`-labelled, so it will likely land first — at which point that import breaks the build. No textual conflict check catches this, which is why I'm flagging it here. Rebase onto master after [#10437](https://github.com/open-learning-exchange/planet/pull/10437) lands and migrate those two call sites to the new service in the same pass.

Re-run the gate once rebased and this should promote.

---

dogi:
**Re-review — rating: 88/100** (was 86)

Re-audit triggered by the head moving: `a0ad0f33` → `f8482dad`, and the label going to `merge` after that push. Re-ran the full gate at the new head — it passes, so `merge` stands. **But this cannot merge on the same day as [#10437](https://github.com/open-learning-exchange/planet/pull/10437); see the bottom.**

Gate at `f8482dad`: CI green on that exact commit (22 runs `success`, suite started 2026-09-17T10:06:27Z vs head committed 10:06:01Z — 26 seconds, but a real result); `git merge-tree --write-tree origin/master pr/10399` exits 0; 0 commits behind master.

**The add/add spec resolution is correct, and I checked it rather than assuming.** Both spec files were the ones I flagged as having no common ancestor, where keeping one side silently drops the other's coverage. Comparing test names across master, the old branch head and the merge:

```
reports.service.spec.ts          master 5 + branch 19 → merged 24, nothing missing
reports-detail.component.spec.ts master 6 + branch  8 → merged 15
```

The one master name that doesn't appear in the merged file — `replaces the raw app value with a localized source label in login exports` — was folded rather than dropped. It's now `exports a localized source and user demographics instead of the raw app value for logins`, and it keeps both original assertions:

```ts
expect(exportedRow.app).toBeUndefined();
expect(exportedRow[$localize`Source`]).toBe('myPlanet Lite');
expect(exportedRow[$localize`Gender`]).toBe('Female');
expect(exportedRow[$localize`Age (years)`]).toBe(28);
```

A strict superset. That's the right way to resolve an add/add on a spec, and it's why `reports-detail.component.ts` shrank from 225 diff lines to 73 — master's `appSourceLabel` work landed independently and the demographics layer now wraps it instead of duplicating it.

**The rebase also brought a fix that wasn't in the version I reviewed.** `sortData` used to be:

```ts
comparison = a[field].localeCompare(b[field]);
```

which throws outright on any record missing that field, and sorted numbers lexicographically. Now `createdDate` joins the date branch, numbers get a numeric comparison, and the fallback is `` `${valueA ?? ''}`.localeCompare(`${valueB ?? ''}`) ``. The new spec covers both cases directly (`sorts on a field the older records never stored`, `keeps ordering dates by their value, not their text`). That's worth the two extra points.

Minor: `.eslintrc.json` picked up a trailing-whitespace-only change on one blank line. No behavioural effect — and with `eslint.config.mjs` present and `"lint": "ng lint planet-app"`, that file looks vestigial for the root app anyway — but it's unrelated to this PR and worth dropping to keep the diff honest.

---

**Blocking for the merge queue: this and [#10437](https://github.com/open-learning-exchange/planet/pull/10437) are both labelled `merge`, both green, both 0 behind — and they cannot both land.**

```
$ git merge-tree --write-tree pr/10399 pr/10437
CONFLICT (content): src/app/submissions/submissions.service.ts
CONFLICT (content): src/app/submissions/submissions.service.spec.ts
```

[#10437](https://github.com/open-learning-exchange/planet/pull/10437) deletes `converter` and `markdownToPlainText` from `shared/utils.ts` (both present in master, both gone at its head). This branch still calls them in five places in `submissions.service.ts` — lines 349, 401, 463, 475 and 605. Each PR merges clean against master on its own, so nothing in the per-PR gate catches this; it only shows up when you check them against each other.

**Merge this one first, then rebase [#10437](https://github.com/open-learning-exchange/planet/pull/10437).** This reverses what I recommended yesterday, because the facts changed: back then this branch was 13 commits behind with four conflicts and [#10437](https://github.com/open-learning-exchange/planet/pull/10437) was the clean one, so [#10437](https://github.com/open-learning-exchange/planet/pull/10437) first was cheaper. Now both are current, and the asymmetry is in who owns the migration — [#10437](https://github.com/open-learning-exchange/planet/pull/10437) has already rewritten all five of those call sites against `MarkdownRenderService`, so its author re-applies work they've done, on code they wrote. The other order asks this author to learn a service they didn't write and migrate five call sites during a rebase.

---

dogi (approved):

PR #10404, opened 2026-09-08
all: smoother member delimiter linting (connects #9082)

paulbert:
Adding back in rules for member delimeters and space before function.

For member delimeter I went with comma for one line and semicolon for multi line type definitions. This matches more closely to our general style across the app currently.

In addition:
- The commented no-host-metadata-property is removed because that rule no longer exists.
- Uncommented naming-convention and object-shorthand rules from the old linting file because they were already added back into the new file.

connects #9082 

## Summary by CodeRabbit

* **Chores**
  * Standardized TypeScript and function formatting across the codebase.
  * Enabled ESLint rules for naming conventions, object shorthand, function spacing, and member delimiters.
  * Updated type declarations and method signatures to follow consistent delimiter and spacing conventions.
  * No user-facing functionality or behavior changed.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -75,7 +75,6 @@
         "@angular-eslint/contextual-lifecycle": "error",
         "@angular-eslint/directive-class-suffix": "error",
         "@angular-eslint/no-conflicting-lifecycle": "error",
-        // "@angular-eslint/no-host-metadata-property": "error",
         "@angular-eslint/no-input-rename": "error",
         "@angular-eslint/no-inputs-metadata-property": "error",
         "@angular-eslint/no-output-native": "error",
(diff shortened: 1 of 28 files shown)


Mutugiii (approved):

dogi:
**Review — rating: 62/100**

**`.eslintrc.json` is invalid JSON on this branch.** Uncommenting the `@typescript-eslint/naming-convention` block left a trailing comma:

```json
        "@typescript-eslint/naming-convention": [
          "error",
        
          // Not transferring over the default case, ...
          {
            "selector": "default",
            ...
          },

        ],
```

The `},` before `]` has no following element. Strip the comments and parse and it fails at line 100; master's copy of the same file parses fine. There's also a line of trailing whitespace after `"error",`.

CI doesn't catch this because **nothing reads that file.** `npm run lint` is `ng lint planet-app`, the `@angular-eslint/builder:lint` target carries no `eslintConfig` option, and `eslint.config.mjs` is present at the root — so ESLint 9 uses flat config and never loads `.eslintrc.json`. Per `AGENTS.md` the legacy config is the gateway's; the root app is flat.

So the `.eslintrc.json` half of this diff is inert *and* broken, and it contradicts the half that isn't: the legacy file sets `member-delimiter-style` `singleline: "semi"` while `eslint.config.mjs` sets `singleline: "comma"` — and the comma form is what the 25 source files were just rewritten to:

```ts
-  team: { _id: string; name: string; type: string };
+  team: { _id: string, name: string, type: string };
```

Anyone who later flips `ESLINT_USE_FLAT_CONFIG=false`, or copies the legacy file forward, gets a parse error and then a rule that fights the codebase.

The `eslint.config.mjs` changes and the source rewrites are fine on their own — consistent with each other and mechanical.

**To unblock:** either drop `.eslintrc.json` from this PR entirely (leave it as master has it), or, better, delete the file — it's dead weight at the root and having two disagreeing configs is how this happened. If it stays, fix the trailing comma and match `singleline` to the flat config.

mutugi approved at the current head `f9c18264` and CI is green, but green here only means the broken file is never loaded. Moving to `change`.

---

dogi (approved):

PR #10431, opened 2026-09-10
dashboard: smoother courses cards dragging (fixes #10429)

### LINKED ISSUE #10429: Dashboard: courses with images overflow on card move
Card UI glitches when trying to move a dashboard course card with an image attached. 

## Card on idle state

Image

## Card while dragging across row

Image

RyanS4:
Fixes #10429

### Description

Dragging a dashboard course card with a cover image creates an Angular CDK preview under `<body>`. The original card styles were nested beneath the component host, so they did not match
the hoisted preview. This allowed cover images to render at their intrinsic size and overflow the card.

### Changes

- Moved the shared dashboard-item layout into a top-level, component-scoped rule that matches both the original card and CDK’s cloned preview.
- Constrained course cover images and preserved card text, spacing, and alternating backgrounds while dragging.
- Used Material elevation for the preview instead of a hardcoded shadow.
- Hid the remove button from the preview.
- Added placeholder, sibling movement, and drop-return transitions.
- Let CDK preserve the source card’s measured dimensions instead of forcing preview sizing.
- Removed duplicated preview styles from the global stylesheet.

### Verification

- `npm run lint`
- `npm run build`
- `npx vitest run` — 57 test files and 398 tests passed

### Manual verification

- [ ] Course cover remains constrained while dragging
- [ ] Grey cards remain grey in the preview
- [ ] Preview animates into position when dropped

## Summary by CodeRabbit

- **Style**
  - Improved the appearance and consistency of dashboard items, including links, text truncation, and course covers.
  - Added visual elevation to draggable item previews.
  - Added smoother drag-and-drop transitions and clearer placeholder states.
  - Applied dashboard item styling more consistently across dashboard layouts.
  - Preserved appropriate spacing for empty links and visible focus outlines for improved navigation.

### DIFF
--- a/src/app/dashboard/dashboard-tile.component.spec.ts
+++ b/src/app/dashboard/dashboard-tile.component.spec.ts
@@ -1,5 +1,6 @@
 import { ElementRef, EnvironmentInjector, runInInjectionContext } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
+import { provideRouter } from '@angular/router';
 import { MatDialog } from '@angular/material/dialog';
 import { of } from 'rxjs';
 import { vi } from 'vitest';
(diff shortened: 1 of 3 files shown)


Mutugiii (approved):

dogi:
**Review — rating: 87/100**

Promoting to `merge`. Gate checked at head `0419fbe6`: CI green on that exact commit (22 runs `success`, suite started 2026-09-11T20:03:17Z, and the head commit predates it); `git merge-tree --write-tree origin/master pr/10431` exits 0; 3 commits behind master, merges clean. APPROVED by @Mutugiii at head.

**The hoist is the fix, and it's worth being explicit about why**, because at a glance this looks like 132 lines of SCSS shuffled for no reason. `cdkDrag` renders the dragged element as a clone in an overlay attached to `<body>`, outside the component's host element. Every rule for `.dashboard-item` was nested under

```scss
:host { ... .card-container { ... .dashboard-item { ... } } }
```

which compiles to a selector requiring the host ancestor — so the clone matched none of it and the card lost its layout, padding, cover image and text clamping the moment you picked it up. Moving those rules to a top-level `.dashboard-item` block makes them apply to the detached preview too, since the clone keeps its `_ngcontent` attribute under emulated encapsulation. What stays nested is exactly what should stay nested: `grid-column-end`, `grid-row-end`, `:nth-child(even):last-child` — grid placement that only means anything inside the container.

The rest follows from that: `&.cdk-drag-preview` can now style the clone (white background, rounded, `.delete-item` hidden so you're not dragging a delete button around), `cdkDragPreviewClass="mat-elevation-z8"` gives it the lift, and `.cdk-drop-list-dragging .dashboard-item:not(.cdk-drag-placeholder)` animates the other tiles getting out of the way.

Specificity works out — the hoisted block sits before `:host` and carries lower specificity than the nested rules, and the two sets don't declare overlapping properties, so nothing is silently overridden.

**On CodeRabbit's finding** (hide `.delete-item` inside `.cdk-drag-placeholder` the way `.cdk-drag-preview` does): I checked, and I'd leave it. The placeholder is the ghost of the tile left in the grid, and the whole thing is already at `opacity: 0.3` — the delete button fades with it. Hiding it would make the placeholder a different shape from the tile it's holding space for, which is the opposite of what a placeholder is for. Cosmetic preference, not a defect.

Everything here goes through `v.$primary`, `v.$white`, `v.$grey` and `v.$dashboard-tile-width` rather than literals, which is what `Style-Guide.md` asks for.

---

dogi (approved):

PR #10437, opened 2026-09-11
all: smoother markdown rendering (fixes #10436)

### LINKED ISSUE #10436: sanitize rendered markdown centrally
Planet renders stored Markdown through `PlanetMarkdownComponent`, which delegates to Covalent's `TdMarkdownComponent`. Covalent converts Markdown with Showdown and sanitizes the resulting markup using `SecurityContext.STYLE` before assigning it through `innerHTML`.

Image[/+][-]

Consequently, unsafe Markdown can produce executable HTML. Confirmed payloads include:
```md
[x [y]](javascript:alert(1))
[x \] y](javascript:alert(1))
![x](bad"onerror="alert(1))

These produce a javascript: link or a live onerror attribute after Showdown 2.1.0 rendering and browser HTML parsing.

Sanitizing Markdown source with regular expressions is not sufficient because Showdown accepts nested/escaped labels and destinations that can inject attributes.

There are also inconsistent rendering paths:
- PlanetMarkdownComponent uses Covalent's Showdown configuration.
- dialogs-prompt.component.html uses td-markdown directly.
- utils.ts exports a separately configured Showdown converter used by plain-text
  and PDF exports.
- Tables and other Markdown can therefore render differently on screen and in exports.

## Goal
Provide one safe-by-default Markdown rendering boundary for application views, dialogs, previews, and exports.

## Scope
- Add a centralized Markdown rendering service.
- Define one explicit Showdown configuration, including:
    - strikethrough
    - ghCodeBlocks
    - ghCompatibleHeaderId
    - tasklists
    - tables
    - literalMidWordUnderscores
    - emoji
    - disableForced4SpacesIndentedSublists
    - configurable simpleLineBreaks
- Render Markdown before sanitization.
- Resolve hosted/parent URLs without bypassing URL sanitization.
- Sanitize the final generated HTML using Angular's SecurityContext.HTML.
- Never use bypassSecurityTrustHtml.
- Make safe rendering unconditional; do not expose an untrusted opt-in.
- Replace the direct td-markdown use in DialogsPromptComponent.
- Migrate the shared converter used by plain-text and PDF exports to the centralized configuration.
- Keep @Input() values immutable and use derived render results.
- Add a data-driven Markdown security and compatibility regression suite.

### PR DESCRIPTION
**Pull request by Mutugiii**
Fixes #10436

image

## Summary
Centralizes read-only Markdown rendering in MarkdownRenderService and sanitizes the generated HTML before it reaches the DOM, CSV exports, or PDF generation.

This replaces regex-based source sanitization, which could not reliably account for Showdown’s Markdown grammar. Nested link labels, escaped brackets, multiline reference definitions, and quote-based attribute injection could bypass that approach.

## Changes
- Added one canonical Showdown configuration and rendering service.
- Sanitized rendered HTML with Angular’s SecurityContext.HTML.
- Made planet-markdown the shared display component.
- Removed the unsafe chat rendering directive and direct td-markdown usage.
- Added a chat profile preserving bare URL links and single line breaks.
- Centralized Markdown-to-plain-text conversion for CSV and PDF exports.
- Preserved hosted media URL rebasing and Markdown fragment navigation.
- Preserved table alignment and task-list state across browser and PDF output.
- Added lint restrictions preventing direct Showdown or Covalent Markdown usage outside the renderer.
- Added a data-driven security regression corpus covering nested links, attribute injection, multiline references, unsafe schemes, and raw HTML payloads.

## Behavior decisions
- Safe raw HTML is rendered; unsafe elements and attributes are removed.
- Unsupported link schemes lose their href.
- Raster data:image sources are allowed; SVG data sources are blocked.
- Task-list markers use visible ☐ and ☑ characters so their state survives copying and exports.
- Heading IDs remain canonical for deep links, while in-page fragment handling is scoped to the clicked Markdown instance.

## Summary by CodeRabbit

- **New Features**
  - Added a shared Markdown renderer with sanitized output, safe links and media, code blocks, tables, task lists, emoji, and heading navigation.
  - Added chat-specific Markdown formatting, including line breaks, bare URLs, and code-copy controls.
  - Added consistent Markdown rendering across chats, dialogs, meetups, achievements, and question views.
  - Improved PDF and CSV exports with formatted Markdown content and plain-text conversion.

- **Bug Fixes**
  - Unsafe links, media, scripts, and embedded content are now removed from rendered Markdown.
  - Improved hosted image handling and external-link security attributes.
  - Added responsive styling for code blocks, tables, and long Markdown content.

### CODE REVIEW

**Comment by dogi** [comment]
**Review — rating: 89/100**

Promoting to `merge`. Gate checked at head `51dd3f65`: CI green on that exact commit (22 runs `success`, suite started 2026-09-11T20:03:17Z vs head committed 20:02:49Z); `git merge-tree --write-tree origin/master pr/10437` exits 0; 0 commits behind master. No human approval, which the gate doesn't require.

**This is the version of this change that doesn't introduce an XSS hole**, which is the main thing I was looking for. Replacing `td-markdown` with your own renderer normally ends in `bypassSecurityTrustHtml` so the markdown survives `[innerHTML]`. It doesn't here — `render()` terminates in

```ts
sanitizeHtml(html: string): string {
  return this.sanitizer.sanitize(SecurityContext.HTML, html) || '';
}
```

and `git grep bypassSecurityTrust` across the branch returns only the pre-existing `bypassSecurityTrustResourceUrl` calls for static local icon assets in `app.component.ts`. Nothing in the markdown path bypasses. The ordering is right too: normalization mutates the parsed DOM and sanitization runs last on the result, not before.

The service is also stricter than the sanitizer it delegates to, which is the right instinct. Angular's URL sanitizer permits `data:` (its pattern only excludes `javascript:`); `normalizeMediaUrls` rejects it unless it matches `safeImageDataUrl`, and `normalizeLinks` drops any `href` outside `http/https/mailto/tel` rather than trusting the downstream pass. Cross-origin links get `rel="noopener noreferrer"` — and `target`/`rel` both survive, since they're in Angular's allowed attribute set.

**`markContainersAsMarkdown` is the subtle part and it's handled carefully.** Stashing code spans before rewriting container tags means a `<div>` inside backticks doesn't silently acquire `markdown="1"`, and the placeholder is collision-proofed:

```ts
let placeholder = 'planetmd';
while (text.includes(placeholder)) {
  placeholder += 'x';
}
```

**Behaviour is preserved where it looks like it changed.** In non-preview mode the old template rendered `[content]="content"` while `limitedContent`/`textOnly` were computed and discarded — dead work. The new code renders the normalized content directly and drops the unused `markdownImageRegex` import. Preview mode still renders with no hosted URL. Same output, less machinery.

**All four CodeRabbit findings are stale** — it reviewed `dc4a65db`, head is `51dd3f65`. I checked each at head rather than repeating them:

- `addCopyButtons` idempotency → guarded by `if (!navigator.clipboard || container.querySelector('.code-block-wrap')) { return; }` plus a `clearCopyListeners()` before rebuilding.
- Missing `i18n-alt` → present: `alt="Preview Image" i18n-alt`.
- Chat payload not using the chat profile → `submissions.service.ts:478` reads `render(analysisPayload.chat, '', 'chat')`.
- Literal white in `styles.scss` → the added lines use `v.$white`.

**One cross-PR note, not a defect in this PR.** This deletes `converter` and `markdownToPlainText` from `shared/utils.ts` along with all their callers. [#10399](https://github.com/open-learning-exchange/planet/pull/10399) is still importing both in `submissions.service.ts` and is 13 commits behind, so whichever lands second needs the other's changes reconciled by hand — see my note there. Landing this one first is the cheaper order, since it merges clean today and [#10399](https://github.com/open-learning-exchange/planet/pull/10399) needs a rebase regardless.

---

dogi:
**Correction to my note above — the merge order should be the other way round.**

Nothing has changed on this PR; head is still `51dd3f65`, CI green, `merge` label stands. This is only about ordering against [#10399](https://github.com/open-learning-exchange/planet/pull/10399), which has since merged master in (`f8482dad`) and is now also labelled `merge`.

**Both are now at `merge`, both green, both 0 behind master — and they cannot both land:**

```
$ git merge-tree --write-tree pr/10399 pr/10437
CONFLICT (content): src/app/submissions/submissions.service.ts
CONFLICT (content): src/app/submissions/submissions.service.spec.ts
```

Whichever lands second hits this. Nothing in the per-PR gate sees it, since each merges clean against master on its own.

I said above to land this one first. That was right when [#10399](https://github.com/open-learning-exchange/planet/pull/10399) was 13 commits behind with four unresolved conflicts — it needed a rebase regardless, so it was the natural place to absorb the change. It's current and green now, so the cheaper order flipped: **let [#10399](https://github.com/open-learning-exchange/planet/pull/10399) merge first, then rebase this one.**

The reason is who owns the migration. This PR has already rewritten all five `converter` / `markdownToPlainText` call sites in `submissions.service.ts` against `MarkdownRenderService` — lines 349, 401, 463, 475 and 605 on the other branch. Rebasing this one means re-applying a migration you wrote, over code you already understand. The other order asks [#10399](https://github.com/open-learning-exchange/planet/pull/10399)'s author to learn this service and migrate five call sites mid-rebase, with no test of their own covering the markdown path.

The conflict is mechanical either way; this just puts it in front of the person who can resolve it fastest.

---

dogi (approved):

PR #10439, opened 2026-09-11
all: smoother ratings handling (fixes #10438)

### LINKED ISSUE #10438: extract rating persistence from PlanetRatingComponent
PlanetRatingComponent currently owns database persistence, loading and error handling, and rating aggregate reconciliation alongside its UI responsibilities. This makes the behavior harder to reuse and increases the risk of persistence logic drifting between rating entry points.

Move shared rating persistence into RatingService while preserving existing user-facing behavior.

Mutugiii:
Fixes #10438

## Summary
- centralizes rating save and delete operations in RatingService
- centralizes rating aggregate normalization
- extracts shared rating form types and field definitions
- simplifies PlanetRatingComponent by delegating persistence
- preserves existing rating behavior and regression coverage

## Summary by CodeRabbit

* **New Features**
  * Added a shared rating form with rating and comment fields.
  * Ratings can be saved, updated, and deleted while keeping aggregate scores and rating counts accurate.
  * Rating records support both resource and course content.

* **Bug Fixes**
  * Improved handling of rating errors with visible alerts and consistent loading-state cleanup.
  * Preserved existing comments when updating a rating without providing a new comment.

### DIFF
--- a/src/app/shared/forms/planet-rating.component.spec.ts
+++ b/src/app/shared/forms/planet-rating.component.spec.ts
@@ -17,6 +17,7 @@ const initialRating = () => ({
   userRating: {
     _id: 'r-1',
     _rev: '1-a',
+    item: 'item-1',
     rate: 5,
     comment: 'Original comment',
     user: { name: 'learner', gender: 'female' }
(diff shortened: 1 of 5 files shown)


dogi:
**Review — rating: 87/100**

Promoting to `merge`. Gate checked at head `5ba020da`: CI green on that exact commit (22 runs `success`, suite started 2026-09-11T19:51:16Z vs head committed 19:50:27Z); `git merge-tree --write-tree origin/master pr/10439` exits 0; 2 commits behind master, merges clean. No human approval, which the gate doesn't require.

**The component drops from 7 injected dependencies to 4** and stops knowing that ratings live in a `ratings` database. `CouchService`, `DialogsLoadingService` and `StateService` all leave `planet-rating.component`, which is the right direction — a presentational rating widget shouldn't be assembling `createdOn`/`parentCode` from configuration.

**Two details here are easy to get wrong and are both right.**

The loading indicator starts inside `defer`, with the comment explaining why:

```ts
private withLoading<T>() {
  return (source: Observable<T>): Observable<T> => defer(() => {
    this.dialogsLoadingService.start();
    return source.pipe(finalize(() => this.dialogsLoadingService.stop()));
  });
}
```

Start at pipe-construction time instead of subscribe time is the standard way this leaks a spinner that never stops; deferring keeps `start`/`stop` paired per subscription.

The destructuring default is the other one:

```ts
{ item, type, rate, existingRating = {}, comment = existingRating.comment ?? '', ratingInfo = {} }
```

`existingRating` is bound before `comment` references it, so this evaluates correctly — and the comment above it says exactly why it has to be in that order. It matches the old `...this.rating.userRating, ...form.value` spread semantics, where rating via the stars alone preserved the stored comment rather than blanking it; the component's `'comment' in value ? ... : this.rating.userRating.comment || ''` keeps the same behaviour from the other side. `createdTime: existingRating.createdTime || this.couchService.datePlaceholder` likewise reproduces what the old spread ordering did by accident.

Removing `showAlert` from the component's `ratingError()` while the service's `ratingError()` adds it means one alert, not two or zero. Checked that `planetMessage` is still used for six other messages in the component, so it's not an orphaned injection.

**I checked the default change and it's inert.** `@Input() ratingType` goes from `''` to `'resource'`, which would silently retype ratings saved by any call site that omitted it. All four `planet-rating` usages pass it explicitly:

```
courses.component.html:302                [ratingType]="'course'"
courses-view-detail.component.html:4      [ratingType]="'course'"
resources.component.html:308              [ratingType]="'resource'"
resources-view.component.html:80          [ratingType]="'resource'"
```

So nothing in-tree changes. Typing it as `RatingType` means a future omission is a compile error rather than a `''` written to the database.

**Two small things, neither blocking.** `itemId()` prefers `item._id` over `item.doc._id`, while `itemDoc` prefers `item.doc` over `item` — so a wrapped item takes its id from the wrapper and its title from the doc. Harmless given current call sites pass one shape or the other consistently, but the asymmetry will confuse someone eventually. And CodeRabbit's note that the `saveRating` spec mock doesn't mirror the service's append-vs-replace logic is a fair test-fidelity point (raised against `7cc9ecdf`, not head) — worth a follow-up, since a mock that can't distinguish those two paths can't catch a regression in them.

---

dogi (approved):

PR #10447, opened 2026-09-17
all: smoother parent session logout handling (fixes #10378)

### LINKED ISSUE #10378: logout: takes 1 minute
https://github.com/user-attachments/assets/d7061616-4b06-4a57-9d7e-28e3f8f919e3

Mutugiii:
## Summary

Logout currently waits for the parent CouchDB `_session` deletion before clearing the local user and navigating. When the parent is unreachable, this can block logout for about a minute.

- completes the local logout without waiting for parent-session cleanup
- runs the parent deletion as best-effort cleanup and logs failures
- preserves the configured-admin check and skips planets without a parent
- reads the user identity before `unset()`
- safely handles missing configuration or user names

## Testing
- `npm run lint`
- `npx vitest run src/app` — 65 files, 452 tests passed
- `git diff --check`

Added coverage confirms that logout:
- proceeds while the parent deletion remains pending
- proceeds when the parent deletion fails
- skips parent cleanup for non-admin users
- skips parent cleanup on a center and navigates to `/login`
- proceeds when configuration and user names are absent

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved logout handling for local administrators with a configured parent domain.
  - Parent-session cleanup now runs only when applicable and no longer prevents local logout or navigation if it fails.
  - Improved handling when administrator names or configuration values are unavailable.
  - Added coverage for logout, session cleanup, navigation, and error scenarios.


dogi:
**Review — rating: 90/100**

Promoting `review` → `merge`. Gate at head `fb6f13fc`:
- **CI:** 22/22 check runs succeeded. All started 2026-09-17T11:42Z, after the head commit at 11:41:49Z.
- **Merge:** `git merge-tree --write-tree origin/master pr/10447` exits 0. The branch is 12 commits behind but none of them touch `home.component.ts`. It also stacks cleanly with the other nine PRs in this batch.
- **Approval:** none from Mutugiii (not required).

The fix does what the title says. Ending the parent `_session` is moved out of the `switchMap` chain, so local logout no longer waits on the parent:
```ts
if (isLocalAdmin && configuration.parentDomain) {
  this.couchService.delete('_session', { withCredentials: true, domain: configuration.parentDomain })
    .subscribe({ error: error => console.error('Unable to end parent session', error) });
}
```
`isLocalAdmin` is now captured before `pouchAuthService.logout()` and guarded with `adminName?.`. Before, `adminName.split` would throw on a planet with no `adminName`. The new `parentDomain` guard also stops the old fallback, where `domain: undefined` sent the DELETE to the local `_session` a second time. The request still runs to completion in the background, and a 403 would still show the "not authorized" alert through `couchDBReq`. That only reaches local admins, and it was already the behaviour before this change.

Checked and dropped: navigating away does not cancel the in-flight `HttpClient` request, so the parent logout still gets sent.

---

dogi (approved):

dogi (approved):

PR #10452, opened 2026-09-17
all: smoother mobile viewport fitting (fixes #8881)

### LINKED ISSUE #8881: Problem with the movil version
in this screen we lost the 20% of the screen only for show the menu, maybe we can hide all the menu
Image

in horizontal mode with the tittles we lost the 50% of the screen
Image

we have problems with the text render on the movil version

Image

Mutugiii:
Fixes #8881

Full-height views currently use `100vh - 216px` to estimate the space below the app header and page toolbars. That estimate fits some desktop pages, but an expanded filter row can make a list taller than the available space and create an outer scrollbar. On shorter pages it can leave unused space.

A landscape phone also has enough width to be classified as a tablet, even though it has little vertical space. That can leave it with a pinned sidebar and tall toolbar rows.

## What changed
- Size routed full-height views from their available space. Toolbars keep their natural height, and the view fills the remainder. The shared rule covers direct views, views in `.space-container`, and lists inside `planet-users-table`.
- Keep a 40vh minimum for the view. When toolbar rows and that minimum cannot fit together, the page scrolls rather than squeezing or overlapping the list.
- Treat short touch viewports (at most 500px high) as compact in the app shell: use an overlay sidebar, the modern toolbar, and 48px toolbar rows. The sidebar remains scrollable.
- Read width and short-screen state together through `DeviceInfoService.watchViewport()`. Recalculate content margins when the sidebar mode changes so rotation does not leave a blank strip.
- Use `100dvh` where supported, with a `100vh` fallback. Let the course detail view fill its parent instead of using another viewport-height calculation.

The existing fixed-height variables and page corrections remain as fallbacks for embedded views and browsers without `:has()`.

## Summary by CodeRabbit

- **New Features**
  - Improved navigation on short landscape phone screens with overlay navigation and better scrolling.
  - Added responsive support for modern viewport heights and mobile browser interfaces.
  - Added flexible full-height layouts for routed views and course content.

- **Bug Fixes**
  - Improved sizing across home, course, dashboard, and full-height views.
  - Prevented content clipping or misalignment when switching navigation layouts or viewport sizes.
  - Ensured navigation and content margins update correctly after viewport changes.


dogi:
**Review — rating: 83/100**

Promoting `review` → `merge`. Gate at head `59c2e7a7`:
- **CI:** 22/22 check runs succeeded. All started 2026-09-17T19:59:01Z, after the head commit at 19:58:32Z.
- **Merge:** `git merge-tree --write-tree origin/master pr/10452` exits 0. The branch is 12 commits behind but merges clean. It also stacks cleanly with [#10447](https://github.com/open-learning-exchange/planet/pull/10447), which changes `home.component.ts`, and [#10453](https://github.com/open-learning-exchange/planet/pull/10453), which changes `home.scss` and `styles.scss`.
- **Approval:** none from Mutugiii (not required).

What the PR does:
- **Dynamic viewport height:** `--planet-vh` uses `100dvh` behind an `@supports` check, with `100vh` as the fallback. `$view-container-height` and `$dashboard-height` are built on it.
- **Landscape phones:** `usesOverlayNav = isMobile || isShortViewport` sends a landscape phone to the overlay drawer.
- **Drawer margin:** `resetContentMargins()` calls `updateContentMargins()` in `afterNextRender`. `mainContent` is the `#content` `mat-sidenav-container`, which already calls `updateContentMargins()` at line 84, so this relies on the same API.

Checked and dropped:
- **Unsupported `:has()` breaking layout:** the new flex-column rule for `mat-sidenav-content > :has(…)` only adds to the existing `.view-container.view-full-height { height: v.$view-container-height }`. That rule stays, so a browser without `:has()` falls back to the old fixed height. `courses-view.scss` gets its own `.view-full-height { display: flex }`, so `.course-container { flex: 1 1 0 }` doesn't need `:has()`.
- **Stale `forceModern`:** `syncToolbarLayout` now reads `isShortViewport`. Rotating the phone fires `resize`, so the value can't go stale.

Keep in mind: `$short-viewport` in `_variables.scss` and `SHORT_VIEWPORT_QUERY` in `device-info.service.ts` are two copies of the same query. They are only linked by a "keep in sync" comment.

---

dogi (approved):

